PR #8, opened 2015-05-14
Bug fix: Button markers not hiding when character looks away from button

MMaster:
Bug fix for button markers not hiding when character looks away from button and raytrace does not hit anything.


Frontrider:
I can confirm the bug, looking at the button again temporarily fixes the issue.

MMaster:
Closing - Will issue new pull request from separate branch.

PR #10, opened 2015-05-14
Mass property for IMyCubeBlock and IMySlimBlock

mexmer:
Property to retrieve block mass in modapi and ingame scripts
This property retrieves block mass from block definition (eg. not virtual mass, like spaceball hass, but blueprint one)

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMyCubeBlock.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyCubeBlock.cs
@@ -116,6 +116,11 @@ namespace Sandbox.ModAPI
         /// Maximum coordinates of grid cells occupied by this block
         /// </summary>
         VRageMath.Vector3I Max { get; }
+        
+        /// <summary>
+        /// Block mass
+        /// </summary>
+        float Mass { get; }
         /// <summary>
         /// Minimum coordinates of grid cells occupied by this block
         /// </summary>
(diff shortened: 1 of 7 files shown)


LeonidasCZ:
MyCubeBlock also has GetMass function, no need to add Mass property and call SB.GetMass() (which then calls FatBlock.GetMass())

mexmer:
GetMass() function is not defined in IMyCubeBlock interface. Therefore you cannot use it in ingame program scripts.

Also Property accessors looks better than functions ....

LeonidasCZ:
Yes property on the interface side is allright. 
Just the MyCubeBlock.Mass property is redundant.
You can implement IMyCubeBlock.Mass same as
       float IMySlimBlock.Mass
       {
             get { return GetMass();  }
       }
You can use any member/method when implementing the ModAPI interface so 
       float IMyCubeBlock.Mass
       {
           get { return GetMass(); }
       }
in MyCubeBlock_ModAPI.cs would be correct.

mexmer:
not sure, if it's this one - if you mean
MyCubeBlock.cs line 118
        public float Mass { get { return SlimBlock.GetMass(); } }

i've done it, to mimic Min/Max definition. might remove it, of course.

LeonidasCZ:
Yes, that one.
I see that makes sense, but its slightly different case. Min/Max are just shortcuts but in case of the mass cubeblock can get it from itself, SlimBlock.GetMass() would do it anyway. Also generally beware of duplication, having both MyCubeBlock.Mass and MyCubeBlock.GetMass() is confusing. :)

mexmer:
done

OndrejPetrzilka:
Thanks, I like how simple it is

PR #16, opened 2015-05-15
Interior Door/Airtight Hangar Door Fixes

Eikester:
-Fixed Airtight Hangar looping Sound Bug
-Fixed Interior Doors being Airtight when not fully closed
-Added OpeningSpeed Parameter for Interior Door in Block Definition
example: `<OpeningSpeed>0.2</OpeningSpeed>`
defaults to 1 if not defined


Eikester:
allows to define a opening speed for the normal Door in the Block definition, if not defined defaults to 1
example: <OpeningSpeed>0.2</OpeningSpeed>

PR #18, opened 2015-05-15
Now you can do voxel indestructible material!

sjsoft2015:
Added IsIndestructible flag to Voxel Material Definition.
Added check for material in RequestCutOutShape.

With best regards and best wishes
to the author of "Adminium - Indestructable stone."
http://steamcommunity.com/sharedfiles/filedetails/?id=306581923

### DIFF
--- a/Sources/Sandbox.Common/ObjectBuilders/Definitions/MyObjectBuilder_VoxelMaterialDefinition.cs
+++ b/Sources/Sandbox.Common/ObjectBuilders/Definitions/MyObjectBuilder_VoxelMaterialDefinition.cs
@@ -26,8 +26,8 @@ namespace Sandbox.Common.ObjectBuilders.Definitions
         [ProtoMember(5)]
         public bool IsRare;
 
-        //[ProtoMember(6)]
-        //public bool IsIndestructible;
+        [ProtoMember(6)]
+        public bool IsIndestructible;
 
         [ProtoMember(7)]
         public float DamageRatio;
--- a/Sources/Sandbox.Game/Definitions/MyVoxelMaterialDefinition.cs
+++ b/Sources/Sandbox.Game/Definitions/MyVoxelMaterialDefinition.cs
@@ -16,6 +16,7 @@ namespace Sandbox.Definitions
         public float MinedOreRatio;
         public bool CanBeHarvested;
         public bool IsRare;
+        public bool IsIndestructible;
         public float DamageRatio;
         public int MinVersion;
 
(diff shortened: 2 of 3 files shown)


THDigi:
Can you change the number on that parameter ?
I'm asking because the "enabled" parameter is also on that number (from MyObjectBuilder_DefinitionBase) which I've made working in another PR https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/23

sjsoft2015:
it does not make sense, option has been removed due to a change voxel format. but I returned it to its rightful place.

THDigi:
It does make perfect sense, it has number 6, the same as Enabled from MyObjectBuilder_DefinitionBase which means it will overwrite enabled.

Anyway, I noted in my PR that the number needs to be changed if that is ever re-implemented, hopefully when merging the devs will change that number for you.

sjsoft2015:
Dude, I'm not lazy, and could alter.
So, what are you talking about, only makes sense for you. This option is under the number 6, not wilted anything. These are different classes and they may have a different numbering settings.

THDigi:
Well, nevermind xD
I wanted to test my theory but it seem the Enabled parameter doesn't even get picked up in the voxel material, not sure why.

krolli:
Your implementation has problem that it will create sharp/blocky surface whenever this material is being used. Original implementation in the game required storing amount of indestructible content per-voxel, which doubles size of each voxel from 1 byte to 2 bytes. That is why it was dropped.

sjsoft2015:
This decision is at least working as intended. Also needed. Can you come up with a solution for yourself?

krolli:
I don't know what is your idea of 'working as intended', but you've disabled editing of voxels using voxel hand whenever there is certain material present. As a result, if you want to edit that section, you need to first modify material, then edit it, then modify material again. It's definitely not 'working as intended' when it comes to voxel hands.

Unless there is something I don't know about, voxel hands are not available in survival mode, where having indestructible walls makes sense. In creative mode, there should be no limits (by design) so indestructible material is meaningless when it comes to editing with voxel hands (think of those as tools for designers).

Explosions, collisions and drills should be going through CutOutShapeWithProperties() which takes material damage ratio into account. If you're setting the damage ratio to 0 for your material, then it should prevent any changes.

As for how to do true indestructibility, I wouldn't make it part of material but keep it as separate data type inside voxel storage, optimized by a separate octree structure. It would prevent content at some position from falling below the indestructible value on that same position. Then add a separate type of voxel hand that allows editing of this data, While leaving content and material untouched. If you then wanted different look of the material based on whether it is indestructible or not, you can handle that when generating geometry for rendering. However, adding new type to the storage also requires dealing with backwards compatibility, and I would be quite strict about those before accepting any pull requests.

sjsoft2015:
Ok. It's your product, you can develop it as you think is necessary.

PR #23, opened 2015-05-15
Voxel material spawn params

THDigi:
- Added "SpawnsInAsteroids" and "SpawnsFromMeteorites" parameters for
voxel materials, which are default true if undefined;
- Added the params to the first entry of VoxelMaterials.sbc with
comments, also formatted it for consistency;
- Fixed an issue where Ice could be spawned by meteorites when oxygen
systems were disabled.


THDigi:
Fixed merge issues (sneakily :P)

OndrejPetrzilka:
Hi, Dusan will take a look at it, he preparing materials for planet. So we'll need more flags, e.g. organic materials like soil, grass will spawn only on planets with atmosphere.

PR #24, opened 2015-05-15
Find toolbar group actions by type instead of Definition Id

MMaster:
Find toolbar group actions by type instead of Definition Id to support toolbar actions for group of different variations of same block type.

For example group of different modded doors or different sizes of modded Hangar Door do not currently allow toolbar open/close action and only provide Functional Block actions - this fixes it.


Eikester:
perfect :)

PR #25, opened 2015-05-15
Add MaxDistance to SensorBlock Definition

KionX:


No-Brain:
The reason why this was not done already is the performance issue when sensors try to detect voxels.

KionX:
modders and players can decide.

Tyrsis:
This is incomplete.  Can be exploited via projectors and manually modifying blueprints.

zrisher:
Agreed, that's the problem we were running into before they locked it down.

KionX:
\This is incomplete. Can be exploited via projectors and manually modifying blueprints.\

only one solution, compare the parameters from blueprint with reference sample

Tyrsis:
Yes, there needs to be a max value that can be set in the definition for sensors, and then clamp sensors to this max value.  

In MySensorBlock.cs you already have access to the MySensorBlockDefinition (property: BlockDefinition).  You need to add a float MaxDistance to the MySensorBlockDefinition, and then clamp to that in MySensorBlock.  This also requires modifying CubeBlocks.sbc to use the new value and set a default to it's current default of 50.  This would allow mods to increase the max distance if they feel like it (at a performance cost as NoBrain pointed out), but also use 50 by default, while being safe to use in MP.

PR #30, opened 2015-05-15
Extracted magic constant.

mafagafogigante:


YotaXP:
I know that UNDERSCORE_CASE has been used elsewhere in the code, but should we encourage its use?  It's against .Net naming conventions, and inconsistent with constants like `CornerCount`.

mafagafogigante:
Good catch. I am so used to uc that it just happened. I can change it, though.

mafagafogigante:
Done.

Spartan322:
I guess KSWH so rushed themselves they forgot about using constants?

mafagafogigante:
I am not a guru myself, but unless you overtime under a load of coffee, you should spot a literal repeated 6 times easily. Not only a literal, but a very f-up constant that almost asks for its own field.

Spartan322:
Yup

No-Brain:

JCMais:
Wait, what? So we should be the guy that keep magic numbers around?

So what type of of PR are you guys accepting? No code cleanups? Only new features? Is that it?

g4borg:
I guess he meant, don't be that guy who mocks others' code as if he would have never overlooked stuff or even written bad code; It is easy to spot errors in others' code. We all do it constantly.

mafagafogigante:
This is more than ridiculous. You go "open source" with a game whose development is far from finished and do not accept a magic constant extraction because it *doesn't do anything*.

@JCMais Also didn't get it.

@g4borg I don't think that he meant that. Just ignored valid refactoring.

MMaster:
I think I understand it. Changing code just to make it 'nicer' is not a valid reason to make pull request. If you will have some feature or fix that will be related to this then it makes sense to squeeze it in, otherwise it is not necessary to change code that works correctly.

mafagafogigante:
You may understand their reasons to close the PR. But guess what, I also so. The point is that a refactoring like this would made it to Guava or the Linux kernel (magic constants don't exist in those places btw). It is absurd not to merge something just because everything must be a feature or fix.

MMaster:
I understand your point and I understand their point. This is not open-source project - it is product owned by company that needs to test it and make sure that everything works because they sell it. Even if it is early access the approach is a bit different than in open-source. Please don't get mad because of it. It's not a big deal - they decide what goes in and what not as it is their code. We can just propose changes. Just accept it and move on :)

mafagafogigante:
I'm accepting their decision (not doing so is not an option, anyways). I'm just dazzled by how stupid it is. @JCMais also got this.

A close without merge would be OK if the reason is what @YotaXP pointed out, but I even got to fix it.

zrisher:
I also think it's a little surprising this was closed. What's the harm in letting people submit PRs for small code quality fixes? Even if it's not a major change, it takes two seconds to accept.

mafagafogigante:
Even if you only merge locally, it takes really little time to merge.

Spartan322:
It was probably closed because it creates a useless commit, which should come along another commit, often time people don't accept (Github) PRs because the commits are either useless or poorly described, if this wants to get merged, it needs to be of much higher importance.

zrisher:
But it's not "useless" because it has some value, even if that value is small. And what's the downside of having more commits in the history? I agree that if many commits are added at once that are related, they should be squashed. But otherwise there seems to be no disadvantage to having one more commit in there.

mafagafogigante:
Makes no sense. Serious projects such as rails understand the value of these commits. Additionally, each commit should be atomic, presenting only one change to the repo.

Spartan322:
There is actually no help that this PR would provide if accepted, please stop complaining about it now.

PR #31, opened 2015-05-15
Clearer BoundingBox '!=' operator.

mafagafogigante:


No-Brain:

OndrejPetrzilka:
One thing is maintainability other thing is performance.
For performance this is preferred way (for equals):

                return a.Min.X == b.Min.X
                    && a.Min.Y == b.Min.Y
                    && a.Min.Z == b.Min.Z
                    && a.Max.X == b.Max.X
                    && a.Max.Y == b.Max.Y
                    && a.Max.Z == b.Max.Z;

mafagafogigante:
@OndrejPetrzilka so... I don't get it. Just write this in the equals code and reverse it for the not equals operator. I think the MS compiler is good enough to inline equals code in the other function.

zrisher:
It does seem a bit more DRY to include that logic as the equality method for BoundingBox.

PR #32, opened 2015-05-15
Removed useless using directives.

mafagafogigante:


No-Brain:

PR #33, opened 2015-05-15
Trigger Running state after loading config but prior to loading session

seanc722:
Not sure if this is best way to go about it, or if should just prevent service timeout while loading the session.

### DIFF
--- a/Sources/Sandbox.Game/MySandboxGame.cs
+++ b/Sources/Sandbox.Game/MySandboxGame.cs
@@ -90,6 +90,7 @@ namespace Sandbox
 
         public static bool IsConsoleVisible = false;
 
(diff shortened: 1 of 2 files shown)


OndrejPetrzilka:
Can you describe how to reproduce issue with service timeout please?

seanc722:
Running server on a Windows 7 machine (Don't think this matters). When you start the server it stays in start_pending state until the world is completely loaded. Having a fairly large world sometimes takes about a minute to load and after about 30 seconds it throws a message about failing to start. The server still ends up starting though.

I could think of 2 ways to fix the issue. Either set the running state before loading the world (what I did) or during loading the world update the service status (can still be start_pending) so that it doesn't timeout.

PR #34, opened 2015-05-15
Update Projector text when its state changes

MMaster:
Fixes issue where projector doesn't have any text in terminal.
Updates Projector text on init, on projection init & destroy and on block add & remove. 

This is also needed for ingame scripts to be able to reliably read it.


VyMajoris:
Would this update the projector information for the Configurable Automatic LCDs script?
https://steamcommunity.com/sharedfiles/filedetails/?id=403921671

Nice +1

MMaster:
Yes

Issue #38, opened 2015-05-16
Upgrade modules will not function when loading a world if only using a MySessionComponent script


Phoenix84:
Due to conflicts inherent to using MyGameLogicComponent scripts, using them is not always ideal.
This is especially an issue with upgrade modules and the fact we can't create own own block types.

If you have a custom block script using MySessionComponent, and load a world with already placed upgrades, they will not get refreshed and loaded.
They require an event, like block placement, to trigger again.

I propose triggering this refresh event during a more common operation as well, such as power changes.
I have a proposed patch I will attach to this defect.

PR #39, opened 2015-05-16
Issue #38: Force the upgrade module to refresh its state when the power changes

Phoenix84:
This also happens on world load, which allows upgrade scripts not using MyGameLogicComponent to function.

This should happen infrequently enough to not have any performance impact.


Phoenix84:
I'm rather unfamiliar with git and github (I use SVN), so I made the pull request against master.
Would it be better to issue a new one from another branch?

Spartan322:
unless you were planning a bunch of extra things after this or wanted to separate multiple PRs, no

MMaster:
You don't need to but anything you do to your master will end up in this pull request so I would highly recommend doing that. This pull request can be here for quite a long time.

Spartan322:
I agree with @MMaster on this one

Phoenix84:
I have a workaround by hooking to the OnEntityAdd events, and running during the first update frame.
I also want to clear the master branch, so I don't need this anymore.

PR #40, opened 2015-05-16
Add projector only show can build

yajiedesign:


MMaster:
Any description what it actually does would be welcome. "Only Show Can Build" does not seem to be very intuitive. I think it should show only blocks that can be built at this time. Also please don't commit the global.props file. I also recommend not doing pull requests from your master as anything you will do to your master will end up in this pull request.

mexmer:
don't modify global.props
create user.props (those are ignored upon commit), and put your steamdirectory settings in there.

lord-devious:
You are also missing synchronization code for that check box. None of the other players will have their terminal updated when others change that setting.

yajiedesign:
move to https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/47/
remove global.props change and move to branch

yajiedesign:
i am add  synchronization code to SendNewOffset...Independent synchronization method add more change.

lord-devious:
I noticed, sorry :)

PR #44, opened 2015-05-16
Shorthand numbers for ScreenDialogAmount control

UberMouse:
Expands shorthand numbers to full numbers. I've gotten used to this from Runescape and it's convenient for moving items around. Works with both period and comma decimal separators and is case insensitive.  
2k == 2000
2.62k = 2620
1m = 1,000,000
1b = 1,000,000,000

### DIFF
--- a/Sources/Sandbox.Game/Game/Screens/MyGuiScreenDialogAmount.cs
+++ b/Sources/Sandbox.Game/Game/Screens/MyGuiScreenDialogAmount.cs
@@ -1,4 +1,6 @@
-﻿using Sandbox.Common.ObjectBuilders.Gui;
+﻿using System.Linq;
+using System.Text.RegularExpressions;
+using Sandbox.Common.ObjectBuilders.Gui;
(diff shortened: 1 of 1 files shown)


MMaster:
Why is this needed in game code? Also Regex is terribly slow.

UberMouse:
I'm not really sure what type of PRs are being looked for, I just made this for myself because I wanted the feature and figured I'd see if it would be of any interest, happy to close if not. 

I took the slowness into consideration, but that function is only called when the up/down/confirm buttons are pressed so it's not called frequently. I just tested, to construct and match 100 regexps it takes 1ms.

zrisher:
Yeah 100 in 1ms is pretty slow, but if it's only called on `TryParseAndStoreAmount()` it should be fine right?

I suggest squashing your commits together.

UberMouse:
It could at most be called once per frame if the the up/down buttons were being mashed. 

Just squashed them.

PR #45, opened 2015-05-16
Allow modding the Laser Antenna's line-of-sight requirements

Phoenix84:
Added the option for modders to create laser antennas unrestricted by the line-of-sight requirement.

### DIFF
--- a/Sources/Sandbox.Common/ObjectBuilders/Definitions/MyObjectBuilder_LaserAntennaDefinition.cs
+++ b/Sources/Sandbox.Common/ObjectBuilders/Definitions/MyObjectBuilder_LaserAntennaDefinition.cs
@@ -21,5 +21,7 @@ namespace Sandbox.Common.ObjectBuilders.Definitions
         public float RotationRate = MathHelper.Pi / 20000.0f;
         [ProtoMember(5)]
         public float MaxRange = 40000;
+        [ProtoMember(6)]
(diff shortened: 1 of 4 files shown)


mexmer:
you know it's laser ... it cannot pass trough material, therefore it requires LoS

this is question of principle. wait for regular directional antena

Phoenix84:
In the code, what's the difference between laser and regular directional antenna?
With this, you now have a directional antenna.

I understand what you mean by 'laser', but I didn't change the default behavior, I exposed functionality for it.
I was also not aware it was 'in-progress,' as there's no reference to it in code.

mexmer:
you have good point here. most likely from code point of view, there will be no difference, since p2p antena connection doesn't require to eval list of broadcaster, it just need to track endpoint.

and idk if they are in progress (directional antenas), it just would make sense to me, if they are.

also don't forget to add value to cubeblock.sbc

Phoenix84:
I thought about cubeblocks, but I wasn't sure if that was necessary.
The only reason to change it would be for an example, as it has a default value already.
Is that a general rule to follow then?

mexmer:
i think so ... when you add parameter to definition for block builder, you need also add it to block definition data.

OndrejPetrzilka:
We will probably rename laser antenna to p2p antenna later.

PR #46, opened 2015-05-16
Add MaxDistance to SensorBlock Definition

KionX:


Spartan322:
How about squashing your comments?

PR #47, opened 2015-05-16
 Add projector only show can build

yajiedesign:
add some checkbox with projector only show can build block.


Roberto5:
good idea, maibe now the projector will not slow the game?

Phoenix84:
Just to be clear, what does this do?
Does disable the projection itself, but you can still build if you know where the blocks are?

JCMais:
@Phoenix84 from what I can tell, it hides the projection of blocks that were already welded.

zrisher:
Yes this would be significantly more legible if the names were changed to `onlyProjectBuildable` instead of `onlyCanBuildBlock`

RossM:
You need to add your new field to the sync code, otherwise it won't work correctly in multiplayer.

yajiedesign:
@RossM is add in SetNewOffset

OndrejPetrzilka:
Alex will review that before merging, the names should be changed to:
Show only buildable

yajiedesign:
change ok

yajiedesign:
@OndrejPetrzilka  change ok

PR #49, opened 2015-05-16
Programmable block argument continued (with bug fixes)

lord-devious:
Finished off the programmable block argument feature by running the script when pressing enter in the argument text box. Also, added a check box to make the programmable block clear its argument field after running, to enable console like scripts.

Also added a missing sync message.


MMaster:
:+1:

THDigi:
A small bug with the feature that's in the released game, drag & dropping the PB using right-click will cause the action list to be displayed a second time.

Also, a small suggestion: changing the PB action from a taskbar slot (rightclick on it) should show the argument dialog again with the existing argument if any.

lord-devious:
The timer problem was a symptom of a slightly larger problem, some synchronization messages I weren't aware of.

While I fixed the timer problem, it involved new memory allocations in the update messages. Now fair enough, there were allocations already but I needed to add more. So @OndrejPetrzilka, when the time comes that you or one of your guys are going to review this one, I'd appreciate extra attention to the ToolbarItem, its new Parameters property and its use. Any improvement suggestions would be very welcome. 

@THDigi I will look at the right-click drag-drop a little later :smile:

lord-devious:
Keen, also be aware that the Toolbar changes made here must be used for all new toolbars, because otherwise the action parameters won't synchronize. I have added the fix for the new MyRemoteControl toolbar, but I can't push it because the GitHub source does not compile at the time of writing this comment, so I can't test it.

lord-devious:
There, that should do it.

OndrejPetrzilka:
I've removed comparers in 4721215c7baf67ad279db5f90366f4b6c4e781ad
- it was not used at all
- the pattern was weird

lord-devious:
Glad to hear it :smile:

OndrejPetrzilka:
I've tried to understand it, but it seems to be quite big and there's a lot of small changes I don't fully understand. Would it be possible to split it into multiple pull requests?

1) Refactoring (I like some thing you did there)
2) First feature 
3) Second feature (if possible to split)

lord-devious:
Actually the feature part of this is tiny, everything else is bugfixing meant to make sure the terminal action arguments are synchronized in the toolbars as well as just the action reference, so the programmable block argument feature can work. This is especially true for timers which for some reason uses sync serialization even locally, so the argument never worked with them.

Essentially I don't think it would be any smaller, and if the programmable block argument feature is to be supported, this or something similar is required.

As the bugfix segment grew I regretted not splitting them apart myself.

lord-devious:
I'll attempt to pull the bug fix segment out to a separate PR as that is the most important one as people have been reporting it in the bug forum, and asking me about it.

lord-devious:
This feature is being split into two, starting with #216

PR #52, opened 2015-05-16
Expose block deconstruction to ModAPI

zrisher:
As a mod developer, I wanted to restrict player placement of certain types of blocks and remove them when over a certain limit. The logic's easy enough, but spawning the components contained within an existing Cubeblock / moving them into player inventory was not possible.

This PR contains the bare minimum to get done what I wanted to, but I'm happy to help whitelist other functions if needed. I wanted to keep my footprint small because I don't know why this was blacklisted in the first place, and I'm not a very proficient C# coder.

The first commit in this PR whitelists `MySlimBlock.MoveItemsFromConstructionStockpile()` for the ModAPI so that us modders can take items from a ConstructionStockpile and move them into another inventory. 

The second commit adds and whitelists a new helper function `MySlimBlock.FullyDismount()`, which mimics the logic of `DecreaseMountLevel()` in the much simpler case where you want to deconstruct it entirely. 

I was torn about the best way to implement this. `DecreaseMountLevel()` was not yet whitelisted, but in order to use it I'd also have to use the block's `MaxIntegrity`, `DisassembleRatio`, and `IntegrityPointsPerSec` (the latter two aren't whitelisted either) to determine the `gridAmount` to pass through to `DecreaseMountLevel()` that would fully deconstruct the block. So this seemed like a useful helper function, regardless of its use for ModAPI.

I considered deconstructing the logic of `DecreaseMountLevel()` into various methods to decrease repetition between it and `FullyDismount()`, but it seemed like that didn't fit so well with the dev methodology of the rest of the project, and `FullyDismount()` was fortunately a much simpler case of `DecreaseMountLevel()` so it seemed less necessary. Please let me know if I should adjust.

I've tested this in both Dev and Release mode with a mod making use of the whitelisted functions successfully.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMySlimBlock.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMySlimBlock.cs
@@ -21,6 +21,7 @@ namespace Sandbox.ModAPI
         //void DoDamage(float damage, Sandbox.Game.Weapons.MyDamageType damageType, bool addDirtyParts = true);
         IMyCubeBlock FatBlock { get; }
         void FixBones(float oldDamage, float maxAllowedBoneMovement);
+        void FullyDismount(IMyInventory outputInventory);
         //int GetConstructionStockpileItemAmount(Sandbox.Definitions.MyDefinitionId id);
         Sandbox.Common.ObjectBuilders.MyObjectBuilder_CubeBlock GetCopyObjectBuilder();
         void GetMissingComponents(System.Collections.Generic.Dictionary<string, int> addToDictionary);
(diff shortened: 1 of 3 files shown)


OndrejPetrzilka:
Will probably merge, but with some refactoring which would reduce code duplication.

czestmyr:
I agree, code duplication should be avoided. Also, MoveItemsFromConstructionStockpile should only be callable on the server, the same as FullyDismount.

zrisher:
@OndrejPetrzilka awesome, happy to help with that if you like if you let me know how you'd like it built.

@czestmyr, I was under the impression that this took care of ensuring it only ran on the server:
```
if (!Sync.IsServer)
    return;
```

PR #53, opened 2015-05-16
Changed spelling

sisco035:


Spartan322:
is this really worth a push with nothing else contributed?

PR #54, opened 2015-05-16
Hide exterior terminal access points

Phoenix84:
This change adds a UI button to toggle exterior terminal access for various blocks.
It does this by removing the terminal dummies you interact with.

I wrote this to be an extension of the 'Show In Terminal' feature.
It also doubles as a way for Text/LCD panels to show a display, while not having the yellow dummy covering the entire display when looking at it.

I like to put more work into this, but I'd like to get feedback before proceeding further.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Common issues
 Common when using old versions of visual studio, see [Visual Studio support](https://github.com/KeenSoftwareHouse/SpaceEngineers/wiki/Visual-Studio-support). It can also happen when  TextTemplating.exe was not found for some reason (it should be installed with Visual Studio).
 
 **Assert: unable to find audio/model/texture file: 'xxxxxx'**.
-This happens because repository is slightly ahead of content in Steam folder. Definitions (Content/Data) are taken from repository and may contain new definitions referencing assets which are not yet in Steam content folder. We decided to use definitions from repository by default, so you can easily modify it. You can edit **global.props** to use definitions from Steam (that should fix the issue). More info [here](https://github.com/KeenSoftwareHouse/SpaceEngineers/wiki/Initial-setup#setting-path-to-the-games-content).
+This happens because repository is slightly ahead of content in Steam folder. Definitions (Content/Data) are taken from repository and may contain new definitions referencing assets which are not yet in Steam content folder. We decided to use definitions from repository by default, so you can easily modify it. You can edit **global.props** to use definitions from Steam (that should fix the issue). When running on **Release** asserts won't be shown; missing assets won't crash the game.  More info [here](https://github.com/KeenSoftwareHouse/SpaceEngineers/wiki/Initial-setup#setting-path-to-the-games-content).
 
 Where is 64-bit version?
 ------------------------


MechanizedIT:
I find this a useful contribution, especially the annoyance of having the yellow dummy covering the whole LCD. I wish it was at least much more subtle.

lord-devious:
This is a very good idea, I like it a lot, well done. I assume this works on doors as well? I have a pull request on a previous repo yet to be merged, which disables terminals on doors only. Yours is a superior way to do it, since it even removes the interaction dummies. I was only thinking about airlocks :smile:

 However I'm not sure how well your hardcoding in MyCubeBlock will be received. Earlier conversations with Keen devs have indicated otherwise, but who knows.

Phoenix84:
Thanks for the feedback!
You know, I do remember seeing your pull request, but it didn't click to me that they're related. I did this to have clean LCD panels. :-)

Doors use the detector_terminal dummies, so they are affected by this as well.
The best thing is that the actions are unaffected, so you can assign them to buttons and use them in programming blocks as always. In fact, you can apply this to button panels too, so you have secure public buttons!

I wasn't a fan of the hardcoding either.
Will keen provide feedback if they have issues with it? If they want me to fix something, I'm more than happy to.
What I'd rather do is replace all the strings with constants, including the originals used in the various MyUseObjectAttribute (iirc) scattered around. I just didn't want to go too far with it yet, and didn't know how well that might be received.

The other thing I want to do is turn off conveyors in blocks like the cargo containers and reactors.
Currently this doesn't affect those, as their access ports are also their conveyor ports.
To do that, I'd have to keep track of the Use Conveyor system button as well. That way I can properly alter the state if that button changes as well since I don't want to turn off the conveyors if they are in use. The terminal panels are disabled though.
That's a bit more work, and I wanted to wait and see if people are interested in that before pursuing it.

lord-devious:
Ondrej told me when I did something wrong, gave me a chance to fix it.

I'll add a comment to my request so he can make a decision to ignore mine, if he's interested in this one instead.

zrisher:
If it really needs to be deleted, just include that deletion in a separate commit within your PR.

```
git checkout your-pr-branch
git checkout master that-file
git add -A
git commit -m "added file back"
delete the file
git add -A
add the file to .gitignore
git add -A
git commit -m "removed and ignored auto-generated file"
git rebase -i
```

At this screen, squash your "added file back" commit with your original by changing "pick" to "squash" (or simply "s"). Keep the pick for your second commit. Go through the rebase, choose your message for the squash commit (I usually just keep the original commit msg and drop the squash msg), et voila you should have two separate commits, the latter with your file change.

Phoenix84:
@zrisher Sorry, I don't really know git that well (I use SVN).
I understand pushing the delete as a separate commit, but it is already is.
I understand the concept of .gitignore, SVN has a similar mechanism.

When I've run into a problem I've been googling it, but I don't really know what the commands I'm using are. :confused:

What MMaster said made sense, issuing a separate PR with that and other delete operations.
However I don't know what to do with this.
From what I've gathered I can reset the branch to a previous state (like a rollback, but SVN doesn't support this without a dump and load), then reapply the commits I really want.
I did this before when I was poking around in git, but I don't really know all the ramifications. Is that something that would work here?

zrisher:
Hey! Sorry didn't realize it's already another commit. Yep you do do that, it's super easy:

Assuming you've set origin to your fork:
```
git checkout HideTerminalDummy
git checkout -b remove-generated-spacetexts          ( duplicate your current branch )
git push -u origin remove-generated-spacetexts       (  push it to your fork )
```
Next, remove the extraneous commits from each. So for this branch:
```
git checkout HideTerminalDummy
git rebase -i master
```
Then go to the line that says `pick a80faa0` and delete it, thus telling git to forget about that commit on this branch. Save the file, it'll run the rebase. 

You can check your new commit history with `git log`. If it looks like you have exactly the commits you want, you can go ahead and push back to your fork:
```
git push -f 
```
To push your new branch back to your fork. The force option `-f` is needed when changing commit history.

Next, do the same for your other branch:

```
git checkout remove-generated-spacetexts
git rebase -i master
(remove both pick lines except for a80faa0 and save the file)
(let it run the rebase)
git log (looks good?)
git push -f 
```

PR #55, opened 2015-05-16
Exposed GridTerminal of linked Laser Antenna

MeusRex:
This change exposes the IMyGridTerminalSystem of the receiving laser antenna.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyLaserAntenna.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyLaserAntenna.cs
@@ -16,5 +16,12 @@ namespace Sandbox.ModAPI.Ingame
         void SetTargetCoords(string coords);
 
         void Connect();
+
+        /// <summary>
+        /// Gets the linked receiver's IMyGridTerminalSystem. Returns null if no link is established.
+        /// </summary>
+        /// <returns></returns>
+        IMyGridTerminalSystem GetReceiverGrid();
+
     }
 }
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyLaserAntenna.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyLaserAntenna.cs
@@ -749,6 +749,24 @@ namespace Sandbox.Game.Entities.Cube
             return null;
         }
 
+        /// <summary>
+        /// Gets the linked receiver's IMyGridTerminalSystem. Returns null if no link is established.
+        /// </summary>
+        /// <returns></returns>
+        public IMyGridTerminalSystem GetReceiverGrid()
+        {
+            if (State == StateEnum.connected)
+            {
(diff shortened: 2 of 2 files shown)


MeusRex:
Need to rethink my approach to this a tinsy bit. As this leaves to big a security hole. I will propably replace it with a method to set the arguement line of a specified PB block and a method to run said block's script. This should lead to a behavior akin to messenges.

mexmer:
There is code, that exposes connected antena. which is imo more logical approach

mygrid->antena->otherantena->othergrid

MeusRex:
I'm not going to continue this. Handing another grid to the player is very easy to exploit, since you can easily store the whole grid or specific blocks, making a continued connection unneccessary.

Instead a replaced it with SendMessage(...)

Soren025:
@MeusRex Why not create a wrapper for the other grid that act as the connection and invalidate the wrapper causing an error if still used after the connection is broken?

MeusRex:
Hehe, that was also my first thought on the issue. The problem is that you could either store single blocks, ignoring the decay of the wrapper, or you'd need an interface in the wrapper for every type of blocks.

Messages are, in my opinion, a way more elegant solution. Because you can do everything you could with access to the grid, without punching a giant hole into the security or an extensive wrapper.

Soren025:
True, a lot of work for that.

In regards to messages, instead of them directly manipulating the grid, it would be neat if instead there would have to be a script running on the target grid listening for that message.

mexmer:
@MeusRex that is solved by ownership. Even if you can store and refer other grid, you can't do anything, unless you have correct access.

First of all, you should not be able to connect to other antena, that doesn't grant access (eg. at least faction share), so it's really not issue.

You are now able to manipulate grid on other side of laser antena using terminal (K), there will be no difference in approach trough pb

MeusRex:
Ownership is still the same, so the only thing that can stop you from accessing would be a logical connectivity test, which I'm not sure if it is done.

mexmer:
I see the point where you getting at with connectivity test ... that makes sense. Last time i tested it, it was not done. You could open remote grid in terminal, then disconnect LA, and you still had cotrol. It's worth test it, if it's still possible.

MeusRex:
I might do it a bit later today, when I'm back home.

Btw. do you have any idea if it'spossible to move a pull request and all changes to a branch after you commited them to the main branch?

mexmer:
if you mean on git, then yes. either rebranch whole thing, or just merge it to new branch, depends what you prefer.

zrisher:
If you have some commits on branch `master` and you want to move them to `feature-branch`, simply go to master with `git checkout master`, checkout that state to a new branch `git checkout -b feature-branch`, and push it up `git push -u origin feature-branch`. You can't change the branch that a PR points to, however. That's why they recommend submitting PRs from feature branches. : /

MeusRex:
Thanks for the hints.

I tested it and can confirm that you can still access stored blocks even after any kind of connection between the two grids are broken. (Tested it with Echoing the TargetCoords of the other, disconnected laser antenna.)

PR #57, opened 2015-05-17
Update pool's free resources on quality change

### LINKED ISSUE #43: Only one LCD/text panel renders at a time
Reproduce by placing multiple panels that are set to display text, close enough to each other that you can see both of them. Turn so that only one LCD is in view. When turning back to view both LCDs, the other will read OFFLINE. Affects singleplayer as well as multiplayer.

This one's been bugging me for awhile, so I looked into it. The [GetRenderTexture](https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/master/Sources/VRage.Render/MyRenderTexturePool.cs#L24) method is responsible for returning the texture to which the text is rendered. The method successfully returns one texture, but subsequent calls fail and return null, due to the [`m_currentFreeResources >= textureResourceSize`](https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/master/Sources/VRage.Render/MyRenderTexturePool.cs#L29) check failing. The [initial value for free resources](https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/master/Sources/VRage.Render/MyRenderTexturePool.cs#L17) doesn't seem very large to me. If I just change it, multiplying it by N, I'm capable of rendering N text panels at a time, with no apparent problems.

I think the resource limit should be set dynamically, somewhere. I'm not really sure where that should happen or what the calculation should look like.

cwc:
Fixes #43.

Previously MyRenderTexturePool only reset the current free resources
when ReleaseResources() was called. Since that method was only called
when the renderer unloaded content, clients with higher than normal
quality settings would go into their first world with a NORMAL-sized
resource pool. This meant that only one LCD panel could render at a
time, until returning to the main menu (causing an UnloadContent() call).

### DIFF
--- a/Sources/VRage.Render/MyRenderTexturePool.cs
+++ b/Sources/VRage.Render/MyRenderTexturePool.cs
@@ -14,7 +14,8 @@ namespace VRageRender
     public class MyRenderTexturePool
     {
         const int POOL_RESOURCES = 512*512* 20;
-        static int m_currentFreeResources = POOL_RESOURCES;
+        private static int m_maxFreeResources = POOL_RESOURCES;
+        static int m_currentFreeResources = m_maxFreeResources;
         static float m_renderQualityScale = 1;
 
         static Dictionary<MyRenderTextureId, Texture> m_renderTextureTargetsInUse = new Dictionary<MyRenderTextureId, Texture>();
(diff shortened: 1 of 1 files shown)


lord-devious:
Sweet! :smile:

MMaster:
:+1: Good job. I would recommend not issuing pull requests from your master as anything you will do to your master will end up in this pull request.

cwc:
@MMaster One man's master is another man's feature branch. :wink:

PR #58, opened 2015-05-17
Improve asteroid generation

RossM:
Add a bit more realism and variety to the asteroid generator.

Warning: This will drastically change all the unmodified asteroids in an infinite game.

Add a metaball-based CSG primitive, and use it for the base solid on some asteroids. This generates an irregular bloblike shape that looks more like a real asteroid.

Change the way materials for asteroids are generated. Each asteroid is assigned to one of five classes, loosely based on real asteroid types. Each class has a different combination of surface and core material and in some cases an inner core material. Each class also has modified frequencies for the various ore types:
* Class C asteroids have an ice surface, stone core, and iron inner core. They have more light elements.
* Class S asteroids have a stone surface and iron core. They have more light elements.
* Class M asteroids have a stone surface and a nickel-iron core. They have more heavy elements.
* Class E asteroids are completely stone. They have more heavy elements and more ore deposits overall.
* Kuiper belt objects are completely ice. They have more light elements.

Overall, the material changes result in much less stone and iron, much more nickel and ice, and slightly (3-10%) increased frequency for other ores.

Screenshots: http://imgur.com/a/wiQsq

### DIFF
--- a/Sources/Sandbox.Game/Engine/Voxels/MyCsgShapeMetaball.cs
+++ b/Sources/Sandbox.Game/Engine/Voxels/MyCsgShapeMetaball.cs
@@ -1,0 +1,123 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using VRage.Noise;
+using VRageMath;
+
+namespace Sandbox.Engine.Voxels
+{
+    class MyCsgShapeMetaball : MyCsgShapeBase
+    {
+        private Vector3 m_translation;
+        private Vector3[] m_balls;
+        private float[] m_weights;
+        private float m_radius;
+        private float m_halfDeviation;
+        private float m_deviationFrequency;
+        private float m_detailFrequency;
+
+        private float m_outerRadius;
+        private float m_innerRadius;
+
+        public MyCsgShapeMetaball(Vector3 translation, Vector3[] balls, float[] weights, float radius, float halfDeviation = 0,
+            float deviationFrequency = 0, float detailFrequency = 0)
+        {
+            m_translation = translation;
+            m_balls = balls;
+            m_weights = weights;
+            m_radius = radius;
+            m_halfDeviation = halfDeviation;
+            m_deviationFrequency = deviationFrequency;
+            m_detailFrequency = detailFrequency;
+
+            ComputeDerivedProperties();
+        }
+
+        internal override ContainmentType Contains(ref BoundingBox queryAabb, ref BoundingSphere querySphere, float lodVoxelSize)
+        {
+            for (int i = 0; i < m_balls.Length; i++)
+            {
+                ContainmentType outerContainment;
+                BoundingSphere sphere = new BoundingSphere(m_translation + m_balls[i], m_radius * m_balls.Length + m_halfDeviation + m_detailSize + lodVoxelSize);
+                sphere.Contains(ref queryAabb, out outerContainment);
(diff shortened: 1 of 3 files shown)


cwc:
:+1: I like it. Would there still be at least some toroidal roids? Yeah it's unrealistic, but they're cool. :grin:

lord-devious:
IMHO he donuts should be eliminated, they're ridiculous. I'll call Homer. :smile: 

Love the new look.

JCMais:
Really nice. KSH please consider this!

zrisher:
:+1:

No-Brain:
Can't merge something that is not backwards compatible

THDigi:
They look like floating liquid, odd =)

I belive you can make it as a new generator version so existing saves won't be affected by it unless they're edited to change the generator version.

RossM:
Added a new generator version, so old saves will be unaffected.

No-Brain:
The performance hit from depositCountMult would be quite high.

mrsib:
:+1:

RossM:
No-Brain: I profiled it and you're right. I also noticed that metaballs were excessively slow due to a bad Contains() implementation. I improved the Contains(), and changed the deposit modifier to increase the size of the deposits rather than the number of deposits.

liorabel:
This has been bugging me in-game since infinite worlds were introduced. Fantastic!

Pingger:
+1 I really do like this idea.

thelastWallE:
Can we see an updated screenshot with the changes to Contains()?

OndrejPetrzilka:
krolli will check this before it's merged
performance and possible bugs are biggest concerns

Geneticus:
I'd like to see better subtraction to make procedurally generated caves/tunnel systems.

A couple concerns :
*Should the count be hardcoded or a ratio of available materials (including modded ones)?

*While I like the different types, I would like to be able to set the frequency of each type in the world settings somewhere.( allowing for a custom mostly cometary map)

*There needs to be some kind of coordinate checker for each  type to skew the distribution so that Kuiper types spawn mostly in the outer edges of the world in the last AU or so, with the heaviest close to 0,0,0 and a progressively lighter mix further out.  This really forces players to explore to get materials or pirate ships until they can get to large deposits of the other materials.

RossM:
Modded materials will show up with the heavy element types.

Changing the overall frequency of different asteroid types will make some ores more rare - the current ratios are carefully balanced so that the relative ore availabilities stay roughly the same. It would need balance consideration.

Having more large-scale patterns in the asteroid distribution is an interesting idea for future work.

Geneticus:
It would be better I think,  to have <ElementWeight> be moddable in the voxels.sbc. For example, one of my mods adds 4 types of Ice. (Frozen Noble gasses, Methane, Hydrocarbons and Amino Acids) They should be flaggable as Kuiper types. I also have around 20 different materials from Lighter elements like Lithiium, Carbon, and Sodium to Heavier ones like Mercury and Lead. 

Having mods imbalances things in the first place. With 20 some additional materials I have no way to bring it back into balance, which should be my job as the modder who created the imbalance in the first place.

:) Glad you like it. Although I think tunnel creation via subtraction is boss ;)

RossM:
Adding an ElementWeight property or similar to the data files is a good idea, actually. It would be fairly straightforward to just have the three categories I defined here, but it would be good to have some extensibility, so you could add a category for things like hydrocarbons and amino acids.

PR #61, opened 2015-05-17
Send Message to PB via Laser Antenna

MeusRex:
This addition enables users to set the argument of a programmable block which is connected to the current grid by a pair of Laser Antenna.

It also adds a method to run the script of said pb.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyLaserAntenna.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyLaserAntenna.cs
@@ -16,5 +16,12 @@ namespace Sandbox.ModAPI.Ingame
         void SetTargetCoords(string coords);
 
         void Connect();
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


lord-devious:
Good idea! :smile:

However the TerminalRunArgument property should not be used in this way, since it's purpose is to be used via the Run button in the control panel _only_. It would be better to run the PB with a proper argument, like buttons etc. do.

MeusRex:
So, you say I should just skip setting the TerminalRunArgument field and just run the pb's script with the (string)message as argument?
I guess I can do that, no problem. Thanks btw. for the whole TerminalRunArgument.

lord-devious:
Yes, that's what I'm suggesting. It would probably be more efficient too, a single message rather than two.

MeusRex:
Removed SendRunCommand, since it became redundant.

SendMessage now runs the script of the specified pb with a custom argument, which doesn't replace the argument text field.

MeusRex:
Continued here: https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/75

Now from branch instead of master. (never used git before...)

PR #63, opened 2015-05-17
add weapon camera zoom

yajiedesign:
when you Control weapon,can zoom camera .


Spartan322:
just to note, this behavior may not be wanted in all instances, might be better to make it an optional thing.

zrisher:
I code for SE a lot more than playing it, but I have a feeling this has balance implications. I know turret sniping from a distance is a big thing and it sounds like this would make that a lot easier - no longer a need to have a separate camera so your shots will always be dead-on. I can't speak to if it was a design decision on their part to disable zoom on weapons, but I know some people have put a lot of time into build cross-hairs for exactly this reason.

Spartan322:
I agree, this seems exploitable, at least put a limit on it (or let someone put a limit on it somehow) also this should be an optional variable, because this behavior may not be wanted.

PR #64, opened 2015-05-17
Add unit testing infrastructure

alex-davidson:
This adds the basic infrastructure for using the SpaceEngineers codebase with testing tools like NUnit and NCrunch. It also ensures that files left behind by the Resharper tool will not be accidentally committed.

* Project file default build settings now match those implicitly applied by Visual Studio. This is required for NCrunch to build them.
* .gitignore excludes tempfiles and user-specific files created by Resharper and NCrunch.
* A basic test project has been added: VRage.Library.Tests. It contains a single (trivial) test against IlCompiler using the NUnit library.
* IlCompiler has been refactored slightly to remove dependencies on static state, enabling tests against it to run in isolation.


cwc:
:+1:

PR #67, opened 2015-05-18
Add radar block type for modding

RossM:
Add the ability to add a radar block with modding. No actual block is included, mainly because I am a terrible 3D modeller.

Radar detects voxels (asteroids) and grids (ships) and displays them on the HUD, similar to how ore detectors display ore clusters. Like ore detectors, it can be relayed via antennas.

Radars can be detected by antennas while active, just like antennas and beacons.

To help balance radar, it is restricted in how well it can distinguish close-together objects. Large objects (like asteroids) will prevent the radar from picking up smaller objects nearby when the radar is distant. The closer the radar gets, the better it can detect small objects hiding near larger ones.

Radar also has a reduced range when picking up small objects. The smaller the object, the worse the range.

There is a minimum distance of 100m for radar to detect objects, mainly to reduce clutter from it picking up the ship it's mounted on.

Images: http://imgur.com/a/srQvn


Spartan322:
Might want to do a squash

PR #69, opened 2015-05-18
Add radar block type for modding

RossM:
Add the ability to add a radar block with modding. No actual block is included, mainly because I am a terrible 3D modeller.

Radar detects voxels (asteroids) and grids (ships) and displays them on the HUD, similar to how ore detectors display ore clusters. Like ore detectors, it can be relayed via antennas.

Radars can be detected by antennas while active, just like antennas and beacons.

To help balance radar, it is restricted in how well it can distinguish close-together objects. Large objects (like asteroids) will prevent the radar from picking up smaller objects nearby when the radar is distant. The closer the radar gets, the better it can detect small objects hiding near larger ones.

Radar also has a reduced range when picking up small objects. The smaller the object, the worse the range.

There is a minimum distance of 100m for radar to detect objects, mainly to reduce clutter from it picking up the ship it's mounted on.

Images: http://imgur.com/a/srQvn


zrisher:
First, let me say this is an **amazing** addition that I think a lot of people would benefit from having in core. There are clear balance implications, but not including the radar block itself gives full control to server admins to include mods that implement this in a balanced way - awesome.

We've been working on implementing radar for a long time in the [Autopilot mod](https://github.com/Rynchodon/Autopilot), and one of the main hangups has always been displaying the information - since we couldn't add GPS markers, users would have to copy and paste GPS coords from a text panel and manually add them. Now that GPS modding has been enabled we could add/remove them that way. But dedicated HUD markers are a better solution. It's also really nice that implementing this in core means getting a dedicated TerminalPanel. 

I have suggestions on both functionality and implementation. I'll start with functionality because it's way more interesting:

**Functionality**

* **Control Names** - I would rename some of your control parameters to make better sense to a new user:
  * "Minimum Size" -> "Minimum Detectable Size" or "Minimum Detected Object Size" or "Minimum Detection Size" - anything to help make it clear that this has nothing to do with detected objects and not the size of your detection sphere
  * "Maximum Size" - same as above
  * "Broadcast Using Antennas" -> "Share Detection Info Via Antennas" or "Broadcast Detected Entities" or something to help clarify that this has nothing to do with the *radar signal*. Also, I'm not sure where Keen lands on "Antennas" vs "Antennae".
  * "Tracking Limit" -> "Max Entities Displayed" or "Shown", something that helps clarify this has nothing to do with the tracking angle of the radar signal
* **Signature Algorithm** - in order of how important I think they are:
  *  Detect other radar signals - In real life, the easiest way to give away your position to a radar installation is by running radar of your own, or any particularly noisy electromagnetic signal. I think this should be an element of the detection algorithm that's much stronger than size, and preferably can be exposed to the GUI / via the results method. 
  *  Tech Level - Building on the above, [there are a lot of implementations of radar](http://en.wikipedia.org/wiki/Radar#Antenna_design), some much more sophisticated than others. Generally there's tiers of technology that can detect everything earlier than them but hide from similarly advanced methods (I'm thinking Passive and Active Phased Array here). I would suggest implementing this as a simple 3 tier system - 0,1,2 - and making it an attribute of the block definition. So if you have a radar level 0 block, you can detect other radar level zero. If you have a 1, you can detect 0 and 1. If you have 2, you can detect 0 and 1 (like in real life) or 0,1, and 2.
  *  Jamming - The logical opposite of other signals making it easier to detect, you would have a Radar Jamming (or even all communications jamming eventually) block that would make it *harder* to detect its grid and nearby objects. This would probably be easiest to implement after you'd done the nearby object check [here](https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/69/files#diff-8fb6c14c972293c057b9d6c7b0effd51R122).
  * Orientation - A single radar block should have a semi-circle of detection at best. Many of the largest, most powerful radar systems are actually uni-directional. Many modern military bases have giant passive phased array dishes installed on each side. I would recommend allowing the block definition to define angle limits like turrets have, then reading these and comparing against the angle to a detected object.
  * Object curvature - *Eventually* you could take into account the curvature of the vessel using some real radar maths. This could be cached for a long period of time, it only really needs to be updated when there's a lot of building going on on the object, and then it can wait some frames to see if more building is occurring. This is probably the least important feature I'm mentioning and some of the most work. It'd just be real cool.
* **Detected Entities**
  * Show enemies first - Currently this just shows the biggest X entities, but enemy entities are far more important to know about than neutral and friendly. I'd suggest you sort it enemy, neutral, friendly assuming that neutral and friendly parties would have updated your system with enough info to distinguish between them and enemies. If that's too much of a logical leap, we should at least avoid showing markers for entities we already have HUD markers for. 
  * Show signature size on HUD Marker
  * Remove old markers - it looks like these don't get removed from the targets_cache, and thus not the HUD, until they're replaced with more recent entries. I could be reading that incorrectly though.
* **API access** 
  * Expose signature size and position of detected entities to modders and in-game scripts

**Implementation**
* Magic Numbers - there's a lot of them. These should be moved to constants declared at the beginning of the class or namespace.
* Unlimited constants - The value used to identify "unlimited" for object tracking size and displayed entity count should probably be 0 or a negative number instead of a positive one. Using a positive number limits modder's ability to actually increase past that amount.
* Size Signature - it looks like [your current implementation](https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/69/files#diff-8fb6c14c972293c057b9d6c7b0effd51R115)  is a simple step function. We've got a nice way of doing this via a continuous function in [Autopilot](https://github.com/Rynchodon/Autopilot/blob/master/AntennaRelay/Scripts/Beacon.cs#L138)
* Bug in nearby-object detection - It seems like [this](https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/69/files#diff-8fb6c14c972293c057b9d6c7b0effd51R122) compares each object to the next largest object and finds if its close, great. But it should drop the object at index `j`, not index `i`. 
* Bug in entity broadcasting - it looks like [this](https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/69/files#diff-f069e40ded1d20292037e8e105cbee82R274) uses the terminal block's position, it needs to use that of the radar it's querying and really should just share what radar already has cached
* Performance - Have you done any profiling of this? I know anything that gets a list of entities in a sphere is usually moderately expensive, especially if it's a big sphere. I'm not great at understanding logic flow in this game yet - does this run everywhere at once, or can it run the only on the client who needs to consume the information? I can see having to run the radar detection for my own ships as well as those of everyone else on my server being as CPU-intensive. Could we delay it to run every 300 frames or so more?

Again, awesome PR and I'm happy to help groom this up. I wanted to run this all by you and the community first though. 

Also FYI, when you make changes to this branch you don't have to open a new PR, just push your commits to the same branch. Even if you change the commit history via a rebase or squash, you can just do `git push -f`.

RossM:
Bug in near-object detection: Nope, that's deliberate. We go from largest to smallest, and only compare against objects that have already been accepted. So if there is a nearby object in j, that means i is too near and should be droped.

Bug in entity broadcasting: Also nope. Look just above and you'll see it's just like ore detectors: it's setting the flag that allows radar to transmit its results via antenna system. The transmitting its own location is done via a MyRadioBroadcaster.

Performance: Yes, I've profiled it. Surprisingly it's quite cheap. This code only runs on the client who needs the information.

Logic in general: I went for simplicity over realism.

Magic numbers: Guilty. I'll fix that up.

Unlimited constants: Unfortunately this is forced because they are used in the sliders in the block UI. If they were put below 0, you would get to "unlimited" by dragging the slider to the right of 0, which would be terrible UI.

RossM:
You're welcome to fork this and try to improve the logic.

I wanted to avoid showing radar for things that already had HUD markers, but I couldn't get it working right - it's not easy to see what other HUD markers there will be from the radar code. You'd probably have to do it inside the HUD display code instead. I settled for drawing the radar markers under other markers so they're less distracting.

I don't show signature size because the signature sizes for voxel maps are highly inaccurate, and it seemed to work better not showing them at all than showing an inaccurate size.

Handling enemy objects different from allied objects seems weird - how does the radar know which ones are enemies?

The targets cache gets cleared every time Update is called (line 87). It's just there to avoid allocating a new list object every time. Object allocations kill performance.

"Broadcast Using Antennas" is the exact text used for ore detectors.

zrisher:
Yes, sorry I meant to get back to you and I'd love to work on this with you, pushing PRs to your branch. There are several other SE projects that are super high priority for me though first, but I should be able to work on this this weekend. KSH will probably get to this first and I'm sure they'll have plenty to say, so there's opportunity for me to help build this out down the line I'm sure. Like I said, very cool PR.

zrisher:
Sizes - why not use a separate class of sizes for asteroids? Is reported size at least reliably comparable between them?

RossM:
Asteroid sizes are based on the full size of the voxel map, which is either 100m, 200m, 400m, or 800m. The radius of the bounding sphere is slightly larger than half the voxel map's size. However the asteroid itself doesn't necessarily take up the full voxel map.

Ship sizes are based on the radius of the bounding sphere, which is roughly comparable to half the length of the longest dimension.

If you want to reveal size classes somehow, the best resolution you can reliably get from the bounding sphere is factors of two (such as 1m, 3m, 6m, 12m, 25m, 50m, 100m, 200m, 400m, 800m). I tried displaying that but it didn't feel right to me so I removed it.

I didn't want to report a separate size class for asteroids because I didn't want it to be too easy to tell the difference between a small asteroid and a large ship.

cbbetelgeux:
Hi, I am not a programmer, just looking at contributions, and this is an awesome idea, something most of us was waiting for.
If I may I would like to gave a bit of suggestion on the functionality. I think the radar module would be much more awesome if a separate 3d space map interface would be added by somebody in the future.
Until then, the logic behind the radar could be based on signal strength instead of range.
I will try to explain. The radar (code) would designate coordinates based on a fuzzy logic, not just perfect targets. Every designated coordinate has a signal strenght value ( like 1-255 ). This value would depend on a function of numerous factors like: the distance of the detected object, the size of the detected object (mass or cube block amount), the amount of power the radar is using (could be set on the radar terminal ui). And after that, the signal strength value could affect the opacity of the marker icon on the visual interface (3d space map), or the way-point icons. Also the lower the signal strength is, the radar marker coordinates could be miscalculated by a higher and higher random values from the actual object coordinates so they could appear further away from where they are actually.
So an electronic warfare module, like a jammer could play a good role, by lowering the signal strength value of the friendly ships/stations in its vicinity, when detected by an enemy radar.

zrisher:
I like the idea of displaying miscalculated positions for things with lower radar signatures. 

Putting this info on a 3d space map interface would require GUI functions, which aren't currently exposed to modders.

zrisher:
Just a thought on architecture - it seems like the way the game determines if you see a Beacon or not is by using `Sandbox.Game.Entities.MyRadioBroadcasters`, which `MyRadioBroadcaster` attaches itself to on creation an a `MyRadioReceiver` queries on update. It seems like a pretty efficient implementation, storing everything in an AABB tree. 

Those RadioReceivers are based on the more generic `MyDataReceiver`. Why not follow the game's model and create a new type of DataReceiver that holds a bit more info and gets stored in `MyRadarBroadcasters`? For things like hiding nearby grids, you could simply calculate the impact of each additional grid on its intersecting neighbors on the internal tree when it's added and moved.

Drakoskai:
Sounds cool. Would be neat to figure out some logic for the concept of size vs radar signature.  Since just because something is big, doesn't mean that it reflects radar back in proportion to it's size.  Some materials are less reflective than others and also, there are geometries that can reduce a signature.  I think the materials aspect could be interesting, all  you'd need to do is add a property to all blocks for it.  Geometries I think would be tricky.  Also consider Active vs Passive radar.  Pinging your active radar is going to light up the em spectrum and reveal you to anyone that is watching.

PR #72, opened 2015-05-18
Allow guns with infinite ammo.

RossM:
This patch makes it so setting the capacity of an ammo magazine to 0 gives it infinite ammo. It's not necessary to actually have a magazine in the gun's inventory, simply having a capacity-0 magazine as the magazine type will allow infinite fire.

This is intended to be used by mods that add laser weapons.


Spartan322:
@RossM SE doesn't have Raknet yet, so it actually relies on client data entirely, it doesn't really filter it at all (or much), so we have to build things that can't be done on the client only (yet)

zrisher:
@mexmer As an aside, could you point us to good hack-detection mods? Would love to use and contribute to that.

PR #73, opened 2015-05-18
Add MaxRange to SensorBlock Definition

KionX:


zrisher:
This is a continuation of #25.

PR #77, opened 2015-05-18
Add show block Mass and MaxIntegrity in Hud information

KionX:
valuable info


THDigi:
This sounds really useful, bump! xD

PR #78, opened 2015-05-18
Fixed #21 Oxygen 100% bottle disassembles indefinitely.

### LINKED ISSUE #21: Oxygen 100% bottle disassembles indefinitely.
Tracking down what could be the reason for this bug. It's one of the serious bugs that ruined a joy of survival mode.

MarcMann:
Is this code required in some case?
It prevents 100% full oxygen bottles to be disassembled correctly.
You gain back raw materials **and** the oxygen bottle.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Inventory/MyInventory.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Inventory/MyInventory.cs
@@ -594,17 +594,6 @@ namespace Sandbox.Game
 
                     MyInventoryItem item = src.m_items[i];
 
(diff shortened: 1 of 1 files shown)


Vxsote:
It appears that the effect of the code that is proposed for removal is to allow oxygen generators and tanks to pull only non-full bottles from other inventories.  However, it also has the direct side effect of breaking the removal of disassembled bottles.

While these proposed changes will fix the problem with disassembling bottles, it will break the logic for pulling only non-full bottles.  I have proposed an alternate fix in pull request #91.

OndrejPetrzilka:
Fixed in #91

PR #82, opened 2015-05-18
Reaction control system

Knsgf:
This is an implementation of automatic off-centre thrust compensation for gyroscopes, thruster-induced torque and thruster-based ship rotation controls.

All these features are optional. Disabling thruster damage in world settings switches them off.

A video demonstrating gyroscope stabiliser:
https://youtu.be/ULsee3ML9yw

First test of gyro-less flight controls:
https://youtu.be/bDBmuqUR1vg

Gyro-less autopilot demonstration:
https://youtu.be/1Vc0m3CpTlA

Notes/caveats:
a) Both gyroscope stabilisation and RCS require an operational remote control block with "Fly-by-wire" switch turned on. Without it controls revert to "gyros turn; thrusters move and couldn't care less about torque" scheme,
b) By default, all engines have their linear movement mode enabled and rotational assist mode disabled. Use thrusters' switches in the control panel to change that,
c) Turning inertia dampeners off and then trying to rotate a ship using thrusters also produces linear movement unless all rotational thrusters are balanced perfectly,
d) Turning while accelerating/braking is often (but not always) slower than at standstill. This is caused by engines trying to accelerate and turn simultaneously.


Geneticus:
Could RCS thrusters be split off as a different type that isn't affected by dampers? Then Normal Thrusters are for Linear motion only and have damping as-is  while RCS replace Gyros for turning and make them modify rotational dampers instead like an upgrade block.

RossM:
Nice idea, but it needs some more work. Continuous forward thrust without touching the mass caused my ship to start rolling. Even when thrust stopped, it kept rolling in place. (I was using the constructor ship from easy start A.)

I also think the controls are a bit weird - turns end too abruptly, and it's difficult to turn some ships while moving. The fighter from easy start A can't pitch upward at all while moving forward.

liorabel:
This is incredible. Exactly what I'm missing in Space Engineers!
For ages now I've wanted more realistic physics and the ability to have asymmetrical ships behave less like a block.

I will add these points though:
1. RCS thrusters are a great idea but I don't think they should be separate blocks. Instead there should be a setting on thrusters for whether they should participate in RCS behaviour or not.
2. Ships need a logical/realistic way of "sensing" when RCS adjustments are required. Today that is both "sensed" and adjusted by gyros. With thrusters being able to make adjustments as well, I think we could possibly have a second type of passive gyro that will only "sense" the ship's movement but not be able to correct it. This will be much cheaper to build and would be a requirement for RCS motion.

RossM:
Very nice! This version seems quite playable. Just some notes from a few minutes of playing with it.

The controls seem very springy, particularly turning. Understandable, it's an RCS system. Weirdly, moving the mouse just slightly at the end of a turn keeps the ship pointed in the final direction of the turn, while releasing the mouse entirely results in springing back towards the direction you turned from.

You can *almost* control a ship with no gyros at all. However, unless the thrust is perfectly balanced, it will tend to fly in circles. It's possible to correct it using the mouse - the RCS is very good at turning using only thrusters it seems - but it would be nice if it could hold a course even without gyros.

It would be nice if turning large ships had more inertia. This is a problem in the base game too. It's kind of annoying to be trying to turn a large ship, reach the end of the mousepad, pick up the mouse to move it, and have the ship's thrusters fire and completely stop the turning momentum you managed to get going.

PassivePicasso:
A way to make it easier to balance the RCS thrust would be to make it so thrusters can have their Max Thrust set.  I could see this getting pretty involved but it would allow you to slam a bunch of rear thrusts in an Asymmetrical setup, and still achieve stable forward thrust

ArghArgh200:
I like this idea. Maybe have it usable as a per-ship option?

Knsgf:
RossM,
The version 3 was "quite playable" solely due to a subtle but serious bug, which produced phantom braking forces. With this bug squished, all stock small ships -- none of which are balanced very well -- cannot fly properly without gyroscopes, especially when drifting with dampeners on.

PR #91, opened 2015-05-19
O2 bottle infinite disassembly - improved implementation.

Vxsote:
This is an alternate fix to the infinite O2 bottle disassembly problem.  The proposed changes in pull request #78 will fix the disassembly problem, but leave the bottle pulls from O2 tanks and generators working incorrectly.

This version loops through the inventory an extra time if an oxygen container is specified - the first time ignoring full bottles (as the current code does), and the second time (if an amount to transfer remains) taking any bottles.


OndrejPetrzilka:
This is hack to fix another hack. It will be eventually removed.

PR #93, opened 2015-05-19
Add MaxRange to SensorBlock Definition

KionX:


OndrejPetrzilka:
Merged and made some cleanup.
1) MaxRange is not stored in block, always used value from definitions
2) Added default value into sensor definition object builder

aaraujo666:
I did something similar in my copy of the source. However I did not go as far as making it come from the definition file. Focused more on not letting it get over powered by just artificially increasing the range.

Changed CalculateRequiredPowerInput in Sandbox.Game.Game.Entities.Blocks.MySensorBlock.cs:

protected float CalculateRequiredPowerInput()
        {
            if (Enabled && IsFunctional)
            {
                float sensorVolume = (m_fieldMax - m_fieldMin).Volume;
                float factor = (((float)Math.Floor(Math.Abs(m_fieldMax.X) / 50f + 0.5f) * 0.5f) / 10000f) +
                               (((float)Math.Floor(Math.Abs(m_fieldMax.Y) / 50f + 0.5f) * 0.5f) / 10000f) +
                               (((float)Math.Floor(Math.Abs(m_fieldMax.Z) / 50f + 0.5f) * 0.5f) / 10000f) +
                               (((float)Math.Floor(Math.Abs(m_fieldMin.X) / 50f + 0.5f) * 0.5f) / 10000f) +
                               (((float)Math.Floor(Math.Abs(m_fieldMin.Y) / 50f + 0.5f) * 0.5f) / 10000f) +
                               (((float)Math.Floor(Math.Abs(m_fieldMin.Z) / 50f + 0.5f) * 0.5f) / 10000f);
                
               return factor * (float)Math.Pow(sensorVolume, 1f / 3f);
            }
            else
                return 0.0f;
        }

At vanilla max of 50m, power consumption is the same. Every increment of 50m, power consumption increases.

PR #95, opened 2015-05-19
Drill push items if volume >= 70%

KionX:
reduces load on server and network


MeusRex:
What exactly is the use of this? The push request happens only every 1.5 - 2 seconds. In that time you will easily fill the last 10% of the drills inventory. You will just waste ressources. Also, will the drill continue to empty into a cargo container if it is set up like this? And finnaly, the main problem with the drill is the manipulation of the voxelmap and not the inventory.

No-Brain:
No description provided.
as @MeusRex said, there is no reason to buffer the ore in the drills instead of sending it to cargo containers as it only introduces wasting of resources.

KionX:
reduces load on server and less network packets.

nothing wasting of resources.

PR #99, opened 2015-05-19
Fix Drill inventory size and volume

KionX:


No-Brain:
No description provided.
Increases inventory sizes for no reason.
Defeats the purpose of inventory size in definition.
Don't do pull request from master branch.

KionX:
/No description provided./
  now inventory size is calculated as well as for cargo containers.

/Increases inventory sizes for no reason./ 
  does not increase but reduces the size.

/Defeats the purpose of inventory size in definition./
  no definition of inventory size and never was.

/Don't do pull request from master branch./
  where I want from there and make.

PR #101, opened 2015-05-19
Interior Light HasPhysics moved to Definition

Eikester:
Currently we are limited to flat designs if we dont want to let Players
fly trough our Models. This change gives us more options.

An example for when its useful to have Collision enabled:
http://i.imgur.com/ya6zryB.jpg

Default value = false


THDigi:
Wouldn't this be useful for all blocks actually ? Only that it would be default false for interior lights and true for everything else.

Eikester:
Yes it would be but i think it would be much better if we could define what gets blocked (characters, ammo, ships etc.) (i.e. Force Field), idealy even per subpart, i may look into this further

Issue #96, opened 2015-05-19
Assembler queue access with ingame script


mexmer:
One of this ingame PB pb is missing is access to assembler queue.

It seems not hard to make current queue visible, but i'm not sure if it will be safe (well i'm pretty sure, it will not), therefore queue will need to be accessed either by functions (addtoqueue, removefrom queue), or using interface (IMyQueue), like is for IMyInvetory.

I'm more inclined to IMyQueue (or whatever we will call it) as safe access to queue

ideas welcome

mrsib:
I was thinking it didn't need to be too complicated. Just add items using their type/subtype name/amount and a function to clear the queue:
`myAssembler.QueueItem("MyObjectBuilder_Component", "SmallTube", 100));
myAssembler.ClearQueue();`

MMaster:
I think PB also needs access to read the queue so scripts can figure out what is being produced. So IMyQueue approach would be probably better.

mexmer:
@MMaster that's my thought about it. I want safe exposure of queue, so we can do real scripts for watching production and queuing production.

MMaster:
@mexmer Definitely see your point - it's really needed for production automation scripts. I am really happy that there is so great modding community around SE. Half of the things I wanted to address already have pull requests or issues opened :)

mrsib:
@mexmer @MMaster Yes I think it would be good to have access to the queue, I have created a merge for the basic functioned I mentioned as it was very easy to do.
The problem with the Queue is that it uses MyBlueprintDefinitionBase which isn't really exposed to InGame programming. 
What I am thinking is a copy of the Queue where instead of MyBlueprintDefinitionBase, there is type and subtypename. This would mean defining a struct (exposed to InGame), and returning a list with that struct:
`GetQueue(); // returns List<AssemblerQueueItem>`
Then queue removal would be easier, just expose a function to remove at index. i.e. `RemoveQueueItem(int itemIndex);`

mexmer:
@mrsib problem with this list is, that is unsafe. i was already doing some stuff with assembler in my local branch, and there are lot of issues involving sync and safe operation, therefore i would not be hasty with implementing proprietary functions, that might get removed soon (more likely will be obsoleted)

mrsib:
@mexmer Ok, 
Can you help me understand why the list idea is unsafe and what sync issues it could cause.
Also, other than failure to use if checks in my previous commit, was there any issues with safety/sync?
[...]

PR #102, opened 2015-05-19
#96 Added Ingame Assembler Functions

mrsib:
Added Ingame Assembler functions QueueItem and ClearQueue to enable
players to be able to queue items in the assembler and clear the queue.
QueueItem has three parameters, string itemType (I.e.
MyObjectBuilder_Component), string subtypeName(I.e. SteelPlate), int
amount
QueueItem returns bool: true if item was queued, false if not. False
indicates invalid type or subtype name.

ClearItem has no parameters

Both functions work in either assemble mode or disassemble mode.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyAssembler.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyAssembler.cs
@@ -1065,5 +1065,22 @@ namespace Sandbox.Game.Entities.Cube
         {
             return base.GetOperationalPowerConsumption() * (1f + UpgradeValues["Productivity"]) * (1f / UpgradeValues["PowerEfficiency"]);
         }
+
+        bool Sandbox.ModAPI.Ingame.IMyAssembler.QueueItem(string itemType, string subtypeName, int amount)
+        {
+            try
+            {
+                InsertQueueItemRequest(-1,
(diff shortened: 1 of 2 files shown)


Sources/Sandbox.Game/Game/Entities/Blocks/MyAssembler.cs
```diff
@@ -1065,5 +1065,22 @@ protected override float GetOperationalPowerConsumption()
...
+
+        bool Sandbox.ModAPI.Ingame.IMyAssembler.QueueItem(string itemType, string subtypeName, int amount)
+        {
+            try
+            {
+                InsertQueueItemRequest(-1,
+                                       MyDefinitionManager.Static.TryGetBlueprintDefinitionByResultId(new MyDefinitionId(MyObjectBuilderType.Parse(itemType), subtypeName)),
+                                       (VRage.MyFixedPoint)amount);
+            }
+            catch { return false; } // User probably entered invalid type or name
```
OndrejPetrzilka: I would prefer something else than this. Ideally if checks.

mrsib: Understood, will look into further.

mrsib:
Closing for further alterations, will make new pull request when ready.

mexmer:
since issue is dicussed here #96 i would like to see better approach to this issue.

PR #104, opened 2015-05-19
Ship shield

yajiedesign:
Ship shield,not Complete,Collect opinion.
Reference http://steamcommunity.com/sharedfiles/filedetails/?id=399121944


No-Brain:
Why do you create pull request for incomplete things?

THDigi:
Maybe make the fundamentals for it to be more possible in modding, but not as vanilla, it doesn't fit with the SE lore in my opinion xD

PR #105, opened 2015-05-19
Allow editable mission screen in ModAPI

Spcemarine:
- added the argument editEnabled to the API methods
- added a parameter that contains the text of the mission screen to the
  callback

Note: This might be a breaking change for some scripts because it modifies the callback Action but I'd find it quite useful.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMyUtilities.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyUtilities.cs
@@ -36,7 +36,8 @@ namespace Sandbox.ModAPI
         string SerializeToXML<T>(T objToSerialize);
         T SerializeFromXML<T>(string buffer);
         void InvokeOnGameThread(Action action);
-        void ShowMissionScreen(string screenTitle = null, string currentObjectivePrefix = null, string currentObjective = null, string screenDescription = null, Action<ResultEnum, string> callback = null, string okButtonCaption = null, bool editEnabled = false);
+        void ShowMissionScreen(string screenTitle = null, string currentObjectivePrefix = null, string currentObjective = null, string screenDescription = null, Action<ResultEnum> callback = null, string okButtonCaption = null);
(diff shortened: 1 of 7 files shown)


Tyrsis:
Needs to be backwards compatible.  A lot of mods use this screen as a way to display information.  Just make an editable mission screen something completely or maybe try to figure out a way of overloading the current method.

Spcemarine:
As long as you haven't specified a callback action (what is supposed to be the case when you just display information) it is compatible.
The current solution is the easiest and the most logical for the API (at least in my opinion). Creating a second method would be weird because it would be some kind of duplication. Furthermore a lot of mods use methods or interfaces that change sometimes. Our mod breaks approximately every two versions, so I'm not really motivated to change that as long as someone from KSWH asks me to do it.

Tyrsis:
So you expect everyone to fix their mod because you would find the feature useful?  That isn't playing nice.  But you're right, let KSWH decide.  Though look at other pull requests that do not provide backwards compatibility, and you'll notice a trend that they tend not to merge those.

Spcemarine:
Well, 'everyone' is a little hard, don't you think so? It's only the case if someone really uses the callback, all other mods won't be affected. However, I wouldn't mind if there never are breaking changes but I don't get double standards.

MeusRex:
From what I could glean from the Asteroid generator PR, PR's are required to be backwards compatible. https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/58

Spcemarine:
But that is about the main game, not modding stuff, isn't it?

MeusRex:
Hmm, no idea. I guess we'll have to wait for a dev to comment.

Spcemarine:
Yes. That makes me wondering if the devs are going to release some sort of guidelines. Would be good to know what we are allowed to do and what we shouldn't.

MeusRex:
Idd. Would be quite useful.

No-Brain:
Everything should be backwards compatible. Especially when there is no problem to do it without hacks like in this case.

Spcemarine:
Ok, then I'll just make sth like ShowEditableMissionScreen(...).

Issue #74, opened 2015-05-18
Idea about upgrading LCDs


stafyniaksacha:
There is some idea for improving LCD screens : 
- Able to choose a monospacing font
- Able to set colors (background/foreground) of each characters (and not only screen)
- Access to the display matrix of the screen

kapitanov:
> Able to choose a monospacing font

+1, monospace font is a must have feature

MMaster:
I would prefer moddable font. I was already looking into the code, but unfortunately LCD does not have font property at all. Render function does not take font as argument, but I did not look into it any deeper. If we would be able to specify font per LCD definition we would be able to do much more than just monospace font.

kapitanov:
Looks like the actual text rendering takes place [here](https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/07519235308594be64a1a96254897942206899bf/Sources/VRage.Render/MyDebugDraw.cs#L809), but unfortunately the font is hardcoded right there to `MyRender.GetDebugFont()` value.

The rendering parameters go all the way down:
* `MyTextPanel`,
* `MyRenderComponentTextPanel`,
* `MyRenderProxy`
* `MyRenderMessageDrawTextToMaterial`
* `MyRender`
* `MyDebugDraw`

And the font pops up only on the very last stage of pipeline. However it looks like an easy job to pass an optional `MyFont` object though the call chain.

Access to the display matrix should be easy too since the text buffer is stored right in `MyTextPanel`. A few lines of code should be enough to help represent the text buffer as a matrix. However display matrix access doesn't make much sense to me until we have a monospace font.

Per-character color attributes will definitely take a lot of work and might kill the performance. It would be an awesome feature anyway - imagine Norton Commander-style interfaces made inside SE game :)

cwc:
For reference, game fonts are currently hardcoded in [MySandboxGame](https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/a41226a238e66b66cc2ef4461982d769be126ba1/Sources/Sandbox.Game/MySandboxGame.cs#L1039-L1059). Perhaps the font descriptions could be loaded from a file?

kapitanov:
OK, I've got a working prototype that allows you to select normal or monospace font. But things are not as nice as they might be:

[...]

PR #107, opened 2015-05-19
Issue #74 Moddable custom fonts for LCDs

kapitanov:
As it had been discussed in issue #74, I've added support to load custom fonts for LCD screens from workshop mods.
An example of custom font mod can be found [here](https://github.com/kapitanov/SE_CustomFonts/tree/master/ModExample)
I've added `Font` property to `IMyTextPanel`, the corresponding GUI to select custom font and an API for PB and Scripts.

I should note that my implementation loads the fonts on demand in the render thread and might be not the best idea.

Also, custom fonts affect only LCDs, there's no code to apply them to any other ingame text, including GUI.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyTextPanel.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyTextPanel.cs
@@ -7,6 +7,10 @@ namespace Sandbox.ModAPI.Ingame
 {
     public interface IMyTextPanel : IMyFunctionalBlock
     {
+        string Font { get; }
+        void GetFonts(List<string> ids);
+        void SetFont(string id);
+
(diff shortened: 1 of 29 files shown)


Phoenix84:
I second this.
I looked at doing this because I wanted a monospaced font.
However it looked like it was already in progress, so I held back.
Glad to see someone else wants it too.

JustBurn:
Wish i knew about this earlier... it could had saved me the whole work into hacking MyRenderFont.DrawString() to generate monospaced text when "Monospace" in TextPanel is checked :weary:

On top of that this does load custom fonts too, really awesome! :+1:

Now only if you add a list of all available fonts into a Combobox on TextPanel this PR should be perfect :smile:

PR #108, opened 2015-05-20
Allow attaching MyGameLogicComponent to specific subtype of a block type

Tyrsis:
This allows mods to attach MyGameLogicComponent to specific subtypes of a block type instead of just the block type only (and causing collisions / overwrites if multiple mods want to attach to a block type). 

This adds a new attribute [MyEntitySubIdComponentDescriptor] which is a text attribute that matches the subtypename of the block you want to attach the gamelogic to.  You can specify multiple attributes to attach to multiple subtypes (large, small for example). 

Is backwards compatible, and defaults to old behavior if new attribute is not found.


THDigi:
This still won't fix the issue where two mods need to hook on the same object type or even subtype.

For example a gravity generator upgrade module mod would collide with a mod that wants to add something to the gravity generators as well, that's where subtypeid would be irelevant.

Tyrsis:
Yes it doesn't fix that problem, but it does allow mod makers to isolate their scripts from one another.  If you attach gamelogic using this, it will override gamelogic attached the old way.

I think they are two separate problems.  This fixes one of them. I am still experimenting with fixing the issue you bring up, or at least addressing it.

THDigi:
They're 2 halves of the same problem :package: 

At least change the notice to ErrorSeverity.Error.

Tyrsis:
Closing this because I did a PR from master instead of a branch

PR #109, opened 2015-05-20
Ship shield

yajiedesign:
Preliminary completion,
Collect opinion.
Reference http://steamcommunity.com/sharedfiles/filedetails/?id=399121944


Spcemarine:
It would be better to expand the modding abilities to get this kind of mod working than to integrate this into the main game. I can't speak for KSWH but I think the chances are rather low that this pull request will be accepted.

MeusRex:
I agree with Spcemarine. This should be a toolbox for modders instead of a finished solution.

I for one dislike this implementation of shields. You could just as well increase the hp of all blocks, that would achieve exactly the same.
I'd rather have partial fields, floating at an offset to the emitter and creating an umbrella over the emitter and adjecent blocks. Your implementation doesn't allow for any of this, so should I just create my own class and make a pr? And should the ten other guys with their own visions for shields?
I don't think so. Therefore a discussion is necessary for planning a decent shield interface.

Spcemarine:
I think an interface dedicated to shields would be a waste of efforts. I'd rather se interfaces for creating custom block's behavior.

yajiedesign:
shield
like this?

MeusRex:
Jep, that's it. This way good placement would be imperative. Also, it would make small ships more interesting to design and more useful in combat against large vessels, as a good pilot could hit the space directly between shield and hull if the shield isn't flush to the hull.

zrisher:
That's a really exciting concept, would love to see this opened up for implementation.

thelastWallE:
http://steamcommunity.com/sharedfiles/filedetails/?id=326282818
If these work, then you can go from there and add some more functions to these blocks with more api.
Someone mentioned to use the door code from the door block to have the shields pop up.

OndrejPetrzilka:
Hi, door code is really not good example. We would like to rewrite it at some point. There's issues with door collisions, especially on moving ship and ships vs ships.

Better would be to just add some more collision shapes to block when shield is up and refresh physics.

PR #111, opened 2015-05-20
Ore position dictionary for ingame programming

plaYer2k:
Added a simple functionality to get the detected ores position as
Dictionary<string, Vector3D>. Thus each ore yields only the closest
position in any asteroid per OreDetector.

Example output from the below world and code:
http://i.imgur.com/Q2cGJrR.png

To test this feature use the workshop world:
http://steamcommunity.com/sharedfiles/filedetails/?id=445920958

or build a ship and use this example code ingame:

```C#
void Main(string args)
{
string output = "";
List<IMyTerminalBlock> blocks = new List<IMyTerminalBlock>();
GridTerminalSystem.GetBlocksOfType<IMyOreDetector>(blocks);
output += string.Format(">>> {0} OreDetectors\n", blocks.Count);
for(int i = 0; i < blocks.Count; i++)
{
IMyOreDetector oreDet = blocks[i] as IMyOreDetector;
output += string.Format("\n>> {0} ({1} ores)\n", oreDet.CustomName,
oreDet.DetectedOres.Count);
var enumerator = oreDet.DetectedOres.GetEnumerator();
while(enumerator.MoveNext())
{
var element = enumerator.Current;
output += string.Format("> {0}\n{1}\n", element.Key,
element.Value.ToString("{0.00}"));
}
}
Echo(output);
}
```

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyOreDetector.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyOreDetector.cs
@@ -9,5 +9,6 @@ namespace Sandbox.ModAPI.Ingame
     {
         float Range {get;}
         bool BroadcastUsingAntennas {get;}
+        Dictionary<string, VRageMath.Vector3D> DetectedOres { get; }
     }
 }
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyOreDetector.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyOreDetector.cs
@@ -12,6 +12,7 @@ using VRageMath;
 using Sandbox.Game.Components;
 using Sandbox.ModAPI.Ingame;
 using Sandbox.Game.Localization;
+using System.Collections.Generic;
 
 #endregion
 
@@ -184,5 +185,6 @@ namespace Sandbox.Game.Entities.Cube
         }
         bool IMyOreDetector.BroadcastUsingAntennas { get { return m_oreDetectorComponent.BroadcastUsingAntennas; } }
         float IMyOreDetector.Range { get { return Range; } }
+        Dictionary<string, Vector3D> IMyOreDetector.DetectedOres { get { return m_oreDetectorComponent.detectedOres; } }
     }
 }
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyOreDetectorComponent.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyOreDetectorComponent.cs
@@ -191,6 +191,8 @@ namespace Sandbox.Game.Entities.Cube
 
         private readonly Dictionary<MyVoxelBase, MyOreDepositGroup> m_depositGroupsByEntity = new Dictionary<MyVoxelBase, MyOreDepositGroup>();
 
+        public Dictionary<string, Vector3D> detectedOres = new Dictionary<string, Vector3D>();
+
         public MyOreDetectorComponent()
         {
             DetectionRadius = 50;
@@ -203,6 +205,7 @@ namespace Sandbox.Game.Entities.Cube
         public void Update(Vector3D position, bool checkControl = true)
         {
             Clear();
+            detectedOres.Clear();
 
             if (!SetRelayedRequest && checkControl && !OnCheckControl())
             {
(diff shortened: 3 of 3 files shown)


mexmer:
wanted to add this, but i was not satisfied with output, because it was missing yield (Amount of stored ore) ... so i postponed this until later.

tho' i believe even without yield it's usefull to have this. you can for example create drone, that will mark the spots on gps for you.

plaYer2k:
Yes i thought about the total ore amount per deposit aswell but i abandoned that idea again as it seems to have little use.

I find the idea to get only the closest ore kinda appealing as that still encourages a certain kind of exploration around the deposit in order to "properly meassure" it.

Furthermore my previous attempt to use a List<Class> led to some very weird issues where iterating over the List ingame resulted in OutOfBound exceptions and false informations within the read list and i couldnt figure out why for several days, thus i am glad the dictionary attempt works as well as it does and it also should be more performant than a straight bloaty list of instances.

mrsib:
:+1:

mexmer:
@plaYer2k list will not work ... ore list is updated every 100 frames (if you have 60fps that means in less than 2 sec), so if you iterate slow enough, your reference will become invalid (or rather will point to modified list), only option is to make copy, or create new container.

KionX:
detection ore only works for player controlled ships
if you do a detection for all ships possible a strong decrease performance

oreDet.CustomName... another type of ore need

plaYer2k:
Sorry but i dont understand your point there @mexmer.
The List was updated in the same method like the Dictionary. So in both cases the datas are equally up-to-date.

The main difference however is that the garbage collector has much less to do as the dictionary takes only one ore-position-pair while the list took every. To put that in numbers, in that above example world i previously had 162 entries within the list if List.Count didnt lie (even though i ran into the OutOfBound exception when iterating through the list for 0 < i < List.Count) while the dictionary version has two, thus a factor of 80.
Overall though, instead of a potential maximum of scanvolume_in_m³  / 15.625 ores, considering there is only one ore per asteroid deposit position, they are spread out in 2.5m steps aswell as no asteroids overlap, which for the current spherical detection with r=50m results in up to 33510 ore positions. That is of course highly theoretical and will never be reached practically as that would require the ore detector to sit within an asteroid full of ores.
However the amount compared to the dictionary that caps at the max amount of possible ores ingame (255 max but currently around 13 i think).


@KionX 
"oreDet.CustomName... another type of ore need"
You seem to refer to the example for an ingame script there.
I used that to form a string output for the Echo command. The actual source code does not use the CustomName at all.
So please keep in mind that the text you directly see there is only the ingame script example to demonstrate how to access the information.

mexmer:
@plaYer2k ah, you copied list into another instance, i thought you exposed it direcly ... anyway, i think there is some problem with list interface in general in K#

THDigi:
Wouldn't this dictionary setup prevent you to get multiple pockets of the same ore ?
I think it should be ``Dictionary<Vector3D, string>``.

plaYer2k:
@THDigi
This would indeed only get one pocket of ore.
There were two reasons for why i think that this is the best method so far and i named them above already but i will again.

The first is that the worst case is much better. You got a tightly limited capacity of ores to store and for the garbage collector to dispose each update with a oretype-key over the position-key because the position as key has no difference to a straight list that gets all ores in range. So instead of potentially bloating the collection with up to ~33k entries (see three posts above), we limit it to a very small and fixed amount of ~13 (or whatever the max amount of ores is right now, didnt check).

The second is that it still encourages exploration, moving and positioning.
Instead of just "touching" an ore with your ore detector and perfectly knowing where all small ore chunks are within reach, you only detect the closest and thus still have to circle around an asteroid to "properly meassure that asteroids yield". Thus it also removes a kind of "god-element" by limiting the completeness of information the user can get without further ado.

Thus it might has the same reason behind it like the Sensor has by only returning the LastDetectedEntity instead of all in range.

The note about the DistanceSquard is very nice and i added it to my current code, thanks.
@lord-devious noted about the access restriction for the array and that he thinks passing the dictionary directly should be wrapped in a ReadOnly manner, where i agree after him mentioning it.
Thus i will update the pull request when i finally got myself up to change it there.

THDigi:
Hmm, actually, will this code give you the exact same data as the HUD markers for ores that you see while in the ship ? That's all I'm asking for, no more and no less :P

Also, Clear() is called from multiple places, might be more useful to move ``detectedOres.Clear();`` in Clear() instead...

Or you could add the data to the list/dictionary in the MyGuiScreenHudSpace class since the calculations are already done there and it would be an exact replica of what you see.

midspace:
Any particular reason this was abandoned?
Couldn't the Dictionary be changed to a on demand method instead to reduce overhead, and use the DictionaryReader<> instead?

plaYer2k:
The reason this was abandoned apparently is an automatic one. I did not manually "close" it though i deleted my old repo because i simply am net to git and merged nonsense into the master branch.
Merging the new changes into my fork wasnt possible that way, at least i didnt figure a good way, and thus deleting the old fork as a whole and re-forking was much easier.

I will however, when i am home in a week or so, try my hands on a good other method that is readOnly so that i dont have to pass the whole dictionary to expose it,

If anyone else wants to release that in the meantime, go for it, i wont be back that soon as i am very busy here aswell as i dont have access to my desktop to run and test the changes.

PR #112, opened 2015-05-20
Fixed PermanentDeath/EnableCopyPaste settings resetting

Sibz:
PermanentDeath/EnableCopyPaste where both defaulting to On/Checked during load due to the enabling of the selected game mode. 
I.e. On creative mode, the EnableCopyPaste checkbox was enabled and checked and PermanentDeath disabled and unchecked. 
On survival mode the PermanentDeath checkbox was enabled and checked and EnableCopyPaste disabled and unchecked. 

Now it will enable and only be checked if it is true in the settings. It still disables and unchecks.


OndrejPetrzilka:
Shouldn't there be null check (or HasValue check)?
m_selectedSessionSettings.PermanentDeath.Value;

...because of compatibility with older saves

Sibz:
@OndrejPetrzilka Yes, sorry, an oversight on my part. Have added the required check and tested on save without PermanentDeath in there, Defaults to True/On in this case.

Sibz:
Has been resolved

PR #113, opened 2015-05-20
Ships can pass through closed Doors Fix

Eikester:


No-Brain:
This only works on stations.
Ship would tear itself apart.

PR #115, opened 2015-05-20
Drill push items if volume >= slider value(0-100%)

KionX:
allows you to use the drill as a container for the storage of ore
eliminates spam ore(useful for autosorting mods)
optimization


JJRcop:
You should instead force pull from upstream to your master, make the branch again, and start over. To avoid the junk commits that add something then revert it.

mexmer:
also use rebase instead of merge when pulling from remote upstream (keen source)

No-Brain:
No description provided.
Why do we need this?

KionX:
allows you to use the drill as a container for the storage of ore
eliminates spam ore(useful for autosort mods)

No-Brain:
Drills are already used as storage when there is no place to put the ore.
This is not an optimization in any way shape or form as this has even worse performance from additional game logic and extra bandwidth requirements.

KionX:
I watched the code. and I see that every time there is sending packets with information about the shifting of ore

sometimes it is necessary that the ore remained in the drill. so as not to interfere other items

No-Brain:
The only thing this does is delay the moving by 100 frames. It is exactly the same from that point.
This also wastes resources when the drill is full.

KionX:
this idea comes from my personal experience.
I find it useful. and the game loses nothing from this option.

No-Brain:
This doesn't improve anything. Adds no depth to the gameplay, just more complexity and ways to let the user shoot themselves in the foot.

KionX:
and I think that the upgrade modules for the refinery does not add depth.
you are too categorical.

KionX:
if you keep the ore in a drill you can save on containers into survival

mexmer:
if you don't have container, ore is not pushed out, but stored in drill even now.

when i use my drill drone, i always fill cargo, connector and drill to full, until i head back, if i remove cargo, i can build "smaller" drone, but i don't see benefit in that.

KionX:
I do not mean it. ore store in drill and the rest in other containers.
this is useful when resources are limited or when digging is not important.

JJRcop:
@KionX Your master is tainted, so any branch you make is not based off of master and includes junk commits. You should fix that by inputting these git commands on your local repo
```
git checkout master
git fetch https://github.com/KeenSoftwareHouse/SpaceEngineers.git master
git reset --hard FETCH_HEAD
git push --force origin master
```

**WARNING: THIS WILL DELETE ALL OF YOUR UNMERGED COMMITS ON THE MASTER BRANCH**

Copy-pastable version:
```git checkout master;git fetch https://github.com/KeenSoftwareHouse/SpaceEngineers.git master;git reset --hard FETCH_HEAD;git push --force origin master```

After this, just make new branches when you want to develop another feature, and do not commit to master.

PR #116, opened 2015-05-20
Add MemoryLimits checkbox to OptionsGame

KionX:
most people don't even know about this option
now it is well illustrated and easy to switch


lord-devious:
Something is weird with your PRs. How come all these old commits show up?

mexmer:
because of wrong branching

lord-devious:
Definitely seems like it.

No-Brain:
This should not be a user facing option as it can cause crashes when disabled.
It is already possible to disable this from config.

KionX:
most people don't even know about this option
now it is well illustrated and easy to switch

MeusRex:
And even less people will know what the consequences are... Which will propably result in a wave of "bug" reports.

No-Brain:
This only creates problems and solves nothing.

KionX:
I enjoyed it a lot. noticed nothing wrong. much worse when you suddenly can't put a block

No-Brain:
This only happens on 32bit systems. Windows supports 64bit since 2001. You should really consider upgrading.

KionX:
but this is 32 bit version

Spartan322:
This doesn't appear to add anything, seems only to be made to crash

PR #120, opened 2015-05-21
Added Height Offset for Suspension Wheels

THDigi:
Added a "Height Offset" option for MotorSuspension blocks which allows players to change the height of the wheel relative to its base, by default it's 0 which is in the middle, just like before.

This is useful for vehicles to travel better on very rough terrain having a bigger clearence under them and it can also be useful for racing vehicles that need a lower base.

Additionally MinHeight and MaxHeight parameters for cubeblocks to define the height slider range.

EDIT:
I plan on adding Max Steering Angle, Steer Speed, Steer Return Speed and Invert Steer Direction to the control panel as well but first I wanna see how this is received :}

### DIFF
--- a/Sources/SpaceEngineers/Content/Data/CubeBlocks.sbc
+++ b/Sources/SpaceEngineers/Content/Data/CubeBlocks.sbc
@@ -18066,6 +18066,8 @@
       <MaxForceMagnitude>3.36E+07</MaxForceMagnitude>
       <RotorPart>RealWheel</RotorPart>
       <SuspensionLimit>0.7</SuspensionLimit>
+      <MinHeight>-1.5</MinHeight>
+      <MaxHeight>1.3</MaxHeight>
       <DamageEffectId>212</DamageEffectId>
     </Definition>
     <Definition xsi:type="MyObjectBuilder_MotorSuspensionDefinition">
@@ -18110,6 +18112,8 @@
       <MaxForceMagnitude>3.36E+07</MaxForceMagnitude>
       <RotorPart>RealWheel5x5</RotorPart>
       <SuspensionLimit>2.5</SuspensionLimit>
+      <MinHeight>-1.5</MinHeight>
+      <MaxHeight>1.3</MaxHeight>
       <DamageEffectId>212</DamageEffectId>
     </Definition>
     <Definition xsi:type="MyObjectBuilder_MotorSuspensionDefinition">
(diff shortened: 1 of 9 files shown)


Phannd:
Retested. Rev. 55 438. Found a bug. 

1. build a small grid car (ship)
2. place 1x1 suspension wheel on it
3. steer the car forward (you can try also other directions)
4. change offset of the wheel to 0.26 it will start vibrating and shaking

note: other wheels are fine so I consider this a bug

Sibz:
@Phannd @THDigi 
If I'm not mistaken, this could be line 18161 in Sources/SpaceEngineers/Content/Data/CubeBlocks.sbc

THDigi:
@Phannd 
Weird, it seems to do that even if I move slowly, and the higher it is the more it shakes... and I have no idea what causes it :/ The largeship 1x1 wheels have the same problem.

While messing with that, found some more issues related to the 1x1 wheels:
It seems that the handbrake has no effect on the 1x1 smallship wheels (on largeship that works fine) and the largeship wheels can't provide power while turning, I think they collide with the body or something.

@Sibz 
Not sure what you mean, that line points to `<RotorPart>RealWheel1x1</RotorPart>` of Suspension1x1 (largeship), what about it ?

Sources/SpaceEngineers/Content/Data/CubeBlocks.sbc
```diff
@@ -18156,6 +18160,8 @@
       <MaxForceMagnitude>3.36E+07</MaxForceMagnitude>
       <RotorPart>RealWheel1x1</RotorPart>
       <SuspensionLimit>0.6</SuspensionLimit>
+      <MinHeight>-1.5</MinHeight>
+      <MaxHeight>1.3</MaxHeight>
       <PropulsionForce>700</PropulsionForce>
       <DamageEffectId>212</DamageEffectId>
     </Definition>
```
Sibz: This line here, This closes Definition but is also closed again below.

THDigi: The definitions are longer than you see them there, look at the full file.

Sibz: Ahh, I thought the 'Expand' would have shown me everything in between, but obviously not.

THDigi: I see, indeed, that's either broken or misleading :/ you could report that to github.

PR #123, opened 2015-05-21
Upgrade module effect values in block interface for assembler and reffinery

mexmer:
Direct field accessors for upgrade modules on affected blocks (refinery, assembler) for ingame scripting, so you can see for example productivity, without parsing it from detailed info.

might be usefull for example for task prioritization when using refinery/assembler management scripts.


Phoenix84:
Can you expose it from the upgrade BlockDefinition directly? This would allow upgrades provided by mods to be supported by this as well.

mexmer:
that's another thing i plan to do ... as well as interface for upgrade modules.
but waiting for this to be merged first.

generic upgradevalues will need little different approach, since they are list. also upgrade module will need better interface.

mexmer:
closed in favor of #190

PR #124, opened 2015-05-21
Add checkbox for auto push items from drill

KionX:
This not adds more depth to the gameplay but sometimes useful.
This allows you to use the drill for storage of ore without the use of other containers.
Please merge it. Make this request last time.


lord-devious:
@KionX You *need* to fix your branching method. This is very confusing.

PR #127, opened 2015-05-21
Changed power consumption to vary according to max range of sensor

aaraujo666:
This change increases the power consumed by the sensor based on the range being used.

With vanilla max of 50m, nothing changes.

With every increment of 50m that you set any of the extents, above the vanilla 50m max, power consumption increases.

Example: with all extents set to 50m, power consumption doesn't change.

If you were to increase the maxrange to 100m, AND set the extents all to 100m, then power consumption would be double that of the vanilla max of 50m. If you set all the extents to 50m (even when maxrange is set to 100m) power consumption is the same as vanilla 50m max.

In other words, it increases power consumption on range USED. So regardless of where the max range is set, power consumption only increases as you actually set the extents above the norm.

Also increased number of toolbars to 9, since that seems to be a popular request among players.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MySensorBlock.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MySensorBlock.cs
@@ -678,7 +678,17 @@ namespace Sandbox.Game.Entities.Blocks
         protected float CalculateRequiredPowerInput()
         {
             if (Enabled && IsFunctional)
-                return 0.0003f * (float)Math.Pow((m_fieldMax - m_fieldMin).Volume, 1f / 3f);
+            {
+                float sensorVolume = (m_fieldMax - m_fieldMin).Volume;
+                float factor = (((float)Math.Floor(Math.Abs(m_fieldMax.X) / 50f + 0.5f) * 0.5f) / 10000f) +
+                               (((float)Math.Floor(Math.Abs(m_fieldMax.Y) / 50f + 0.5f) * 0.5f) / 10000f) +
+                               (((float)Math.Floor(Math.Abs(m_fieldMax.Z) / 50f + 0.5f) * 0.5f) / 10000f) +
+                               (((float)Math.Floor(Math.Abs(m_fieldMin.X) / 50f + 0.5f) * 0.5f) / 10000f) +
+                               (((float)Math.Floor(Math.Abs(m_fieldMin.Y) / 50f + 0.5f) * 0.5f) / 10000f) +
+                               (((float)Math.Floor(Math.Abs(m_fieldMin.Z) / 50f + 0.5f) * 0.5f) / 10000f);
+
+               return factor * (float)Math.Pow(sensorVolume, 1f / 3f);
+            }
             else
                 return 0.0f;
         }
--- a/Sources/Sandbox.Game/Game/Screens/Helpers/MyToolbar.cs
+++ b/Sources/Sandbox.Game/Game/Screens/Helpers/MyToolbar.cs
@@ -40,10 +40,10 @@ namespace Sandbox.Game.Screens.Helpers
         }
 
         public const int DEF_SLOT_COUNT = 9;
-        public const int DEF_PAGE_COUNT = 4;
+        public const int DEF_PAGE_COUNT = 9;
 
         public int SlotCount = 9;
-        public int PageCount = 4;
+        public int PageCount = 9;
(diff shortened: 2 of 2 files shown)


No-Brain:
Currently power consumption is based on the volume covered by the sensor. This is completely unrelated to max range of the sensor.

aaraujo666:
Understood. However this change is a means to avoid creating over powered sensors. 

If you think of the standard vanilla sensor, it can be seen as using a transducer X for 50m detection. If you want a 100m detection it must require a more powerful transducer Y, hence the increased power consumption. 

If you don't do this, someone can create a sensor that can scan 1m left, right, top, bottom, and back; and a front of 250,000 m and use the exact same amount of power as a sensor with all extents set to 50m (it's the same volume)

mexmer:
@aaraujo666 you got it wrong. it's base on volume
so you can cay max power consumption of vanilla sensor is computed for volume
(2*50) * (2*50) * (2*50) aka 100 * 100 * 100 eg 1,000,000 cube meters is vanillavolume

if you set sensor to 100m front (rest is 50m) you have now
150*100*100 volume, eg 1500000 cube meters, and you need to adjust power consumption according to that change

if you create sensor with max limit 50,50,10,10,1000,1000, it's max power consumption will be
(100*20*2000/vanillavolume)*vanilamaxpowerconsumption

aaraujo666:
I understand it's based on volume. What I don't agree with is two sensors with the same power consumption as below

50, 50, 50, 50, 50, 50 = 1,000,000 cubic meters

1, 1, 1, 1, 1, 250,000 = 1,000,000 cubic meters

So in the first example (vanilla sensor set to max extents) you have a power consumption X. 

In the second example, it's the same volume, hence the same  power consumption, yet this sensor manages to detect 250,000 meters out.

MMaster:
@aaraujo666 which makes sense, because second one is only straight line while first one covers area. I guess it can be compared to real world example omnidirectional antenna and directional antenna. Both would have the same power consumption while the directional one would have much further range, but would cover only small arc in front of it.

aaraujo666:
Just to clarify... the change STILL uses the volume of the scanned area in it's calculations. It just increases the W/m3 consumed by the sensor for the extents that are above vanilla 50m.

So in your example above: "100m front (rest is 50m like vanilla)" my calculation will yield a slightly higher power consumption because you have one extent (front) above 50m. For every 50m that you increase a given extent, it increases the power consumption on that axis.

If you do the math... the sensor consumes 100W for every meter of extent that you set it.

Examples: All extents set to 1m, power = 600 W, all extents set to 10m 6,000w, etc.

What my calculation does is: 

All extents set to 50m, 30 KW (like vanilla) (50+50) * (50+50) * (50+50) ^ 1/3 * 0.0003
Front extent set to 100m, rest at 50m, (50+100) * (50+50) * (50+50) * 0.00035
Front and back extent set to 100m, rest at 50m, (100+100) * (50+50) * (50+50) * 0.0004

So in my second example, it increased the "wattage" to 0.00035/m3 because you had one extent over 50m. In third example, 0.0004 because you had 2 extents above 50m.

aaraujo666:
@MMaster I guess that makes sense, your omni vs directional argument.

So if I have, let's say a radio transmitter, using an omni antenna. And it's range is 100 km (i.e. people 100km away, in all directions, can receive my transmission). If I switch out the omni antenna, with a directional antenna which covers 1/100th of the transmission arc of the omni, using the same amount of power, my transmitter can now reach 10,000 km, provided the receiver is within the directional's TX arc?

I'm not being sarcastic here... really trying to understand. I understand the omni vs directional concept, but I don't think it "scales" like above.

aaraujo666:
Would it make more sense for an "enhanced" sensor (something that supports ranges over 50m) to require more radio and/or detector components to build instead of more power to operate?

Maybe an "upgrade module" similar to refinery/assemblers?

MMaster:
If we ignore the losses caused by atmosphere then in real world it works like this:
Omnidirectional antenna consumes power proportional to (PI * r^2)
Directional antenna consumes power proportional to ((arc angle radians)/2 * r^2)
so for your example:
omni = PI * (100 ^ 2) = ~ 31 416
direct = PI / 100 * (10 000 ^ 2) = ~ 3 141 593
but that takes arc into account which gets wider and wider the further you go. Essentially it is just volume of the circle and volume of the arc. So the current calculation in SE is correct and corresponds to real world behavior.

aaraujo666:
@MMaster Then I stand corrected! Thanks for the explanation!

PR #133, opened 2015-05-22
Fix typo in assembler/refinery control panel: Efficinecy -> Efficiency

Maeyanie:
It was bugging me. :)


lord-devious:
Finally :wink:

mexmer:
most likely they planed to fix this when adding localized strings, instead of hardcoded ones
i suggested you change your code to use localized strings.

lord-devious:
Also; shouldn't make pull request from your master.

No-Brain:
This should be using localized strings.
Any hardcoded texts are either bug or debug.

PR #134, opened 2015-05-22
Projector Opacity Sliders

phreakscp:
Add opacity sliders to projector block.
One slider is for blocks able to be welded; the other for blocks that aren't yet ready for welding
Should even sync across multiplayer (untested)


mexmer:
also add EnableActions() for each slider, that way they will be exposed to scripts api.
and if you also add direct fields for exposing values (read only is enough) to Ingame.IMyProjector ... it will be perfect ...

phreakscp:
Updated with your suggestions.

phreakscp:
Going to move this to a branch on my fork instead of the master

PR #138, opened 2015-05-22
Add cargo volume to ship HUD.

RossM:


mexmer:
could you please add something about value scaling?

i mean, for energy there is MW (many block use only kW, therefore you most likely not have values with more than 3 or 4 digits even on large ships/station

but in case of cargo. think about servers that have 10x inventory, or if you enter station/large ship cockpit) you can have number with 12 or more digits, that will be hard to read (and probly overlow hud)

Not sure if will be better
1.234kl
45.678Ml
890.123Gl
or 
1.234e3
45.678e6
890.123e9

but some scaling will be needed anyway space in hud is limited

RossM:
Used "1234 l" for volumes up to 1000 m³ (1,000,000 l), "123.4 m³" for volumes up to 10,000,000 m³, and "1.2345e+006 m³" for larger volumes.

Issue #89, opened 2015-05-19
Add an asteroid using an existing MyVoxelMapStorageDefinition.


midspace:
Currently, calling ...
Sandbox.Definitions.MyDefinitionManager.GetVoxelMapStorageDefinitions()
will return a list of ...
public ListReader<MyVoxelMapStorageDefinition>

We need a method on IMyVoxelMaps, that will allow us to specify a MyVoxelMapStorageDefinition to create a voxel map in game.
CreateVoxelMap(string id, Vector3D position);

or....

allow us clone the Storage on the MyVoxelMapStorageDefinition, to use in the existing CreateVoxelMap() method.

fabricator77:
I have update the branch with two extra commits, should fix sandbox issues, but I'm not able to test it. I've not called it CreateVoxelMap as that creates confusion with the existing function that requires more input data. The existing CreateVoxelMap is more a SaveVoxelMap to be honest.

midspace:
Has been added to code now.

PR #141, opened 2015-05-22
Create asteroid from prefab voxel definition.

midspace:
For #89


midspace:
Sorry, I'm still trying to get the pulls right. Github is difficult.

PR #142, opened 2015-05-22
Attach GameLogic to Blocks multiple times

Tyrsis:
This allows multiple scripts to attach to blocks without collisions.  If a block already has gamelogic attached, it will keep track of new gamelogic attachments and run those as well.

This also allows scripts to attach only to subtypes that they care about, so their logic only runs on the blocks they care about by SubTypeName.  I modified MyEntityComponentDescriptor to allow you to specify multiple subtypenames, so you can attach the same gamelogic to multiple subtypes (small and large for example).

Should be completely backwards compatible.  If new attributes are not used, it defaults to old behavior.  This also enhances old behavior in that multiple mods won't collide anymore if they want to attach logic to the same block as all scripts gamelogic will run.

Example:
`[MyEntityComponentDescriptor(typeof(MyObjectBuilder_Beacon), "SmallBlockRadar", "LargeBlockRadar")]`

When the above attribute is attached to a class that modifies beacon behavior, it only attaches to entities that have the subtypename of SmallBlockRadar or LargeBlockRadar.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,4 +38,3 @@ user.props
 3rd/SteamSDK/release/x64
 3rd/VRage.Native/debug/x64
 3rd/VRage.Native/release/x64
-global.props
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ user.props
 3rd/SteamSDK/release/x64
 3rd/VRage.Native/debug/x64
 3rd/VRage.Native/release/x64
+global.props
--- a/Sources/Sandbox.Common/Components/MyComponentBase.cs
+++ b/Sources/Sandbox.Common/Components/MyComponentBase.cs
@@ -11,23 +11,14 @@ namespace Sandbox.Common.Components
     public class MyEntityComponentDescriptor : System.Attribute
     {
         public Type EntityBuilderType;
+        public string[] EntityBuilderSubTypeNames;
 
-        public MyEntityComponentDescriptor(Type entityBuilderType)
+        public MyEntityComponentDescriptor(Type entityBuilderType, params string[] entityBuilderSubTypeNames)
         {
             EntityBuilderType = entityBuilderType;
+            EntityBuilderSubTypeNames = entityBuilderSubTypeNames;
         }
     }
-
-	[AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
-	public class MyEntitySubIdComponentDescriptor : System.Attribute
-	{
-		public string SubIdName;
-
-		public MyEntitySubIdComponentDescriptor(string subIdName)
-		{
-			SubIdName = subIdName;
-		}
-	}
 
     public abstract class MyComponentBase
     {
--- a/Sources/Sandbox.Common/Components/MyComponentBase.cs
+++ b/Sources/Sandbox.Common/Components/MyComponentBase.cs
@@ -18,17 +18,6 @@ namespace Sandbox.Common.Components
         }
     }
 
-	[AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
-	public class MyEntitySubIdComponentDescriptor : System.Attribute
-	{
-		public string SubIdName;
-
-		public MyEntitySubIdComponentDescriptor(string subIdName)
-		{
-			SubIdName = subIdName;
-		}
-	}
-
     public abstract class MyComponentBase
     {
         public IMyEntity Entity { get { return CurrentContainer != null ? CurrentContainer.Entity : null; } }  // to be obsolete once components are finished
--- a/Sources/Sandbox.Common/Components/MyComponentBase.cs
+++ b/Sources/Sandbox.Common/Components/MyComponentBase.cs
@@ -23,7 +23,7 @@ namespace Sandbox.Common.Components
     public abstract class MyComponentBase
     {
         public IMyEntity Entity { get { return CurrentContainer != null ? CurrentContainer.Entity : null; } }  // to be obsolete once components are finished
-        public MyComponentContainer CurrentContainer { get; private set; }
+        public MyComponentContainer CurrentContainer { get; set; }
 
         public virtual void OnAddedToContainer(MyComponentContainer container)
         {
--- a/Sources/Sandbox.Common/Components/MyComponentBase.cs
+++ b/Sources/Sandbox.Common/Components/MyComponentBase.cs
@@ -18,6 +18,17 @@ namespace Sandbox.Common.Components
         }
     }
 
+	[AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+	public class MyEntitySubIdComponentDescriptor : System.Attribute
+	{
+		public string SubIdName;
+
+		public MyEntitySubIdComponentDescriptor(string subIdName)
+		{
+			SubIdName = subIdName;
+		}
+	}
+
     public abstract class MyComponentBase
     {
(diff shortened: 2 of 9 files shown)


Sources/Sandbox.Game/Game/Entities/MyEntity.cs
```diff
@@ -477,18 +504,33 @@ public void CreateSync()
         public virtual void UpdateOnceBeforeFrame()
         {
             m_gameLogic.UpdateOnceBeforeFrame();
+			if (m_gameLogicSet != null)
```
harag-on-steam: This looks like inconsistent indentation settings.

Sources/Sandbox.Game/Game/Entities/MyEntity.cs
```diff
@@ -100,6 +100,13 @@ public MyGameLogicComponent GameLogic
             set { Components.Add<MyGameLogicComponent>(value); }
         }
 
+		private HashSet<MyGameLogicComponent> m_gameLogicSet = null;
```
harag-on-steam: Why not just make m_gameLogic a set? Now the class has two fields that basically serve the same purpose.

Tyrsis: I felt there was no need to needlessly enumerate when gamelogic had nothing attached to it.

THDigi:
Great, thanks!

Sources/Sandbox.Game/Game/Entities/MyEntity.cs
harag-on-steam: Hm, that could also be achieved by introducing a MyCompoundGameLogicComponent that has the collection and would be used as soon as there are more than one MyGameLogicComponents for the same entity.

Or maybe it would be enough to just introduce 3 delegates for the three Update*() methods and attach the set of MyGameLogicComponents to them? That could also serve as a more fine-grained way for update-notifications. If a component isn't interested in Update*100 it isn't registered to that delegate (by some opt-out or opt-in mechanism).

Now that I look at it again, is HashSet appropriate here? How is equality of the elements determined?

.gitignore
```diff
@@ -38,3 +38,4 @@ user.props
 3rd/SteamSDK/release/x64
 3rd/VRage.Native/debug/x64
 3rd/VRage.Native/release/x64
+global.props
```
THDigi: You should use user.props instead and undo this change :P

Sources/Sandbox.Game/Game/Entities/MyEntity.cs
Tyrsis: I thought about delegates, but then felt that would just be inconsistent with how logic is currently attached, and a bit confusing, while also forcing mods to change (a bad thing).

.gitignore
Tyrsis: Yeah, woops, let me try to fix things up again, heh.

Tyrsis:
Sorry about all the changes, I need to learn to git properly.

Sources/Sandbox.Game/Game/Entities/MyEntity.cs
```diff
@@ -100,6 +100,13 @@ public MyGameLogicComponent GameLogic
             set { Components.Add<MyGameLogicComponent>(value); }
```
harag-on-steam: `Components.Add` does not happen for the elements of `m_gameLogicSet`. That can't be a good thing.

Looking at `Components.Add` I guess each logic component should be added with it's own type or the components will overwrite each other in the component-list.

Sources/Sandbox.Game/Game/Entities/MyEntity.cs
```diff
@@ -477,17 +504,32 @@ public void CreateSync()
         public virtual void UpdateOnceBeforeFrame()
         {
             m_gameLogic.UpdateOnceBeforeFrame();
```
harag-on-steam: Beside the `Update*` methods there's also Init, MarkForClose, Close and UpdatingStopped that need to be handled.

Sources/Sandbox.Game/Game/Entities/MyEntity.cs
```diff
@@ -463,6 +470,26 @@ void Components_ComponentRemoved(Type t, Sandbox.Common.Components.MyComponentBa
...
         }
 
+        public void AssignGamelogicFromHashSet(HashSet<Type> items)
+        {
+            bool first = true;
+            foreach (Type logicType in items)
+            {
+                if (first)
+                {
+                    GameLogic = (MyGameLogicComponent)Activator.CreateInstance(logicType);
```
harag-on-steam: Hm, iteration order is unpredictable so it's unknown which component will end up in here.
Also, this single component can still be overwritten by setting GameLogic.

harag-on-steam:
And please don't take this the wrong way. I think this is an important change so I'm putting a lot of thought into the corner cases :)

Sources/Sandbox.Game/Game/Entities/MyEntity.cs
Tyrsis: Order does not matter here.  This is just to fill GameLogic first before filling the set.

Sources/Sandbox.Game/Game/Entities/MyEntity.cs
Tyrsis: Yes good eye, I will fix those.  UpdatingStopped wasn't ever called originally.  So I didn't include it.  Not sure if that was on purpose or not.

Sources/Sandbox.Game/Game/Entities/MyEntity.cs
Tyrsis: I'm not actually sure what they are doing with components.  I almost feel like it's a Type container for validation or something.  When I did my tests before, it didn't really seem to matter.  I will ask someone in Keen though.

Tyrsis:
There is an issue still with this I'm tracking down.

Tyrsis:
This should be fine now.  All tests passed.  Multiple attachments to a block type and sub block type, all working together in harmony.

THDigi:
This'll work even if multiple mods attach to the same typeid, right ?

Tyrsis:
Yes, this works with multiple mods attaching to same type, and subtype, and any combination within.

THDigi:
Ok so I understood it correctly then, awesome job :+1:

Sources/Sandbox.Common/Components/MyComponentBase.cs
```diff
@@ -11,17 +11,19 @@ namespace Sandbox.Common.Components
...
         {
             EntityBuilderType = entityBuilderType;
+            EntityBuilderSubTypeNames = entityBuilderSubTypeNames;
         }
     }
 
     public abstract class MyComponentBase
     {
         public IMyEntity Entity { get { return CurrentContainer != null ? CurrentContainer.Entity : null; } }  // to be obsolete once components are finished
-        public MyComponentContainer CurrentContainer { get; private set; }
```
OndrejPetrzilka: This seems wrong, protected is not enough?

Phoenix84:
Sadly it's not completely backwards compatible.
If there is only one script to attach, it all works fine. If there are multiple, however, Entity.GameLogic no longer represents what it's assumed to.

For example, in my Stargate mod, I case entity.GameLogic to my own class to access properties and methods on another entity.
I can't do that anymore. I looked at the code, but I can't find a way to get a specific MyGameLogicComponent object out of it.

I can filter by subtype, but that's not a solution for me, as I have a second mod that adds to it, but also includes the same script to be used independently if desired.

PR #144, opened 2015-05-22
Assembler queue functions for in game programming

Sibz:
Added functions:
GetQueueItems - Returns list of items in queue with string format type/subtypename
AddQueueItem - Adds item to queue using string type/subtypename and int amount
RemoveQueueItem - Removes an item from the queue
ClearQueue - Clears all items from queue
GetQueueItemAmount - Returns total amount of item in queue by typeid/subtypename


Sibz:
Closing to re submit tidier PR

PR #145, opened 2015-05-22
Opens the internal isActivated variable Fixes #118

### LINKED ISSUE #118: Add true/false read-only variable to determine if a welder/grinder/drill is welding/grinding/drilling
By including this variable in IMyShipWelder, IMyShipGrinder, and IMyShipDrill, in-game programs will able to determine if a tool block is being used, and take action according to that value.

skeltonn:
This allows in-game programs to determine if a tool is currently activated/working. This is further described in issue #118.


lord-devious:
Am I seeing the right code? This won't compile, properties have no parentheses.

Spartan322:
Also, the In-game scripts will not see this unless you modify the interface object too

PR #146, opened 2015-05-22
Complete Russian translation

KionX:


sjsoft2015:
In Russian this is wrong: "Мало энергии скафандра."

mexmer:
this doesn't belong to source code.
for translations there is subforum on forums also this
https://www.getlocalization.com/SpaceEngineers/

KionX:
every word 10 times translated. and does not determine the translation.
let it be here. maybe someone should.

lord-devious:
These files will be overwritten by their translation tool the second they regenerate the translations.

PR #148, opened 2015-05-23
Fix Drill Invisible Wall

IhanaMies:
Fix drills hitting invisible wall when mining

I can't explain how it works in detail. But a large drilling ship starting at 2 m/s to drill did not bump at once into the invisible wall. Speed drops very slowly to 1 m/s where it stays without a problem.


No-Brain:
This effectively increases the speed of the drills by 33%.
I don't think doing this is a good idea as the performance would suffer too much.
The results from this change would be different depending on the used hardware.

zrisher:
Is there a way for us to fix the invisible wall behavior without increasing drill speed?

Geneticus:
Isn't the drill rate effected by the material strength? i.e. some materials are harder/softer and are slower/faster to drill through.

PR #149, opened 2015-05-23
Enable hiding of terminal interaction dummies.

Phoenix84:
This change adds a UI button to toggle exterior terminal access for various blocks.
It does this by removing the terminal dummies you interact with.

I wrote this to be an extension of the 'Show In Terminal' feature.
It also doubles as a way for Text/LCD panels to show a display, while not having the yellow dummy covering the entire display when looking at it.

This is a re-request of #54 with generated file left in and synced to latest master.


Phoenix84:
I hate github...

PR #65, opened 2015-05-17
Turret AI improvements

NivvyDaSkrl:
* improved turret accuracy against targets with high tangential
velocity; added consideration of acceleration
* added consideration of target's likelihood of impacting the turret
(for meteors and missiles only)

PR #153, opened 2015-05-23
Improve turret aim.

RossM:
This uses a formula for aiming turrets based on more straightforward math. It gives the same result (without rounding) as the existing formula, but uses fewer floating-point operations. It is more accurate due to reduced round-off error and using double for all operations.

Also, prioritize targets based on the distance they will be in the near future, rather than the distance they are right now. This makes turrets much better at targeting meteors that are likely to hit.

### DIFF
--- a/Sources/Sandbox.Game/Game/Weapons/Guns/MyLargeTurretBase.cs
+++ b/Sources/Sandbox.Game/Game/Weapons/Guns/MyLargeTurretBase.cs
@@ -68,85 +68,12 @@ namespace Sandbox.Game.Weapons
 
             public MyLargeTurretBase Turret { get; set; }
 
-            //http://danikgames.com/blog/?p=809
             public Vector3D GetPredictedTargetPosition(IMyEntity target)
             {
-                Vector3D predictedPosition = target.PositionComp.WorldAABB.Center;
-
-                Vector3D dirToTarget = Vector3D.Normalize(predictedPosition - Turret.GunBase.GetMuzzleWorldPosition());
-
-                float shotSpeed = 0;
-
-                if (Turret.GunBase.CurrentAmmoMagazineDefinition != null)
-                {
-                    var ammoDefinition = MyDefinitionManager.Static.GetAmmoDefinition(Turret.GunBase.CurrentAmmoMagazineDefinition.AmmoDefinitionId);
-
-                    shotSpeed = ammoDefinition.DesiredSpeed;
-
-                    if (ammoDefinition.AmmoType == MyAmmoType.Missile)
-                    {
-                        //missiles are accelerating, shotSpeed is reached later
-                        var mDef = (Sandbox.Definitions.MyMissileAmmoDefinition)ammoDefinition;
-                        if (mDef.MissileInitialSpeed == 100f && mDef.MissileAcceleration == 600f && ammoDefinition.DesiredSpeed == 700f)//our missile
-                        {//This is very good&fast correction for our missile, but not for some modded exotics with different performance
-                            //still does not take parallel component of velocity into account, I know, but its accurate enough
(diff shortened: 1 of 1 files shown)


RossM:
Note that with this patch, turrets become extremely effective at destroying meteors. The two gatling and one missile turrets in easy start A can completely protect the starting platform in Armageddon difficulty for as long as they have ammo.

Spartan322:
Then the a good question is does it waste more ammo? Also I'd like to question balance if that's possible in Armageddon.

zrisher:
I know that I personally would like turrets to be really effective, because then people on my server would be happy with less of them, which means I can have fairly low limits and keep UPS higher for everyone.

RossM:
With this patch I don't think I've seen a turret miss a shot against a meteor that was going to pass within 200m. The only times they really waste ammo are when they fire a second shot before the first hits and destroys the meteor, or when the meteors just pass through the outer edge of their range.

Spartan322:
Then that still leaves balance as a question...

RossM:
Are you concerned about PvP balance or PvE balance? For PvE, turrets are currently fairly underpowered as a missile defense system, most people end up just building in the shadow of an asteroid. I'm hoping that make turrets more effective will open up more options at higher difficulty levels.

Spartan322:
@RossM I say adding a little inaccuracy might be a good idea, as this probably makes it much easier for the turrets to target a fast moving object.

zrisher:
Like I said, I think for my PvP server the overall sentiment is turrets could be a lot more accurate. You already can just snipe them from outside their range with a Gatling gun. Any additional fixed-point mod weapons usually have greater range than turrets as well.

Just out of my own curiosity, why is it that turrets much have such short range? I figured it was a performance thing - it looks like `CheckNearTargets` happens every 10 frames, and uses a bounding box the size of `range` to test for nearby targets against `MyGamePruningStructure` in `GetNearestVisibleTarget`. My understanding was that these pruning structure substantially decrease the effort required to do collision checks. Perhaps this check doesn't need to be done quite so frequently? Even every 100 frames seems fine to me - they'd take about 1.5 seconds to acquire a target, but if you could double their max range and have 3x as many with a similar amount of CPU use, wouldn't it be worth it?

image

https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/master/Sources/Sandbox.Game/Game/Weapons/Guns/MyLargeTurretBase.cs#L1759

demolish50:
IF that does not get accepted, know that you can pretty much perfect turret aim by turning the ammo speed to like 30k. Just fyi :)

zrisher:
You could, but don't projectiles moving too fast encounter the same issue as grids moving too fast - they move past physical barriers before Havok has a chance to check them for collision?

demolish50:
Has worked fine for me for 7 months on my server.. They instantly hit. Speeds of like 30k. You can't kite them and small ships are lol death. One problem is that maxrange doesn't really work at that speed. I've found 10k is a good sweet spot.

yakri:
It might be worthwhile to also increase turret bullet spread if you're increasing their accuracy. They are already overly accurate tbh.

MMaster:
What I don't like about improving accuracy this way and doing spread as weapon attribute instead is that currently you can dodge the bullets if you are skillful enough, after this change you will dodge if you are lucky enough.

RossM:
Dodging bullets is not really affected by this patch. Turrets still don't account for acceleration, so you just have to keep accelerating perpendicular to the line of fire.

MMaster:
Oh right, my fault :+1:

RossM:
Note that pull #65 also improves turret AI, but fixes different things (mostly).

NivvyDaSkrl:
They closed #65 without merging for balance reasons, FYI. May be applicable to this request as well.

PR #154, opened 2015-05-23
Projector visibility

phreakscp:
https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/134

Add opacity sliders to projector block.
One slider is for blocks able to be welded; the other for blocks that aren't yet ready for welding
Opacity settings can be bound to the toolbar from the G-menu
Opacity settings persist during save/restore and copy/paste
Should even sync across multiplayer (untested)


Pingger:
Nice, but there are "merge conflicts that must be resolved".
I would like to see this implemented =)

PR #155, opened 2015-05-23
Production & Inventory Screen Quantity Multipliers 

phreakscp:
Pressing the following key combinations to enqueue or transfer more materials on either the Production or Inventory screens:
Alt: 1000
Control + Shift: 1000 (note: The inventory screen already did this, but the production screen did not)
Control + Alt : 10K
Shift +Alt : 100K
Control + Shift + Alt : 1M

### DIFF
--- a/Sources/Sandbox.Game/Engine/Physics/MyDestructionData.cs
+++ b/Sources/Sandbox.Game/Engine/Physics/MyDestructionData.cs
@@ -64,7 +64,7 @@ namespace Sandbox
             Static = this;
             BlockShapePool = new MyBlockShapePool();
 
-            TemporaryWorld = new HkWorld(true, 50000, MyPhysics.RestingVelocity, MyFakes.ENABLE_HAVOK_MULTITHREADING);
+            TemporaryWorld = new HkWorld(true, 50000, MyPhysics.RestingVelocity, MyFakes.ENABLE_HAVOK_MULTITHREADING, 4);
             TemporaryWorld.MarkForWrite();
             TemporaryWorld.DestructionWorld = new HkdWorld(TemporaryWorld);
             TemporaryWorld.UnmarkForWrite();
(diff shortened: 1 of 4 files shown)


Wrennsun:
Drats, You beat me by 20 minutes. I just added a pull to use alt to pull a single item from a stack.

I also didn't change the production screen.

I am not sure what other way we could use to grab single items, which I think would be a useful feature.

Perhaps if you modified your code to use alt for just 1 item then your other additions would still work.

Wrennsun:
Something like this would leave your changes intact and also allow for alt to transfer single items.

var amount;

if( altPressed && !ctrlPressed && !ShiftPressed )
{
amount = 1;
} else {
amount= (altPressed ? 1000 : 1) *
               (shiftPressed ? 100 : 1) *
              (ctrlPressed ? 10 : 1);

MeusRex:
I'd rather have the other one, with a single item moved with Alt/LMB. If you have to move 1mio items, you can just as well pull the whole stack over...

THDigi:
Too many key combinations, just let the player type in their desired amount, way faster and way more accurate, I don't remember what combination for that is in inventories, right click drag I belive ?

thelastWallE:
Then just don't use these key combinations. ;)

THDigi:
I personally don't have a problem, I'm just providing feedback to make it better and more consistent with the rest of the game.

RodrigoT:
Could we also support these multipliers for the [+] and [-] buttons in the numeric input dialog?

Also, I suggested this on the forums before: use mid click to split values/stacks. not only when moving stacks around, but also for the production queue.

andrew-w:
I'm hoping keen reviews this and adds it for tomorrow's update!

I for one have been waiting for single item and larger multiple stack control on hot keys!!  And now it's sitting here waiting to go in :+1:

Geneticus:
How about a doubleclick modifier to move 1 of an item to another inventory?

phreakscp:
Closing pull request - replaced with PR #313 - Production screen quantity dialog

PR #159, opened 2015-05-24
Increase number of toolbars to 9

aaraujo666:
Increases the number of toolbars to 9 since this seems to be a popular request among players


Spartan322:
Wouldn't it make more sense for PageCount and SlotCount to use the DEF_PAGE_COUNT and DEF_SLOT_COUNT constants respectively.

aaraujo666:
I agree, but I just changed the value, the code was already like that.

Spartan322:
Magic Numbers are really annoying but ok, also, why are PageCount and SlotCount being set before construction and then being set again in the constructor, when you can't construct MyToolbar any other way, that also makes no sense.

Eikester:
i'd like a dynamic setup i.e. only add new toolbars if the old one is full

zrisher:
Awesome!!

PR #160, opened 2015-05-24
IMyShipConnector: ConnectStatus added, and comments

Sibz:
Added read only tri-state property ConnectStatus
Added comments to all properties.
Added Obsolete flag to IsLocked and IsConnected

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyShipConnector.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyShipConnector.cs
@@ -36,7 +36,7 @@ namespace Sandbox.ModAPI.Ingame
         /// OutOfRange - Not in range of another connector
         /// ReadyToLock - In range but not connected
         /// Connected - Connected to OtherConnector
-        /// i.e. if (myConnector.ConnectState == ConnectorState.Connected) DoStuffWhenConnected();
+        /// i.e. if (myConnector.Status == ConnectorStatus.Connected) DoStuffWhenConnected();
         /// </summary>
         ConnectorStatus Status { get; }
 
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyShipConnector.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyShipConnector.cs
@@ -18,25 +18,38 @@ namespace Sandbox.ModAPI.Ingame
         bool CollectAll { get; }
 
         /// <summary>
-        /// DEPRECATED use IsReadyToLock or for same functionallity IsReadyToLock || IsConnected
+        /// DEPRECATED use ConnectState
         /// Connector is working and in range of another connector, could be connected or ready to lock (Read Only)
         /// </summary>
-        [Obsolete("Deprecated, Use IsReadyToLock instead")]
+        [Obsolete("Deprecated, Use ConnectState instead")]
         bool IsLocked { get; }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


lord-devious:
Isn't this what IsLocked does?

mrsib:
Not Quite, IsLocked returns true when ready to lock and when connected. where as this returns true only when ready to lock.
To be honest the naming of IsLocked is confusing and perhaps should be deprecated.

lord-devious:
Oh, I see! I think I like that. And I agree about IsLocked, that's why I added IsConnected (and OtherConnector). Actually your property will deprecate IsLocked, I suggest you add an Obsolete tag to it.

mexmer:
would not be better to make it tristate and obsolete both of them?

mrsib:
@mexmer whats the best way to do that? I understand Enums don't work for in game, so would it be best to return string. i.e. "OutOfRange","ReadyToLock", "Connected"?

lord-devious:
Enums created in game doesn't work, API created ones does. And I'll tend to agree with @mexmer , it's what I did for doors and pistons. Question is how many obsolete properties we want. At some point one just has to accept how things are.

mrsib:
@mexmer @lord-devious 
Well funny, my original intention was for a tri-state property, but wasn't sure on enums and saw IsConnected was already there so decided to just add IsReadyToLock

Added the above commit, Is the ConnectorState Enum structure good enough? Is it in the right place?

lord-devious:
Heh that's what went through my mind when I added IsConnected :) (although I knew enums work) 
I don't really like obsoleting stuff, because it leaves a mess in the files so I went through that 

Looks good to me although it should have its own file.

Sibz:
@lord-devious Yeah own file probably best, but I wouldn't know where to put it in the giant folder structure

Sibz:
@lord-devious I see what you did with pistonstatus, I'll do that.

Sibz:
@lord-devious done, and renamed to status to keep naming in line with pistonstatus

Sources/Sandbox.Common/ModAPI/Ingame/IMyShipConnector.cs
```diff
@@ -7,10 +7,43 @@ namespace Sandbox.ModAPI.Ingame
...
+        /// <summary>
+        /// Connector is set to pull items from connected inventories (Read Only)
+        /// </summary>
         bool CollectAll { get; }
+
+        /// <summary>
+        /// DEPRECATED use ConnectState
+        /// Connector is working and in range of another connector, could be connected or ready to lock (Read Only)
+        /// </summary>
+        [Obsolete("Deprecated, Use ConnectState instead")]
```
harag-on-steam: There is no ConnectState any more ;-)

Sibz: Haha, Thanks, Fixed.

mexmer:
@lord-devious it doesn't need to be in separate file, if it's used only by that interface, it's better if enum is same file as interface

lord-devious:
I disagree. All types should be in their own files, much cleaner that way. Anyway it's up to OP. I was merely stating my opinion.

Sibz:
I am thinking both are right, its nice to have it cleanly in its own file, however lots of little enums will clutter the file list and it can be in same file as interface.
As PistonStatus is separate, and the Devs oversaw the introduction of that, I will keep it the same. It can be tidied up later if necessary.

Sibz:
Conflicts

PR #162, opened 2015-05-24
Improved DX9 lighting (not affect to performance)

KionX:
Unfortunately it needs some tweak content.
Asteroids specularity too much.
Some blocks also.


THDigi:
Some comparison pics ? :P

PR #165, opened 2015-05-24
Add hotkey for color change.

RossM:
Some users have no middle mouse button, such as laptop users with trackpads. Add slash ('/') as a keyboard alternative for painting blocks.


kapitanov:
Game actually has a "Controls" screen in its main menu, users can change their key bindings there. And I'm pretty sure there is a "Change color / weapon mode" option under "Systems 3" page so there's no need to hardcode any alternate buttons.

Issue #164, opened 2015-05-24
Feature Request: Adjust mass of Artificial Mass block


kapitanov:
Space Ball block has a special slider in terminal to modify its mass. It would be nice to have the same slider for good old Artificial Mass block.

PR #168, opened 2015-05-24
#164 Virtual mass slider for Artificial Mass Block

kapitanov:
I've added a slider in Artificial Mass block's terminal that allows to adjust its mass (see issue #164).
Maximum mass is set to 10000 and can be changes from block definition.


JJRcop:
Just kinda gonna say that the space ball was an April Fools joke.

PR #169, opened 2015-05-25
Dynamic notification for the API

THDigi:
Added MyAPIGateway.Utilities.CreateNotification() which returns an MyHudNotification object  (under the IMyHudNotification interface) that can be used to dynamically show, hide, change text, change font and change the alive time of the notification without needing to create another object.

A snippet from my test mod:
```
        bool init = false;
        private static Random rand = new Random();
        int fontSwap = 0;
        const int fontSwapMax = 15;
        IMyHudNotification notification;
        
        public void Init()
        {
            init = true;
            
            notification = MyAPIGateway.Utilities.CreateNotification("Test :}", 20, MyFontEnum.Red);
            notification.Show();
        }
        
        public override void UpdateAfterSimulation()
        {
            if(MyAPIGateway.Session == null)
                return;
            
            if(!init)
            {
                Init();
            }
            
            if(++fontSwap >= fontSwapMax)
            {
                fontSwap = 0;
                notification.Font = (notification.Font == MyFontEnum.Red ? MyFontEnum.Blue : MyFontEnum.Red);
            }
            
            notification.Text = "random=" + rand.NextDouble();
            notification.ResetAliveTime();
       }
```

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMyHudNotification.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyHudNotification.cs
@@ -1,0 +1,45 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+using Sandbox.Common;
+
+namespace Sandbox.ModAPI
+{
+    public interface IMyHudNotification
+    {
+        /// <summary>
+        /// Get or set the notification text.
+        /// Setting the text will immediatly update it if the notification is shown.
+        /// </summary>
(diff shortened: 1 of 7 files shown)


kleshchevand:
Limited exposure of this to ingame sripts would have been an interesting feature - 'personal UI/HUD' for each vessel

THDigi:
@kleshchevand hmm, interesting idea, that could be a thing for the cockpits/seats which PBs can loop through and send message to their pilots/passengers.

PR #170, opened 2015-05-25
Ingame Door Position properties

midspace:
For issues #3


midspace:
Sorry, I'm still trying to get the pulls right. Github is difficult.

PR #172, opened 2015-05-25
Ingame Door Position properties

midspace:
For issue #3


lord-devious:
I was just told by Ondrej that he doesn't like multiple properties in interfaces which essentially returns the same data in just different "formats". This is why he didn't merge my DoorStatus PR. I don't really know where the ratio came from, but it can easily be used to determine if a door is actually open or actually closed. Perhaps retain only the max value? Because the current position can be calculated using the ratio.

midspace:
Can I have direct confirmation from @OndrejPetrzilka on that?

PR #175, opened 2015-05-25
Adding AlphaSaturation tags in TransparentMaterials.sbc

aaronfranke:
Added "<AlphaSaturation>1</AlphaSaturation>" to every material with no <AlphaSaturation> tag.

This change fixes a bug when running the game under WINE which causes light sources, such as thrusters and the sun, to have a solid black box around them. Even though WINE is not officially supported, the fix is so easy and doesn't affect Windows players, so there isn't a reason not to add it.

### DIFF
--- a/Sources/SpaceEngineers/Content/Data/TransparentMaterials.sbc
+++ b/Sources/SpaceEngineers/Content/Data/TransparentMaterials.sbc
@@ -228,7 +228,7 @@
         <SubtypeId>GlareHeadlight</SubtypeId>
       </Id>								
       <AlphaMistingEnable>false</AlphaMistingEnable>
-      <AlphaSaturation>1</AlphaSaturation>
+      <AlphaSaturation>0.55</AlphaSaturation>
       <CanBeAffectedByOtherLights>false</CanBeAffectedByOtherLights>
       <Emissivity>0</Emissivity>
       <IgnoreDepth>true</IgnoreDepth>
--- a/Sources/SpaceEngineers/Content/Data/TransparentMaterials.sbc
+++ b/Sources/SpaceEngineers/Content/Data/TransparentMaterials.sbc
@@ -7,6 +7,7 @@
         <SubtypeId>Square</SubtypeId>
       </Id>
       <AlphaMistingEnable>false</AlphaMistingEnable>
+      <AlphaSaturation>1</AlphaSaturation>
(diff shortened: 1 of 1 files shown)


No-Brain:
How about changing the default value instead?

aaronfranke:
That may work, possibly even better, since we won't have to add entires for every material. You just have to make sure the default value being changed isn't affecting anything else.

aaronfranke:
I don't actually know where the default value is stored.

THDigi:
@aaronfranke 
It's in Sandbox.Common.ObjectBuilders.Definitions.MyObjectBuilder_TransparentMaterialDefinition

aaronfranke:
For some reason, I see the file "TransparentMaterialDefinition.cs" when looking under GitHub's website, but on my cloned repository, I only have "TransparentMaterial.cs" and "TransparentMaterials.cs". I tried deleting and re-forking, and the file is still missing? I would download it manually, but I fear I'd be missing references to it or I'm missing other files too or something.

THDigi:
One's in `/ObjectBuilders/` and the other is in `/ObjectBuilders/Definitions/`

https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/2073ef1743fca874b9ac748c0ee07677e151b208/Sources/Sandbox.Common/ObjectBuilders/Definitions/MyObjectBuilder_TransparentMaterialDefinition.cs

https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/2073ef1743fca874b9ac748c0ee07677e151b208/Sources/Sandbox.Common/ObjectBuilders/MyObjectBuilder_TransparentMaterial.cs

^ this one already has AlphaSaturation set to 1 though, set it in the other one and test it.

aaronfranke:
Ok, I did that, and sorry to bother you with more issues, but I can't build the game. SE is stored for me on S:\Steam\SteamApps\common\SpaceEngineers, and I changed the file in global.props to reflect this, but Visual Studio can't find any of the folders in the SpaceEngineers\Content folder. I tried granting everyone access, unsetting read-only, and running Visual Studio as administrator, but it still can't find anything. 

When I try to click on one of the missing files, it tells me that it "Cannot find the file"  and "It may have been moved or deleted" even though it is there under Windows Explorer. One thing that's odd, under "Properties" and "Misc" it lists "File Name" as "Audio", shouldn't it be folder name?

PR #176, opened 2015-05-25
IMyInventoryItem: Added OxygenBottleLevel

Sibz:
Readyonly property OxygenBottleLevel for retrieving how full an oxygen bottle is.
Returns -1 if called on a non OxygenContainerObject.
Compiled and tested Ok.


THDigi:
Hmm, wouldn't it be better to create an interface for the bottle instead ?

No-Brain:
Inventory should not have functionality for specific item types.

Sibz:
Roger will do it that way.

PR #179, opened 2015-05-25
Added IMyOxygenBottle with OxygenLevel property

Sibz:
Enables retrieval level of oxygen in the bottle as float 0-1.
I.e. 0.5 = 50%
Usage Example:
```c#
var OxyGen = GridTerminalSystem.GetBlockWithName("Oxygen Generator");
var inv = OxyGen.GetInventory(0);
var items = inv.GetItems();
items.ForEach(item =>
{
  if (item.Content is IMyOxygenBottle)
    Echo("OxyLevel:" + (item.Content as IMyOxygenBottle).OxygenLevel);
});
```
Compiled and Tested OK.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMyCubeBlock.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyCubeBlock.cs
@@ -6,6 +6,9 @@ using VRage.Utils;
 
 namespace Sandbox.ModAPI
 {
+    /// <summary>
+    /// base block interface, block can be affected by upgrade modules, and you can retrieve upgrade list from <see cref="IMyUpgradableBlock"/>
+    /// </summary>
     public interface IMyCubeBlock : Ingame.IMyCubeBlock, IMyEntity
     {
         event Action<IMyCubeBlock> IsWorkingChanged;
--- a/Sources/Sandbox.Common/ModAPI/IMyTerminalBlock.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyTerminalBlock.cs
@@ -5,12 +5,22 @@ using System.Text;
 
 namespace Sandbox.ModAPI
 {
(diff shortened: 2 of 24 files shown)


Sibz:
@mexmer It is:
Sources/Sandbox.Game/MySandboxGame.cs
```c# IlChecker.AllowedOperands.Add(typeof(Sandbox.ModAPI.Interfaces.IMyOxygenBottle), null);```

lord-devious:
If it is intended for use in ingame it should be placed in that namespace, I think.

mexmer:
@Sibz this doesn't add native ingame programming interface, you will need to change to this type, and is not in ingame namespace as said @lord-devious

also, since you adding summary for filed, interface should have also summary, this summary will be visible in documentation

Sibz:
@mexmer 
> this doesn't add native ingame programming interface, you will need to change to this type
I am not sure what you mean here

lord-devious:
I forgot to say; great idea :smile:

Sibz:
Updated as per last commit. This means line in MySandboxGame.cs wasn't needed, is that what you meant @mexmer ?

mexmer:
yes, it is not needed, if you define it this way.
also it has now propert inheritance and placement in both modapi and ingame script .... at least that's what i observer on other interfaces, that are available in PB

Sibz:
@OndrejPetrzilka Doesn't appear possible now the base definition `MyObjectBuilder_OxygenContainerObject` is in Vrage.Game as it cannot implement interface in sandbox.modapi.ingame namespace.
If it is possible, hints on how would be appreciated.

mexmer:
don't be haste, i will check it later ... there might be a way
oxy bottle is special type of item, since it's kind of container.

Sibz:
@mexmer Thank you!

mexmer:
here you go #299
approach from different side tho', not just because of object builders.

PR #181, opened 2015-05-26
Improved assembler pull for faster assemblers.

joemorin73:
Modified calculations of the assembler updates to properly calculate the correct amount of ingots for the assemblers to use.  This should also work with 10x assemblers in 10x environments.

(This does not address the max 1 item per 10 frames.  That is coming in my next commit.)


joemorin73:
I apologize for the updates mess.  Getting used to the GIT system is a learning curve.

RossM:
You just beat me. I have exactly the same fixes plus some others in #182.

joemorin73:
I was going to work on the coop too.  I couldn't figure out what to do with the calculations and was going to sleep on it.  I guess I can jump over to the UpdateProduction instead.

joemorin73:
This isn't needed anymore.

PR #185, opened 2015-05-27
Improved Assembler coop with repeat

### LINKED ISSUE #196: Issues with Co-Op Assembly
@RossM RE: #182
Seems the Assemblers are pulling ingots better, but not working so well in Co-Op mode:

See Thread:
http://forum.keenswh.com/threads/assembler-community-fix.7360798/

joemorin73:
Slaved assemblers will now queue more items.

 Fixes #196

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyAssembler.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyAssembler.cs
@@ -494,6 +494,8 @@ namespace Sandbox.Game.Entities.Cube
 
         private void GetItemFromOtherAssemblers(float remainingTime)
         {
+            var factor = MySession.Static.AssemblerSpeedMultiplier * (((MyAssemblerDefinition)BlockDefinition).AssemblySpeed + UpgradeValues["Productivity"]);
+
             var masterAssembler = GetMasterAssembler();
             if (masterAssembler != null)
             {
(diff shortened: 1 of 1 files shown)


RossM:
If you want to handle repeat assemblers like this, you should take the item from the master queue and readd it to the end. It would be easiest to merge that with the case below.

joemorin73:
You mean remove it from the masterqueue?  Would that work if I need more than what is currently in the queue?

It definitely works better with the previous update.  Probably right, I should have combined the updates.

RossM:
If you don't move in the master queue, the number of items produced will go imbalanced.

Imagine you have 500 steel plates and 100 construction components queued in the master. (Assume the master is making no progress - maybe it's turned off.) The slave comes along, grabs 5 steel plates. It finishes one and grabs another one. However, if you haven't moved it in the master queue, there's still 500 steel plates at the front of the queue. The slave will continue grabbing nothing but steel plates, and never make any construction components.

By moving things to the end of the master queue when a slave grabs them, you ensure that the ratio of items produced is the same as the player wanted.

joemorin73:
But I'm ensuring I grab a complete snapshot of what is in the queue, no partial.  Even is it starts 500 steel and 100 construction, it can turn to 250 steel, 100 construction and 250 steel, I only copy the total in the queue.  So even if I move each item from the beginning to the end, the queue will return exactly where I left it.

The slaves won't be copying the individual components, they will be copying the batch.  It's not perfect, but then again, a single assembler on repeat can't guarantee it will finish either.

RossM:
Ah, I misread your code. Never mind then!

joemorin73:
No worries.  ;)

I had to double check my code to make sure I did it right.

PR #186, opened 2015-05-27
BugFix for MyPlugin

sjsoft2015:
1) Keeping all switches that run the process, when it is restarted.
2) Removed pieces block raises an exception when loading with switch
-plugin


OndrejPetrzilka:
Why not to write into MyLog at all?

sjsoft2015:
Because application crash on start. If u run dedication server with -plugin option u will see this. 
I think logging is required, but there are two process starts, it loads plug twice. The idea should be tweaked to run did not exist, but it can affect large numbers of files.

OndrejPetrzilka:
ok, agree

sjsoft2015:
Created a new pull request #234

PR #187, opened 2015-05-27
Relayer HUD markers

andrew-w:
Changed draw order of HUD markers and elements so that GPS markers, location markers, button markers etc, do not render in front of HUD elements like the toolbar or block build progress/info box.


MeusRex:
Is it a good idea to move the base.Draw()? That part seems to abort the drawing if there is a problem. Moving it to a point after the first HUD draw calls doesn't seem like a good plan.

andrew-w:
Fair point!  I did think about that, but looking at the base.Draw() method it always returns true, there's no detection to return false.  The reason for moving it down is that having the base.Draw earlier in the method meant that the toolbar and some other HUD info boxes still rendered behind the markers.

I could move it back to earlier in the method, but then I'd have to start re-arranging the base class and still being largely unfamiliar with the code base, that could cause many more issues across the board.

Keen can always reject it if it will be a problem, then I guess I'll have to start delving into the base class.

MeusRex:
I just worry a bit about what would happen if some day keen added a new part to the base which could return false...

Though I guess that is for keen to decide/know.

andrew-w:
Closed pull request as I merged it incorrectly and I want to rework the way I handle the relayering.  Will resubmit in a new branch and pull request.

PR #189, opened 2015-05-27
Fix characters dying suddenly when exiting cockpit.

RossM:
This seems to fix the issue with characters dying instantly upon exiting a cockpit, especially while travelling at high speed, or in creative mode suddenly flying off uncontrollably. It is motivated by the observation that removing and putting back on your helmet will restore control. To be quite honest, I am not entirely sure why this works, but it seems to fix the problem.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyCockpit.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyCockpit.cs
@@ -627,6 +627,12 @@ namespace Sandbox.Game.Entities
             if (usePilotOriginalWorld || allowedPosition.HasValue)
             {
                 Hierarchy.RemoveChild(m_pilot);
+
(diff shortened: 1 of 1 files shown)


Spartan322:
The reason this works is you are resetting the pilot's player entity when he exits his vehicle

RossM:
Well, yeah. But I don't know what about the player entity is causing the bug; I'm just hitting the problem with a huge hammer to get it to go away.

Spartan322:
Wouldn't this cause some weirdness if he exits right before he crashes into a wall.

yajiedesign:
Are some rough?

No-Brain:
This is a hack. Try to find out why this happens in the first place.

OndrejPetrzilka:
Thanks for provided info, we'll look at it and see why this 'helps'.
This won't be merged because of possible side effects.

OndrejPetrzilka:
I've just got info that this is fixed on latest internal version. We're not sure why this happens, but it was fixed by resetting ragdoll when leaving cockpit.

PR #190, opened 2015-05-28
Upgrade module interface changes

mexmer:
changed upgrade module parent to functional block to allow enable/disable using pb/terminal
added access to active upgradelist on affected block (IMyUpgradableBlock), and on upgrade module itself (IMyUpgradeModule) - all fields are readonly, not intended to be modified from PB.


mexmer:
this PR is universal version of #123, which is useable only for Assemblers and Refineries, either of these will be fine to be merged (this one is preffered, because works also for other upgradable blocks)

OndrejPetrzilka:
1) This seems weird, can you rephrase it:
Modifier value as float (2 means double on addtive, 1 means +100% on aditive)

2) I don't like putting interface on definition struct, however you should not need interface, you can return struct itself (it will return copy), agree?

Next time I would prefer 2 PR, one for terminal thing, one for interfaces.

mexmer:
i will split it  into two PRs if needed.

as for struct interface, i was not sure about moving struct definition outside where is it now. since where is it now. is not accessible for ingame scripts (pb), that's why i created interface, which is implemented by struct.
and yes, it returns copy.

idea behind this is to provide needed information, for people doing management scripts ... this informations can be readonly snapshot at time of "tuning" block.

as for 2, that's because i mixed two things (at least, when i was writing that), when you check upgrade value on block for aditive, you actually get 2, if you have connected module with 1 aditive .... it's confusing indead.

OndrejPetrzilka:
We will enabled definition types in ingame programming.
It does not mean that PB will have unlimited access to definitions, we cannot do that because if it modifies it, it can break a lot of things.

OndrejPetrzilka:
Definition types are now available for InGame programming 7de5a79648881cf196d10d878fcde569ef91154b

mexmer:
should be done.
should i sync it with latest merges?

PR #193, opened 2015-05-28
Suspension extensions (+fix)

THDigi:
- Added sliders for steering angle, steering speed, steer return speed, suspension travel (which can also be used to lock the suspension) and a checkbox to invert steer direction;
- Reoriganized the control panel: [panel part 1](http://i.imgur.com/b9F6s7n.jpg), [panel part 2](http://i.imgur.com/d9GeHPw.jpg);
- Removed "SuspensionLimit" cubeblock parameter and made "MinHeight" and "MaxHeight" to act as suspension limits instead;
- The added sliders are set by default to the values they were set prior to having sliders but their respective value limits have been increased a bit to allow more customization;
- Added the MyMotorSuspension_ModAPI class and moved the mod API related code there and updated IMyMotorSuspension;
- If the suspension body doesn't have a wheel attached, all suspension controls are disabled;
- Fixed the shaking described in https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/120

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyMotorSuspension.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyMotorSuspension.cs
@@ -8,11 +8,49 @@ namespace Sandbox.ModAPI.Ingame
     public interface IMyMotorSuspension : IMyMotorBase
     {
         bool Steering { get; }
+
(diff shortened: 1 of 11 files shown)


THDigi:
@OndrejPetrzilka updated to latest branch.

PR #194, opened 2015-05-28
IMyBatteryBlock: Added IsCharging property

Sibz:
IsCharging returns true if IsWorking and block has been set to recharge.

Compiled And Tested OK.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyBatteryBlock.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyBatteryBlock.cs
@@ -11,5 +11,7 @@ namespace Sandbox.ModAPI.Ingame
 
         float CurrentStoredPower { get; }
         float MaxStoredPower { get; }
+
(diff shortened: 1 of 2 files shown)


THDigi:
You can already get those values, I think it would be more useful to know if it's actually charging, maybe a check in power input :P

Sibz:
How do I retrieve if block is set to recharge currently?

THDigi:
It doesn't have a GetValue() parameter ?
Then it only leaves the `.GetAction().WriteValue()` which is unreliable, so nevermind xD

In that case, add 2 fields, one only returns the charge checkbox's status (without caring for IsWorking) and one that tells you if the battery currently draws power.

Sibz:
Well thinking about it, I may as well just expose ProducerEnabled true if its not charging, aka producing power.
As for power consumption, that's a different matter, other blocks consume power too, so this would be better if IMyPowerConsumer was exposed with required fields.
Closing this pull in favour of ProducerEnabled.

PR #195, opened 2015-05-28
IMyPowerProducer: Exposed ProducerEnabled

Sibz:
Exposed ProducerEnabled to in game programming.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyPowerProducer.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyPowerProducer.cs
@@ -21,5 +21,10 @@ namespace Sandbox.ModAPI.Ingame
         /// Max power output defined in definition [MW].
         /// </summary>
         float DefinedPowerOutput { get; }
+
+        /// <summary>
+        /// Power production is enabled
+        /// </summary>
+        bool ProducerEnabled { get; }
     }
 }


THDigi:
Ok now I'm confused, what happened to the field for the "Charging" checkbox ? xD

Also, wouldn't `CurrentPowerOutput > 0` be equivalent of ProducerEnabled ?

mexmer:
@THDigi there is not field charging in ingame script api.
there is only "Recharge" action, that allows you to togle recharging.

THDigi:
Well, same thing :P But I'm also talking about the discussion we had here: https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/194

Sibz:
I changed it because all that's needed is to expose `ProducerEnabled`, something that applies to all `IMyPowerProducer` and can be re-used. My previous commit with `IsCharging` was really just an alias and in a sense duplication.
`CurrentPowerOutput > 0` is not the same as `ProducerEnabled` as when `ProducerEnabled` is true `CurrentPowerOutput` may still be 0.
With this you can work out that a `IMyBaterryBlock` is charging by just `!ProducerEnabled`

OndrejPetrzilka:
Fixed missing implementations and renamed to ProductionEnabled in 9fbfc68f1cd1b55e8e6bf43d756ed68610f92958

PR #197, opened 2015-05-29
Add rotation and elevation clamping to the laser antenna.

Phoenix84:
This adds the ability to set rotation and elevation limits to the laser antenna, just like the turrets have.

The code is pretty much copied from the turret itself. I wish the laser antenna and turrets shared common code, but it appears they are each implemented on their own.


Phoenix84:
Done, and done.
Additionally, the message will be shown and in-game boolean will be set if the hardcoded elevation limit is also reached.
I decided to leave that limit in, as the turret still has it, and I didn't want to risk changing behavior there.

mexmer:
nice

Phoenix84:
Ok, I merged master. I hope I did that right. Never did a git merge before.

OndrejPetrzilka:
Sometimes it's good to manually move your code in resx files few elements above to prevent merging issues :)

Phoenix84:
Good idea, Ondrej. I'll do that next time. :+1:

PR #198, opened 2015-05-29
Fixed Check for DX11 support

mexmer:
current check added microsoft software render into list (this is installed by IE7 and newer), therefore game detected DX11 support without DX11 hardware

tested with help of @THDigi

### DIFF
--- a/Sources/VRage.Render11/Render/MyRender-Adapters.cs
+++ b/Sources/VRage.Render11/Render/MyRender-Adapters.cs
@@ -205,7 +205,7 @@ namespace VRageRender
 
                 LogAdapterInfoBegin(ref info);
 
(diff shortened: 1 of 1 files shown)


OndrejPetrzilka:
Is this safe check?
 && vram > 0

What about integrated GPUs?

THDigi:
I tested it on an integrated GPU (chipset not on CPU) and it has VRAM set.
I also tested on my i5's integrated GPU and it also has VRAM specified.

The 'else' bracket is what solves the issue to begin with, the vram thing is extra xD

mexmer:
We can remove the check for ram probly, but since windows need some vram, there will be non nul value. Assuming what i've seen in dxdiag on chipset and apu based cards

PR #199, opened 2015-05-30
Filter blocks from toolbar config

andrew-w:
Added ability to filter from the toolbar config page.


lord-devious:
Me like. :trophy:

Sources/Sandbox.Game/Game/Entities/Blocks/MyBeacon.cs
```diff
@@ -53,6 +53,11 @@ static MyBeacon()
             show.Setter = (x, v) => x.RequestShowInTerminal(v);
             MyTerminalControlFactory.AddControl(show);
 
+            show = new MyTerminalControlOnOffSwitch<MyBeacon>("ShowInToolbarConfig", MySpaceTexts.Terminal_ShowInToolbarConfig, MySpaceTexts.Terminal_ShowInToolbarConfigToolTip);
```
RossM: Why add this here (and in MyRadioAntenna)? The control should already be added by MyTerminalBlock.

lord-devious:
Probably because this is how Keen did it with their ShowInTerminal?

mexmer:
just note, always create new var for your control. you did that in terminalblock, but for some reason you reused "show" variable in others ....

using new variable is better to prevent errors, in case someone rely on that value in later code ... i know it's short, but still ... someone might later add aditional logic to controls at end of fucntions, and here you go ....

OndrejPetrzilka:
Can someone explain me purpose of this? This is to remove mess in G screen when putting things on ship toolbar?

Please use new variable instead of 'show' as someone suggested.

mexmer:
@OndrejPetrzilka yes, seems to be same kind of filter like "Show in terminal", but this time for toolbar (g menu)

andrew-w:
Updated code as requested to use new vars instead of re-using old ones.

And yes, this is to hide blocks in the ship toolbar G screen.  To reduce the mass of blocks listed when configuring cockpit toolbars, timer blocks etc.

I have one question about the code though.  In the MySyncBlockHelpers.cs file, I used a number for the message ID (see line 80) which was close to the ShowInTerminalMsg ID.  Is there a proper way to determining which ID's to use?  or is just finding one that isn't in use good enough?  As that 's what I did for this one.

OndrejPetrzilka:
There's no better way of determining message id now. Please update you branch so it can be automatically merged.

andrew-w:
Updated branch  :)

PR #152, opened 2015-05-23
Create new block in mod

yajiedesign:
you can create new block in mod.
mkdir "Blocks" in modpath\Data\
Place 
IMyNewBlock.cs
Ingame_IMyNewBlock.cs
MyObjectBuilder_NewBlock.cs
MyObjectBuilder_NewBlockDefinition.cs
MyNewBlock.cs
MyNewBlockDefinition.cs
in modpath\Data\Blocks
and you can use new block in Data*.sbc

PR #201, opened 2015-05-30
Add custom DetailedInfo text that can be changed in-game

Phoenix84:
A new CustomDetailedInfo property has been added to allow modders or in-game script writers to append custom text to the DetailedInfo shown in the terminal.

**This will not affect the existing DetailedInfo property, which is still read-only.**

This is useful for custom upgrades, see screenshot.

2015-05-30-02-31-49-901_finalscreen

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyTerminalBlock.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyTerminalBlock.cs
@@ -11,11 +11,14 @@ namespace Sandbox.ModAPI.Ingame
         string CustomName { get; }
         string CustomNameWithFaction { get; }
         string DetailedInfo { get; }
+        string CustomDetailedInfo { get; }
         bool HasLocalPlayerAccess();
         bool HasPlayerAccess(long playerId);
         void RequestShowOnHUD(bool enable);
         void SetCustomName(string text);
         void SetCustomName(StringBuilder text);
+        void SetCustomDetailedInfo(string text);
+        void SetCustomDetailedInfo(StringBuilder text);
         bool ShowOnHUD { get; }
         void GetActions(List<Sandbox.ModAPI.Interfaces.ITerminalAction> resultList, Func<Sandbox.ModAPI.Interfaces.ITerminalAction, bool> collect = null);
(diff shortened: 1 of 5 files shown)


lord-devious:
I'm wondering if this should be disabled for the programmable block. It has this kind of feature already through its Echo function, and I don't think it'd be a good idea for mods to be able to mess with the output as it's integral to the function of the block...

Just a thought, mind you.

Phoenix84:
I don't know what the 'echo' function is. I don't use in-game programming myself.
However this only adds something extra, it doesn't let you change what's already there.
I specifically wrote it this way so you couldn't tamper with the original data.

lord-devious:
The Echo function lets you print out whatever text you want from within the programmable block in question.

yajiedesign:
look my pull,it can add new block in game .you can custom every gui.
https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/152

Phoenix84:
I've seen that. Not really applicable to this though.

OndrejPetrzilka:
I would allow modders to modify original DetailedInfo, would there be any problem with that?

Setting some text through PB (or mods) and syncing it over network looks dangerous (huge traffic, I wouldn't do that). Local PB output with Echo is good.

Either way, I prefer not to sync the text (the values which generates text are already synced)

Phoenix84:
I wasn't sure if the original DetailedInfo was using for anything but display. That's certainly doable.
Regarding network traffic: If a script is only executing on the server (in the case of ModAPI), how would clients see the change?
It sounds like in that case, the script would have to execute on all clients. Is that what you mean?

OndrejPetrzilka:
The antenna state should be synced, not the text. So "Out of Range" state would be same on server and all clients. The text would be then generated on server and every client and would be same.

I'll add support to modify DetailedInfo in ModAPI, then we can get back to this and see if it meets your requirements or not.

Anyway, thank you for the initiative. I can imaging this is wanted feature.

OndrejPetrzilka:
Fixed by 75cea190aac8c49669b51d361c2fc7cb00b1f42e

I've made it so it's possible for multiple mods to update the info.
1. hook event to IMyTerminalBlock.AppendingCustomInfo
2. call IMyTerminalBlock.RefreshCustomInfo()

This way mods won't overwrite it to each other.

Phoenix84:
Thanks, this is probably exactly what I'm looking for.

PR #203, opened 2015-05-30
Introduced a base class for programmable block programs

lord-devious:
There are multiple advantages to have a predefined base class for all grid programs.

1. The original intent: To provide clear documentation as to what is available when you write your scripts.

2. Provide an automatically updated base class that Visual Studio or other external IDE users can derive their program templates from.

3. Reduce the number of reflection calls needed to initialize and run the program.

4. To have the last remaining reflection code extracted out of the programmable block code and into this base class, so each grid program has full control of its own function.


Sibz:
:+1:

kapitanov:
+1. That's a nice start for further PB improvements

OndrejPetrzilka:
Nice, will merge it when you update your branch (also @joeblack616  is making some small changes, so there will be another update)

Also you can use:
MethodInfoExtensions.CreateDelegate which creates delegate from MethodInfo and object instance.

lord-devious:
Thanks @OndrejPetrzilka, very good idea about CreateDelegate! I had forgotten about that 'till this morning when I had to use it somewhere else. Strange coincidences :smile:

lord-devious:
Merged and updated as requested by @OndrejPetrzilka

craigperko:
This changes the game dramatically. To summarize:

I can now get the programmable block from within the program.

I can now get elapsed time directly, rather than using ticks.

I can store a public string on a programmable block, and access it from other programmable blocks.

I can call a programmable block from a programmable block, specifying an argument.

The only thing I'm confused about is the Echo command. Is there somewhere I can see how that works?

lord-devious:
The echo command simply adds text to the detail area of the programmable block, so you can output debug information without the need of an LCD panel.

craigperko:
Excellent. I will make a video about it.

Thanks for your excellent work. This week's update made my drones and fleet management system tenable.

craigperko:
Ohhh, this is gonna need a little fix! When the program doesn't compile, it prints the last successful echo statement rather than a compilation failure notice.

Also, here's the quickie video about a security hole you may not have meant to leave. I like it, though.

https://www.youtube.com/watch?v=cZVYcziJ5ag&feature=youtu.be

lord-devious:
You said a few things that worried me. You claimed you can _set_ the storage string? I really hope not... As for the compilation fail thingie, that remains exactly as it was before I added the Echo feature - which was a while ago, actually, all this PR did was add a base class, all the other features were already there. The storage string, btw, is an official Keen thing. The rest is mine :smile: 

That security hole, unless I misunderstand you completely, has always been there. Once connected you can query out any owned blocks across all connected grids. However, to be sure, I'd like some more information?

lord-devious:
Thinking a bit more about it; you can't actually get the program of a programmable block, so there should be no problem altering things you shouldn't. Phew.

craigperko:
I set the storage no problem. I thought the whole point was to offer some persistent memory, since it can survive loading the game.

Sorry I assumed you introduced the silent fail bug, this is just the first time I noticed it.

Once connected to a GridTerminalSystem, you can save that system and continue to control it remotely even as you march across the cosmos. It offers a whole lot of functionality - huuuuge functionality such as:

Calling programming blocks on another ship.
Getting the location of another ship exactly, frame by frame.
Changing LCD screens.
Etc,etc.

Obviously good for people like me, but nonsensical within the fiction of the game.

lord-devious:
Yes, this is known as the Merge Block Exploit, I've been told. It was around long before the GitHub repo was even opened. I don't think I want the blame for that one :wink: 

You should be able to set the storage of the currently running (i.e. "your own") programmable block, not for any other. But you can't get the program of another PB so you can't change that, so I guess there's just some misunderstanding here, one way or the other :smile:

craigperko:
I guess I just never thought of it before. Anyway, this adds a lot of functionality, thanks for doing it. Don't suppose you have any plans to add current power consumption/thrust level to the detailed info of IMyThrusts? :P

PR #205, opened 2015-05-30
Single- and Multiline Text Box Fixes

lord-devious:
A bunch of quality of life fixes for the single- and multiline text boxes, mostly intended to improve the programmable block code editing but also improves text panels.

### DIFF
--- a/Sources/Sandbox.Graphics/Gui/MyGuiConstants.cs
+++ b/Sources/Sandbox.Graphics/Gui/MyGuiConstants.cs
@@ -686,6 +686,8 @@ namespace Sandbox.Graphics.GUI
         public const int TEXTBOX_MOVEMENT_DELAY = 100;
         //  Delay between we accept same key press (e.g. when user holds left key, or X key for a longer period)
         public const int TEXTBOX_CHANGE_DELAY = 500;
+        public const int TEXTBOX_INITIAL_THROTTLE_DELAY = 500;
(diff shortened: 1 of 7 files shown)


Sibz:
:+1:   Awesome work!

Phoenix84:
Does this fix the arrow keys jumping multiple character positions on a single press?

lord-devious:
Horizontally yes. You just reminded me that I intended to fix the up/down problem for the multiline text box as well. Back to coding :smile:

Phoenix84:
Excellent!
Do you think you could fix the mouse selection issue while you're at it?
When highlighting text, quite often the last few characters don't get highlighted if the mouse leaves the text box area. This is mostly when selecting from right to left.

lord-devious:
If anyone feels called upon to be an "outside" tester for this, so @OndrejPetrzilka and @No-Brain's job gets easier, it'd be appreciated.

THDigi:
Testing this in the PB's text area.

When holding a typed key down it seems to repeat fine (my keyboard repeat setting is at max) but the arrow movement is slower than that.

And while moving down with the arrow key you can press up to instantly go up, in reverse there's a delay.

lord-devious:
I just used Keen's original setup, although I sped up the repeat for the system keys a bit. I could use the system's repeat key setting but I don't feel it's worth the effort.

I'll check the extra delay.

PR #206, opened 2015-05-31
Add Weapon Group In Toolbar

yajiedesign:
now if some group contain weapon, it will show in Weapons and Tool category,
can drop it to toolbar and select all weapon at the same time.
can select gun and missile together.


THDigi:
Currently if you add a weapon group to the toolbar you'll use them like any other toolbar action, will this PR remove that functionality as well ?

yajiedesign:
@THDigi  no ,current function is not remove.current  weapon group is in "GROUPS",this PR weapon group is in " Weapons and Tool".

THDigi:
So when you make a group it will show up in weapons&tools as well as groups and if you add it through groups it will be just like before but if you add it through weapons&tools it will be a selectable item which you can use primary action to trigger the weapons/tools ? That'll be great :P

zrisher:
Great indeed, :+1:

Issue #81, opened 2015-05-18
Shouldn't MyTerminalControlCheckBox also inherit from MyTerminalValueControl?


lord-devious:
[This is a question for the devs]

So that any checkboxes in the UI get their terminal properties for ingame scripting, like the slider and color controls does?

mexmer:
I would extend that to On/Off switches ... since they are only true/false also.
I'm not sure how that will interfere with existing exposed fields tho'

lord-devious:
Agreed. I'd say all controls should, if possible.

lord-devious:
I wonder just how many times I'm gonna do that.

mexmer:
i think, we can close this issue now, since #207 is merged

lord-devious:
True.

PR #207, opened 2015-05-31
Terminalproperty bool value

mexmer:
bool is used for populating state of on/off button/checkbox for PB, so you don't need to compare WriteValue with "On/Off" value, which can be language specific.

On/Off button and checkbox will appear in list retrieved by IMyTerminalBlock.GetProperties, and you can evaluate it using GetValue<bool>("name") or modify using SetValue<bool>("name", true | false)

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Interfaces/ITerminalProperty.cs
+++ b/Sources/Sandbox.Common/ModAPI/Interfaces/ITerminalProperty.cs
@@ -53,6 +53,11 @@ namespace Sandbox.ModAPI.Interfaces
             return property.As<Color>();
         }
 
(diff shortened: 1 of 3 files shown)


mexmer:
this should be implementation of issue #81

Sibz:
((IMyFunctionalBlock)block).Enabled does the trick
Only functional blocks can be turned on/off

Sibz:
To Early in the morning, I see this is a PR :P 
Looking at the code I see it is useful in that you can easily get bool property values. My snippet above still useful for On/Off property but this PR will be useful for any other bool properties.

lord-devious:
Looks good to me :smile:

PR #209, opened 2015-06-01
Fix some memory leaks.

RossM:
Events and object pools are both very prone to causing memory leaks. Be careful with them.


OndrejPetrzilka:
Thanks, render programmer wants to check it out first.

OndrejPetrzilka:
Reviewed by render programmer

PR #210, opened 2015-06-01
Added 2 decimal places to Oxygen Tank capacity 

Geneticus:
Due to Oxygen Farm production being as slow as it is. The time for 1 Farm to add 0.01% to a tank made it difficult to tell if things were working and connected correctly.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyOxygenTank.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyOxygenTank.cs
@@ -369,7 +369,7 @@ namespace Sandbox.Game.Entities.Blocks
             }
             else
             {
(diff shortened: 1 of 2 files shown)


Geneticus:
I see that. 1st commit via Git, lessons learned :)

mexmer:
@Geneticus user.props is in .gitignore so, you can modify it outside project as much, as you want, and will never be pushed, but when you touch it with visual studio. that's different matter. it become part of solution.

i had this few times with other SCM, well except TFS ... but TFS is microsoft thing.

Geneticus:
So the question is should I leave it as is? Close it, revert,  and reissue the pull or other?

mexmer:
just revert change in solution file, that's all.

thelastWallE:
Or you can exclude user.probs in the Team Explorer if you edited or created the file in Visual Studio.

PR #212, opened 2015-06-01
[FEATURE] Improved asteroid/ore spawn modding

stvrbbns:
- MaterialType and SpawnFrequencyWeight properties were added to the VoxelMaterialDefinition so that materials defined in VoxelMaterials.sbc can specify how they spawn as part of a generated asteroid (surface, core, deposit) and with what weighted frequency relative to other materials of the same type.
- The MyCompositeShapes.Generator() function, part of the code pipeline which generates asteroids, was updated to properly use/apply the new properties.
- VoxelMaterials.sbc was updated to correctly define the two new properties for all existing voxel materials (although the properties do have default values).
- Entities/MyMeteor.cs#L442 was **NOT** altered. I'd consider removing the IsRare requirement because SpawnsFromMeteorites should be sufficient to define use of a voxel material in meteorites.

### DIFF
--- a/Sources/Sandbox.Game/Game/World/Generator/MyCompositeShapes.cs
+++ b/Sources/Sandbox.Game/Game/World/Generator/MyCompositeShapes.cs
@@ -479,10 +479,33 @@ namespace Sandbox.Game.World.Generator
                     for (int i = 0; i < material.SpawnFrequencyWeight; i++)
                         m_coreMaterials.Add(material);
                 }
-                else // Deposit
+                else if (material.MaterialType == "Deposit")
                 {
                     for (int i = 0; i < material.SpawnFrequencyWeight; i++)
                         m_depositMaterials.Add(material);
+                }
+                else if (material.MaterialType == "Unspecified") // Backwards Compatibility
+                {
+                    if (material.MinedOre == "Stone") // Surface
+                    {
+                        material.MaterialType = "Surface";
+                        for (int i = 0; i < material.SpawnFrequencyWeight; i++)
+                            m_surfaceMaterials.Add(material);
+                    }
+                    else if (material.MinedOre == "Iron") // Core
+                    {
+                        material.MaterialType = "Core";
+                        for (int i = 0; i < material.SpawnFrequencyWeight; i++)
+                            m_coreMaterials.Add(material);
+                    }
+                    else // Deposit
+                    {
+                        material.MaterialType = "Deposit";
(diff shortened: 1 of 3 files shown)


stvrbbns:
Updated this pull request to be compatible (no conflicts) with SE 1.085.

No-Brain:
Is this backwards compatible?

stvrbbns:
It was mostly backwards compatible, but in reviewing it again one more time I found I'd missed a spot. I have a fix and am working on testing it.

stvrbbns:
Fixed and tested.

RossM:
Note that this will conflict with pull #58.

stvrbbns:
Thanks, @RossM, I had not seen that. I did not realize someone else was already working to improve the asteroids. I was working on adding changes that would leave things as they are but open up significant alterations to modders. Importantly, I want to be able to control both where voxels spawn (surface vs. core vs. deposits -- average accessibility) and overall frequency (percent of overall voxels) for modded voxel materials with modded ores - so no relying on MinedOre or anything hard coded. I was working towards that in baby steps as time permitted.

Could we get that configuration/configurability into your pull request/branch? A lot of your PR looks good, except for the hard coding of the element weights and asteroid classes. I do like the concept of asteroid classes though. Would you be okay with having the "generating materials" section reworked to use a set of asteroid classes {surface materials with frequency, core materials with frequency, deposit materials with frequency, and deposit multiplier} instead of hard coding the classes and instead of any programming concept of "element weight"?

I assume it is possible to pull request into a pull request (they're all just git branches, right?), although I've not attempted to do so before.

RossM:
I think that my pull request definitely needs more configurability. I haven't given a lot of thought to exactly how that should work. As I noted in my pull request, I don't want to make major changes to that branch until KSH says something either way.

I'd recommend leaving this pull request as is, so it can be merged if my change is rejected, and creating a second branch. You can then merge both my branch and this branch into the second branch, and fix things up however you want.

I also suspect that planets are going to make things more interesting for material modability, since they might have their own ore frequencies. 

As a side note, the reason for the element weight thing in my pull request is to simplify balancing. Originally I tried to work out individual frequencies for each element that would keep the overall distribution roughly the same, but it got way too complicated. I realized that by grouping elements of similar weight it was much easier to have them all balanced, since I only needed to worry about balancing three types instead of nine.

stvrbbns:
@RossM That sounds like a fine plan.

stvrbbns:
This PR is once again up to date.

FavioGalvis:
:+1:

PR #214, opened 2015-06-02
access to connected laser antenna block

mexmer:
Access to antenna block on other end with proper LoS check and ownership check
IMyLaserAntenna interface in modapi.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMyLaserAntenna.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyLaserAntenna.cs
@@ -1,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Sandbox.ModAPI
+{
+    /// <summary>
+    /// Implements laser antenna block access
+    /// </summary>
+    public interface IMyLaserAntenna : Ingame.IMyLaserAntenna
+    {
+    }
+}
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyLaserAntenna.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyLaserAntenna.cs
@@ -12,12 +12,9 @@ namespace Sandbox.ModAPI.Ingame
     public interface IMyLaserAntenna : IMyFunctionalBlock
     {
         /// <summary>
-        /// get target coordinates
+        /// get target coordinates (r/o)
         /// </summary>
-        Vector3D TargetCoords
-        {
-            get;
-        }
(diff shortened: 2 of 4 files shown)


OndrejPetrzilka:
It's same as merge block exploit. Once PB gets antenna, it can access it even when it's out of sight or out of range.

mexmer:
hmm, not sure, where test for this can be. since same is happening with ingame terminal.
if you loose LoS, it's not problem, you still have control. only if you change ownership of antenna, or force disconnect, you loose control.

anyway to test on grid, or block level, if caller is on same grid? and if not, if has valid connection?

MeusRex:
Hmm, I see you have the same problem that I had with the first iteration of this: https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/75

To solve this you'll probably have to build a wrapper around the other antenna.

mexmer:
@MeusRex , ownership is not problem, check is there, problem is (as @OndrejPetrzilka ) pointed out LoS, or even disconnection and this applies to any remote access to block. If you can store reference to block and restore it later, even without connection ... guess you see the issue. But this issue is global problem of any method, that retrieves blocks from temporarily connected grids.

Even if you do proper check at function level (what i did in otherantenna getter, there is los check, and owner ship check), after you go beyond (retrieve grid for antenna), there are no longer proper checks.

not sure, how this situation can be resolved with major rework of all functions for ingame interface.

MeusRex:
That's what I wanted to tell you, you can never return one of the basic interfaces or blocks of another grid. Instead you'd return a wrapper containing the object.

In this case you would need another property, which returns IMyLaserAntennaWrapper. This new interface would always check if the connection is still standing and only then would it execute the functions.
While the player could store the wrapper, he would never gain direct access to the IMyLaserAntenna, therefore he would have no way of circumventing the checks.

The only problem is, that you'll need an additional class and interface which creates some overhead.

mexmer:
@MeusRex as to remote wrapper, that effectively limits remote block to only few actions, while in this case is desired to have normal access, like you have with terminal.

MeusRex:
Ya. Though you could extend the interface to contain IMyTerminal methods.

mexmer:
which gets you back to ... my terminal returns blocks, and you want to be able to enumerate blocks. so you need to create wrapper for any and all of block interfaces available in ingame api.

MeusRex:
Idd. That's why I changed my PR from direct access of the other grid to sending messages to a PB on the other grid.

mexmer:
closing this PR, since there is no suitable solution for problems with accessing stored blocks without LoS

PR #215, opened 2015-06-02
Modapi laserantenna interface

mexmer:
modapi scripts interface for laserantenna
added RequireLoS bool, to check LoS requirement on moded antenas

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMyLaserAntenna.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyLaserAntenna.cs
@@ -1,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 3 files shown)


Phoenix84:
I didn't see it until after it was merged.
I don't see the point of it.
LoS is a block property, which is available in either the ObjectBuilder or BlockDefinition for it (I forget which).
Either of those would be available in the ModAPI.

mexmer:
@Phoenix84 you can't access laserantena in scripts without interface, and objectbuilder is not suitable interface for scripts.
as for los, it's so people can make complementary scripts to moded blocks.

Phoenix84:
"objectbuilder is not suitable interface for scripts"
Why not? It's static data that won't change. Maybe I just don't understand the ModAPI structure then.

PR #216, opened 2015-06-02
Feature #49 extracted: Part 1, ToolbarItem code sharing and MP sync + timer bug fixes

lord-devious:
As requested, this is part 1 of #49, only containing the refactor and bugfixes.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/ToolbarItem.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/ToolbarItem.cs
@@ -9,7 +9,7 @@ namespace Sandbox.Game.Entities.Blocks
     [ProtoContract]
     internal struct ToolbarItem
     {
-        public static ToolbarItem FromObject(MyToolbarItem item)
+        public static ToolbarItem FromItem(MyToolbarItem item)
         {
             var tItem = new ToolbarItem();
             tItem.EntityID = 0;
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/ToolbarItem.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/ToolbarItem.cs
@@ -1,0 +1,80 @@
+﻿using System.Collections.Generic;
+using ProtoBuf;
+using Sandbox.Common.ObjectBuilders;
+using Sandbox.Game.Entities.Cube;
+using Sandbox.Game.Screens.Helpers;
+
+namespace Sandbox.Game.Entities.Blocks
+{
+    [ProtoContract]
+    internal struct ToolbarItem
+    {
+        public static ToolbarItem FromObject(MyToolbarItem item)
+        {
+            var tItem = new ToolbarItem();
+            tItem.EntityID = 0;
+            var terminalItem = item as MyToolbarItemTerminalBlock;
+            if (terminalItem != null)
+            {
+                var block = item.GetObjectBuilder() as MyObjectBuilder_ToolbarItemTerminalBlock;
+                tItem.EntityID = block.BlockEntityId;
+                tItem.Action = block.Action;
+                tItem.Parameters = block.Parameters;
+            }
+            else if (item is MyToolbarItemTerminalGroup)
+            {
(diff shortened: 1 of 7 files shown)


lord-devious:
STAND BY... While all other actions now seem to synchronize their action arguments, the remote control waypoints refuses to.

Mihara:
...Please correct me if I'm wrong, is the bug this PR aims to fix the reason that programmable block arguments don't appear to work for players other than the owner of the script when assigned to buttons?

lord-devious:
Yes, and the timer issue with the same thing.

lord-devious:
The reason I was so convoluted is that there is a better explanation in the original PR :smile:

Mihara:
Much kudos to you kind sir, this particular bug is a pain in the posterior. :)

lord-devious:
Weeeellll seeing as I'm the author of that particular feature in the first place... I'm actually a major cause of that particular pain :wink: Except the timer thing, that one I won't take the blame for, it behaves differently than the others.

aaraujo666:
@lord-devious Since you seem to be concentrating on QoL improvements, may I suggest you take a look here: http://www.reddit.com/r/spaceengineers/comments/384q9t/changes_id_like_to_see_in_space_engineers_wall_of/ ?

Awesome work!

lord-devious:
@aaraujo666 Take a look at
http://forums.keenswh.com/threads/tutorial-the-programmable-block-argument.7360794/#post-1286777803

Some of the PB issues raised are explained there.

aaraujo666:
@lord-devious I saw that. I also saw that most of my concerns with the in-game script editor were addressed as well. I'm working on a monospaced font for the editor/LCD panels. If I get the graphical part of the font working, should I add to your repo and you can make the code changes to support it?

You're awesome!

lord-devious:
@aaraujo666 If you saw it before I added that link, you should take another look - I updated it with a tutorial on how to call another PB with an argument. I had forgotten that, so thanks for reminding me.

I think it would be a better idea to make a separate pull request for your font thing. I am most certainly willing to do any required code changes however. The biggest concern is that as far as I know, there's no way to add any graphics assets in the source. We _may_ however find a way around that, we'll see when the time comes.

lord-devious:
@aaraujo666 Someone else have been working on the same thing.
https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/107/files

Sources/Sandbox.Game/Game/Entities/Blocks/ToolbarItem.cs
```diff
@@ -0,0 +1,80 @@
...
+                    }
+                }
+            }
+            return item;
+        }
+
+        [ProtoMember] public long EntityID;
+        [ProtoMember] public string GroupName;
+        [ProtoMember] public string Action;
+        [ProtoMember] public List<MyObjectBuilder_ToolbarItemActionParameter> Parameters;
```
OndrejPetrzilka: Please put these to top and format attributes as we do

Sources/Sandbox.Game/Game/Entities/Blocks/ToolbarItem.cs
```diff
@@ -0,0 +1,80 @@
...
+                var block = item.GetObjectBuilder() as MyObjectBuilder_ToolbarItemTerminalGroup;
+                tItem.EntityID = block.BlockEntityId;
+                tItem.Action = block.Action;
+                tItem.GroupName = block.GroupName;
+                tItem.Parameters = block.Parameters;
+            }
+            return tItem;
+        }
+
+        public static MyToolbarItem ToItemFor<T>(ToolbarItem msgItem) where T : MyCubeBlock
```
OndrejPetrzilka: Please make it non-generic, it should be enough. Only generic usage is TryGetEntityById, but it still uses only CubeGrid property. Should be enough to TryGetEntityById<MyCubeBlock>.

lord-devious:
I'll get to it ASAP (those attributes are a bit strange, I don't use that format myself...)

lord-devious:
Be aware that these final changes were made on a computer where I cannot actually _run_ the game!

PR #220, opened 2015-06-02
Component profiling

Tyrsis:
Gamelogic is one of the biggest strain on servers and even in large single player games.  There are blocks that are specifically causing issues, and there needs to be a way to profile them.  This adds the ability to capture statistics on every component run during the Update of a session (most expensive processing wise) to try and determine which session components are causing the most issues, and to look into their processing / limit their usage if possible.


RossM:
If you want detailed profile results, the Visual Studio profiler does an excellent job.

zrisher:
You can't run the visual studio profiler on a production dedicated server though - this would allow admins and modders to test the performance impact of their session components in real conditions. I don't know enough about C to say what the additional overhead of two != null calls every simulation would be, but if it's minimal enough then this would only be a drag on performance if you were actually using the profiling. :+1:, this would be amazing.

Tyrsis:
Yes, the reason for this is because the github source can not be used as a live server for 2 reasons.  a) It's 2 versions ahead of the release, and is not compatible and b) it's only the 32bit version which doesn't work well with large servers.  This change is required to monitor components in real world environment.  I desperately wanted the ability to profile with visual studio or other profilers, but because the source is so disjointed, that's not really an option.  

The null check per component is absolutely nothing in the grand scheme of things and opens up a lot of monitoring for servers.

zrisher:
If you have time @Tyrsis, could you additionally provide a small example of using this event from a mod's namespace to profile all the running components, or the mod's own? I think that would be really nice for people looking to use this - I know the code's probably very simple but it would still be helpful.

RossM:
I dunno why they don't have the stuff for an x64 compile on github. Anyways, I made a branch that will compile the dedicated server for x64, at the version I think they cut the current release at, so the current client should be able to work with it: RossM/release@af186b0c6a8b5ee83a82c39d4bb1879ec5c8d060. You should be able to use that profiling.

MeltedLux:
@RossM It is due to licencing not allowing it I believe they talk about it in the readme.

RossM:
Ah. I'll fix my branch to not include the x64 files then. You can copy them from the release version.

Tyrsis:
The current client (1.084.011) does not work with the github version.  This is due to sync packet changes.  Go and try it yourself (I try every single patch).  Setup a dedicated server and connect with a real client, in a world with 100+ grids, stations and players.  I bet you didn't try, but every single patch something else changes just enough that it makes testing a production server impossible.

No-Brain:
I don't see why you couldn't use the profiler we have. VRage.MyCompilationSymbols.RenderProfiling = true;

Tyrsis:
For the following reasons:

a) I've said this for at least 2 months now, but it is impossible to run the git source as a live server due to sync packet changes.  This version the spawn list does not show when a user logs in.  Last version asteroids did not get created due to changes.  The version before that configuration was all different.  Every single version on git contains changes that break real clients.  The only way to test issues is to have a full server of 20+ players.  Honestly this wouldn't be an issue if you guys would do trunk releases of the game source for patch/hotfix.

b) Because of a, attaching a profiler is not easy.  I can't run gitsource server, and can't attach a profiler to a release server.  So that leaves me with no chance of using the built in profiler.

c) I don't want to profile everything, just game logic.  And not every single frame either, just once in awhile.

d) You even said yourself the profiler is horrible on performance in the way you guys have it.

There are performance issues that only show up as more users come on.  I want to see if it's possible to alleviate that CPU pressure.  But there is absolutely no ways of figuring out what is going on performance wise without the ability to do even a minimal amount of timing tests.

zrisher:
Getting at this info would be a game changer for people running dedicated servers. It's currently very, very difficult to support even a small population of users without gamelogic time spiraling out of control. This would give admins the insight they need to start fixing that problem and would be a huge boon for the SE multiplayer community.

mexmer:
@Tyrsis as for version, usualy tuesday release is in sync with github ... unfortunately that quite offten can't be said on friday hotfix release.

Tyrsis:
@mexmer They are not.  The version on github is usually a version (and in this week's case 2 versions) ahead of the release.  Believe me I have tried for MONTHS now to get a working github server to run with a release client unsuccessfully.  So unfortunately, that's incorrect.

RossM:
I'm not opposed to this change. It would be nice to also be able to use the VS profiler, since it gives much more information.

I assume you tried building the github branch at an earlier change that corresponds to the release? It should be either the change that changes MyFinalBuildConstants to increment the version number, or the change before that one. So to build 1.085, you'd want to build either KeenSoftwareHouse/spaceengineers@3cee7980f864bd2b2318b29e3695a0c4bd7c09f0 (with the version changed) or KeenSoftwareHouse/spaceengineers@a0c2288de9b349b940912b784cea199cb1d556a6.

Having KSH tag the release branch would be ideal

Tyrsis:
I'm closing this.  After doing some AOP on the release Sandbox.Game.dll, 95% of the gamelogic is coming from blocks (as expected) and this change doesn't really allow you to profile blocks/block types individually, so this change wouldn't really reveal much except that the MySector session component is responsible for most of the processing issues, and digging down from there is not possible due to how MyEntity is implemented.

PR #223, opened 2015-06-03
Dx11 memory check fix

mexmer:
seems there are issues with DX11 memory check
dx render code checks for wrong value, which effectively pulls out discrete graphic on dual graphic notebooks. i've changed memory and added some comments with explanation.

still have not figured out, why DX11 is not intializing nvidia adapter in DX11 on notebook with switchable graphic.

### DIFF
--- a/Sources/VRage.Render11/Render/MyRender-Adapters.cs
+++ b/Sources/VRage.Render11/Render/MyRender-Adapters.cs
@@ -160,8 +160,6 @@ namespace VRageRender
                     adapterTestDevice.CheckThreadingSupport(out supportsConcurrentResources, out supportsCommandLists);
                 }
 
-//                void* ptr = ((IntPtr)adapter.Description.DedicatedVideoMemory).ToPointer();
-//                ulong vram = (ulong)ptr;
                 // DedicatedSystemMemory = bios or DVMT preallocated video memory, that cannot be used by OS - need retest on pc with only cpu/chipset based graphic
                 // DedicatedVideoMemory = discrete graphic video memory
                 // SharedSystemMemory = aditional video memory, that can be taken from OS RAM when needed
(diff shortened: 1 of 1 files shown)


THDigi:
Tested this on my integrated Intel HD 4600 and it allows it to use DX11 now, without this it reverts to DX9.

mexmer:
ondrej, can you please ask guys who doing dx11render also to test it, because this
```
supportedDevice = supportedDevice && vram > 500000000;
```
was added from your internal merge, before there was not this test

anyway, tested on my home pc with HD7660D (A10 APU), and R7-260x, it works also

and patch was made on HP Envy with intel HD4600 and GT750M

OndrejPetrzilka:
Are you sure UInt32 is good for storing memory? What if someone has more than 4 GB? Wouldn't be better to put there ULong?

mexmer:
that values comes as IntPtr (in msdn docummentaton it's UIntPtr) old conversion (from pointer to ulong) created unreasonable high numbers (because intptr is signed 32bit on x86) ... and resulted in having like 16TB of video memory ....
i will look into this and find better solution for conversion ... you are right about 4GB limit problem.

please note 
this is definition from MSDN
```
typedef struct DXGI_ADAPTER_DESC {
  WCHAR  Description[128];
  UINT   VendorId;
  UINT   DeviceId;
  UINT   SubSysId;
  UINT   Revision;
  SIZE_T DedicatedVideoMemory;
  SIZE_T DedicatedSystemMemory;
  SIZE_T SharedSystemMemory;
  LUID   AdapterLuid;
} DXGI_ADAPTER_DESC;
```

OndrejPetrzilka:
All you need to do is correctly convert IntPtr to long on both platforms.
IntPtr.ToInt64 is not working?

mexmer:
IntPtr.ToInt64() returns incorrect value. i've updated conversion, and now it seems working, need retest on 64bit build tho'

OndrejPetrzilka:
Oh, I see, it's PointerSize, this works for me, does it work for you?

    PointerSize a;
    long b = a; // Implicit cast

It works for small and large values even for small and large negative values. If not, can you paste here which values are problem?

mexmer:
```
                long svramsize = adapter.Description.SharedSystemMemory;
                long vramsize = adapter.Description.DedicatedSystemMemory != 0 ? adapter.Description.DedicatedSystemMemory : adapter.Description.DedicatedVideoMemory;
```
```
		vramptr	0x7c4df000	void*
		vram	2085482496	ulong
		svramptr	0x8001f000	void*
		svram	2147610624	ulong
		svramsize	-2147356672	long
		vramsize	2085482496	long
```

PR #225, opened 2015-06-03
Advanced Door

Eikester:
A more advanced version of the Interior Door:
-Infinite Subparts
-All subparts can slide in any direction and rotate around any Axis
-Subparts are defined in the Block Definition file, no need for Dummies
-Modder can define an "Opening Sequence" in the Block Definition

Example Mod including Source Files: http://steamcommunity.com/sharedfiles/filedetails/?id=453603250
Example Gifs:
Iris Door http://imgur.com/ANwFdqx
Safe Door http://imgur.com/qS4AT7T

### DIFF
--- a/Sources/Sandbox.Game/Definitions/MyAdvancedDoorDefinition.cs
+++ b/Sources/Sandbox.Game/Definitions/MyAdvancedDoorDefinition.cs
@@ -1,0 +1,33 @@
+﻿using Sandbox.Common.ObjectBuilders;
+using Sandbox.Common.ObjectBuilders.Definitions;
+
+using VRageMath;
+using Sandbox.Engine.Utils;
+using VRage.Utils;
+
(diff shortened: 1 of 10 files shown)


OndrejPetrzilka:
I'm impressed, I like the idea that it's completely separate and does not interfere with any other feature or existing door.

OndrejPetrzilka:
Thanks @yajiedesign for point out the issue, it was fixed by us.

@Eikester you'll have to rename dummies in model, they should be named "advanceddoor" instead of "terminal". The "terminal" for regular door is hack which needs to be there for backwards compatibility.

Eikester:
ah i see, i will update the example mod. Also thanks for merging :)

wardmatt1:
You should look into finding a way to set a custom action when you press the button to open the door.

PR #226, opened 2015-06-04
Allows turning symmetry planes on/off in Advanced Settings

aaraujo666:
This allows the player to turn the drawing of the symmetry planes on and off in the Advanced Settings screen. It only affects the drawing of the planes when in building mode, it doesn't affect the drawing of the planes when defining symmetry ('M' key).

This is a popular request by players and some players even experience headaches/migraines when playing with symmetry planes showing.

### DIFF
--- a/Sources/Sandbox.Common/ObjectBuilders/MyObjectBuilder_SessionSettings.cs
+++ b/Sources/Sandbox.Common/ObjectBuilders/MyObjectBuilder_SessionSettings.cs
@@ -233,6 +233,11 @@ namespace Sandbox.Common.ObjectBuilders
         [ProtoMember]
         public int PhysicsIterations = 4;
 
+        [ProtoMember]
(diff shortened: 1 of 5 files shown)


plaYer2k:
Instead of turning it on and off, would a transparency/opaqueness slider be possible? I like to have it around but toning it down a bit would be nice.

Spartan322:
I could easily add the alpha parameter but I will first see if this gets merged.

OndrejPetrzilka:
Would you consider put it into Game Options as a slider?

aaraujo666:
A slider for the alpha value? Yes. I could do that.

aaraujo666:
Would you prefer eliminating the check box "Show Symmetry Planes" and adding an alpha slider for symmetry planes, or have both? 

I, personally, think having both might be better, however redundant that might seem, since some players may have no idea what "Symmetry Plane Alpha" might mean.

PR #231, opened 2015-06-04
Added IMyCryoChamber interface, and show occupant in DetailInfo

lord-devious:
Adds an interface to the cryo chamber, where the first members lets you determine if a cryopod is occupied, and if so, get basic information about them (currently faction tag and name).

Also displays the name of the occupant in the detail area of the cryopod.


lord-devious:
Oops forgot localization...

lord-devious:
Don't want to use player, because immersion. Don't want engineer either for the same reason, IRL the occupant of these won't necessarily all be engineers. So the closest I managed to get was Astronaut. (this struct may be used for more than just the cryo chamber)

mexmer:
let them be a spaceman then. we all spacemen

zrisher:
But some of us are space women. : )

sigboe:
From Wikipedia:
An astronaut (or cosmonaut) is a person trained by a human spaceflight program to command, pilot, or serve as a crew member of a spacecraft. Although generally reserved for professional space travelers, the terms are sometimes applied to anyone who travels into space, including scientists, politicians, journalists, and tourists.

Mihara:
Use "occupant". It might not be a human. :)

THDigi:
Yeah, OccupantInfo or PassengerInfo (might be used on some other things in the future), maybe we'll be able to store dogs in the cryo pod :P

lord-devious:
Like I said, it's intended to be usable from more than just this block, the next use may not be occupant.

Mihara:
Contents then. :)

lord-devious:
To be readded later

PR #232, opened 2015-06-04
Added PB method for notifying seat/cockpit/RC controller

THDigi:
- Added `IMyShipController.NotifyPilot()` method which can be used in programmable block scripts to send notification to the player controlling it; It has 500 character limit and can be shown for max 30 seconds, let me know if these limits are too low or too high xD
- Added `IMyShipController.IsMain` which returns if the controller block is marked as the main one;
- Moved the rest of the mod API code from MyShipController.cs to MyShipController_ModAPI.cs.


Sibz:
:+1:  :+1:  :+1:

JustBurn:
This can be useful as an "audible" critical warning for the player that something is wrong with his ship, so, :+1:

PR #233, opened 2015-06-04
Removing redundancies in code, replace some code with LINQ equivalent…

Aminion:
…, fixing code formatting in VRage.Network

### DIFF
--- a/Sources/VRage.Network/MyRakNetClient.cs
+++ b/Sources/VRage.Network/MyRakNetClient.cs
@@ -1,16 +1,12 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
-using System.Net;
-using System.Text;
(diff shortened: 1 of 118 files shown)


mexmer:
first, don't touch network code
second, linq is not used due perfomance

lord-devious:
Performance and memory allocations.

mexmer:
too much memory allocations affects performance ;o)

RossM:
Rule 1 of high-performance C# is don't allocate objects inside loops. The object allocation itself is super cheap, but the extra load it puts on the GC will kill you.

JCMais:
I don't know about Keen, but the rule of thumb for the majority of other projects I've seen, is that You should avoid pull requests that change too many code at the same time, stuff like that is very prone to merge conflicts. :p

No-Brain:
Yeah no.

PR #234, opened 2015-06-04
BugFix for MyPlugin

sjsoft2015:
1) Keeping all switches that run the process, when it is restarted.
2) Removed pieces block raises an exception when loading the switch
-plugin

### DIFF
--- a/Sources/VRage.Dedicated/Configurator/ConfigForm.cs
+++ b/Sources/VRage.Dedicated/Configurator/ConfigForm.cs
@@ -68,7 +68,8 @@ namespace VRage.Dedicated
             }
             else // Local / Console
             {
-                Process.Start(MyPerServerSettings.GameDSName + ".exe", "-console -ignorelastsession");
+                string[] cmdLine = Environment.CommandLine.Split(null as string[], 2, StringSplitOptions.RemoveEmptyEntries);
+                Process.Start(cmdLine[0], cmdLine[1] + " -console -ignorelastsession");
                 Close();
             }
         }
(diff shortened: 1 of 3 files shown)


sjsoft2015:
need update. not worked any more with -ignorelastsession. 
                using (MySandboxGame game = new MySandboxGame(services, new string[] { }))
running without arguments

sjsoft2015:
Fixed

Sources/VRage.Dedicated/Configurator/ConfigForm.cs
```diff
@@ -68,7 +68,8 @@ private void startButton_Click(object sender, EventArgs e)
             }
             else // Local / Console
             {
-                Process.Start(MyPerServerSettings.GameDSName + ".exe", "-console -ignorelastsession");
+                string[] cmdLine = Environment.CommandLine.Split(null as string[], 2, StringSplitOptions.RemoveEmptyEntries);
+                Process.Start(cmdLine[0], cmdLine[1] + " -console -ignorelastsession");
```
THDigi: I got an exception when trying to run the compiled DS, I reverted these 2 lines to the old line and it works again.
Exception: http://pastebin.com/raw.php?i=BkY8Jka4

sjsoft2015: You right! This happens when no arguments in command line..

sjsoft2015:
Process.Start(cmdLine[0], ((cmdLine.Length > 1) ? cmdLine[1] : "" ) + " -console -ignorelastsession");

sjsoft2015:
#264

PR #235, opened 2015-06-05
Relayer HUD Markers

andrew-w:
Re-ordered draw calls in MyGuiScreenHudSpace.cs so that marker
indicators (GPS Locations, Ore Deposits etc) are rendered behind UI elements such as the toolbar.

Added OverrideDrawElements and OverrideDrawControls boolean values to
MyGuiScreenBase so that inherited controls can choose to handle drawing
them instead of letting the base class do it.  This is to facilitate the
change to MyGuiScreenHudSpace.cs so that markers can be rendered behind
the UI elements/controls.


JJRcop:
Feedback from #187 applies

PR #236, opened 2015-06-05
Expose more types and properties as public

dodexahedron:
SEServerExtender uses this field quite a bit. Let's expose it as a public, read-only property, so I don't have to reflect against it any more.


dodexahedron:
Also, if there's a better alternative to this field, I'm all ears.

dodexahedron:
There's a reasonable chance I'll come across more of these as I strip the reflection out of SESE.
These were just the immediate ones that will make a nice difference.
Thanks.

PR #241, opened 2015-06-05
Using ?? operator.

Aminion:


Spartan322:
I'm not sure this has much point, the result doesn't change, it really only removes a few extra characters from the source, and the ?? operator is a very exclusive shorthand.

MeusRex:
This won't make it in. It adds absolutely nothing, just like some of the early PRs of that other guy...

PR #244, opened 2015-06-05
Add inventory mass

IhanaMies:
Add real time mass recalculation to ships when Inventory Mass is enabled from Advanced World Settings.


IhanaMies:
Squashed commits + after squash clean up

Issue #243, opened 2015-06-05
MyAbstractXmlSerializer.cs


wardmatt1:
This file seems to have disappeared. the project wont compile with out it.

zrisher:
I know we were relying on some xml objects that were moved to the new
VRage.Game assembly on Thursday.

Sent from my iPhone

On Jun 5, 2015, at 12:36 PM, wardmatt1 <notifications@github.com> wrote:

This file seems to have disappeared. the project wont compile with out it.

—
Reply to this email directly or view it on GitHub
<https://github.com/KeenSoftwareHouse/SpaceEngineers/issues/243>.

sjsoft2015:
MyStringId now in namespace VRage.Utils

jkells:
The file was added to the project in this this pull request https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/70

Looks like the actual file wasn't committed though.

Tostito:
you can comment out the compile include MyAbstractXmlSerializer.cs file in the sandbox.common.csproj file.

mexmer:
@jkells no file was moved to vrage.game
unfortunately in #70 there was merged project file which relied on old file location (sandbox.game)

you can use following patch, to fix compilation issues
```
diff --git a/Sources/Sandbox.Common/ModAPI/IMyControl.cs b/Sources/Sandbox.Common/ModAPI/IMyControl.cs
index ee0a940..95c44d3 100644
--- a/Sources/Sandbox.Common/ModAPI/IMyControl.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyControl.cs
@@ -5,6 +5,7 @@
 
 using VRage.Input;
 using VRage.Library.Utils;
+using VRage.Utils;
 
 namespace Sandbox.ModAPI
 {
diff --git a/Sources/Sandbox.Common/ModAPI/IMyUserInput.cs b/Sources/Sandbox.Common/ModAPI/IMyUserInput.cs
index 494697c..6c1c19e 100644
--- a/Sources/Sandbox.Common/ModAPI/IMyUserInput.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyUserInput.cs
@@ -6,6 +6,7 @@
 using VRage.Input;
 using VRageMath;
 using VRage.Library.Utils;
+using VRage.Utils;
 
 namespace Sandbox.ModAPI
 {
[...]

PR #246, opened 2015-06-05
Gravgen moddability

N3X15:
# Changelog

## Executive Summary

 * Mods based on Gravity Generators can now define a default field size and strength.
 * Mods can now re-implement some gravity generator code, if needed.
 * Blocks using Gravity Generator behaviors are no longer forced to play the generator's droning sound.

## Details

 * Added MyBoundedVector3 for neatly bundling fieldsize bounds.
 * Added .GetOrDefault(defaultValue) to floats and SerializableVector3s, making detection of uninitialized MyObjectBuilders a bit more standardized (uses NaN rather than negative or huge values).
 * MyGravityGeneratorBase:
   * Set to public, so modders can inherit it from their own classes.
   * Changed the idle noise to be configurable via `<PrimarySound>`, like reactors.
 * MyGravityGenerator (and friends):
   * Set to public, so modders can inherit it from their own classes.
   * `<FieldSize>` (MyBoundedVector3) added to definition and object-builder.
     * Default:
       * Min: <0f,0f,0f>
       * Max: <150f,150f,150f>
       * Default: <150f,150f,150f>
   * `<Gravity>` (MyBounds) added to definition and object-builder.
     * Default: Min=-1f, Max=1f, Default=1f
 * MyGravityGeneratorSphere (and friends):
   * Made public.

## Rationale
 * Some mods need the ability to adjust the size of the default field.  This change also permits proper power usage scaling and balancing, as maximum and minimum values can be established.
 * Modders desiring the ability to change how gravity affects objects can do so without hacking the game itself, which is troublesome for multiplayer support, and a hassle for end-users.
 * This changeset also provides backwards-compatibility for older mods, establishing the current hard-coded values as defaults.
 * Future changes to block definitions that leave behind out-of-bounds savedata for fields/gravity strength will not cause buggy behavior, as this changeset adds sanity checks to deserialized save data for gravity generators.

# Test Script

1. Create a new mod with a gravity generator block. (I suggest the [GravTube](http://steamcommunity.com/sharedfiles/filedetails/?id=417028349) mod.)
2. Set the following in the MyObjectBuilder_GravityGeneratorDefinition:
```xml
<FieldSize>
  <Min x="0" y="0" z="0" />
  <Max x="2.5" y="2.5" z="2.5" />
  <Default x="2.5" y="2.5" z="2.5" />
</FieldSize>
<Gravity Min="-1" Max="1" Default="0.9" />
```
3. Save the mod.
4. Start a new game.
5. Ensure existing gravity generators on the map are at 150m field size in all dimensions and 1G strength, and that those are the maximum values.
6. Open G menu and equip the placer for the modded GravityGenerator block.
7. Place the block on a powered grid.
8. Ensure the placed block is set to 2.5m in all dimensions, with a minimum of 0m and a maximum of 150m.
  * Note:  the values shown in the panel are rounded, CTRL+click on the slider to see true value.
9. Ensure the placed block has 0.9G field strength, with a minimum of -1G and a maximum of 1G.


Phoenix84:
I really like this, and the work you put into it.
However with that class made public, it probably won't get merged.
I would suggest a smaller change, which only lets you mod the gravity force and field size.

Iamgoofball:
:+1: looks great

@Phoenix84 what's wrong with the classes being public for modders?

Phoenix84:
@lamgoofball, What other block classes in the ModAPI are public? I know of none.
I have no problem with it, I'm just saying that it's a paradigm shift from the current behavior.

d3athrow:
Come home senpai

N3X15:
@Phoenix84 It may be a shift from the prior behavior, but it opens up a vast new world of modding possibilities.  Modders would be able to extend existing MyFunctionalBlocks to get what they want.  We did this with Kerbal Space Program, and you can see the vast variety of mods available.  The more open your API is, the more vibrant the community becomes.

I can fix it, if requested by KSH, but I cannot do anything at the moment due to the repo currently being unbuildable. (#243, other errors)

harag-on-steam:
Wouldn't it be even more flexible to change the extend of each side of the gravity field independently? I.e. have it extend 100m above the gravity generator but 0m below it, for example.

Would that change be to invasive to implement? It certainly would required UI changes.

No-Brain:
Don't use NaNs.

N3X15:
Why?  floats don't support nulls, and float.Min/Max don't serialize properly for some reason.  I tested this in-game and didn't notice any issues.  Is there a problem serializing them for MP?

Phoenix84:
You can use a nullable bool.
Besides that, why not have default values?
Use 9.81 as the default gravity, and pick other sane defaults. Let the block definition override it.

No-Brain:
NaNs like to spread and break everything they touch. They are a pain to debug since the problems they cause are seemingly unrelated.

Use sane default values and make sure it is backwards compatible even with mods.

N3X15:
Okay, switched to nullable floats.  Let me know if this is acceptable.  I can't test since master is broken again.

No-Brain:
Why do you have uninitialized values?
Bounded vector will cause issues in multiplayer when multiple players try to change different components of if.

N3X15:
Uh, BoundedVector3 is only used in the block definition.  Players shouldn't be able to muck with that, at all.  If they can, something's seriously wrong. 

RE: Uninitialized values: SerializableBoundedVector3 was made using [SerializableBounds](https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/master/Sources/VRage/Serialization/SerializableBounds.cs) as a template, as I'm not too familiar with ProtoBuf.  If I misunderstood and you were talking about elsewhere in the code, please point it out and I'll fix it.

No-Brain:
Oh about the BoundedVector3, I was thinking about different PR.
Don't use nullable either, look at how other blocks do backwards compatibility.

N3X15:
The point of using NaN or NULL was to have a marker to tell the system that an object is new and needs to have the cube definition's defaults applied.  Hard-coding the values completely defeats the purpose of this PR, which is to let the CubeBlock definition define what the start-up size of a gravity field is.  

If I set MyObjectBuilder_GravityGenerator.FieldSize to 150f,150f,150f like it used to be, all gravity generators would start with those values, and users would have to edit the values after spawning the object.  This is because the game seems to create the world object with the ObjectBuilder.  The Definition doesn't do anything in this instance.    Therefore, the NaN or NULL tells Init() that there's a new object in the world and to initialize it with the FieldSize.Default value rather than some hardcoded values.

I've tested this just now by changing MyObjectBuilder_GravityGenerator.FieldSize to 150f,150f,150f, changing all the nullables to regular scalars, and getting rid of the GetDefault() calls. the gravity generator I'm messing with has 
```xml
      <FieldSize>
        <Min X="0" Y="0" Z="0" />
        <Max x="4" y="4" z="4" />
        <Default x="2.5" y="2.5" z="2.5" />
      </FieldSize>
```

When spawning the object in world, it starts with a fieldsize of 4^3, rather than the 2.5^3 fieldsize the CubeBlock definition specifies.

So, the only way to tell Init() that there's an uninitialized object sitting there is to feed it an invalid number.  Nullables cause the networking stack to freak out, so that leaves us with Infinities and NaN.  We could also set the whole FieldSize object to null, but I don't know how the networking stack will handle that.  What I do know is that NaN worked fine and didn't cause any unexpected behavior because the system REMOVED NaN when it was initialized. 

# tl;dr
  * Hardcoding the default value like other blocks is not going to work because it defeats the purpose of this PR (allowing MODDERS to set the default value)
  * Nullables are out of the question due to networking stack issues with them
  * Infinities cause issues with System.Xml serialization

No-Brain:
The object builder should have default values for the new fields representing the old behavior.

N3X15:
Yes, the [Definition ObjectBuilder](https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/246/files#diff-1fc456f21e85ebacb75802c8ed2f5a10R15) is set up to work that way.

N3X15:
Tested and working again.

Phoenix84:
*Hardcoding the default value like other blocks is not going to work because it defeats the purpose of this PR (allowing MODDERS to set the default value)*

Do you want runtime changeable definitions, or just added cubeblock definitions?
If the latter, look at every other block, they all work fine.
If the former, do the latter, then expose properties in the IMyGravityGenerator interface.

I think you're overthinking it.

N3X15:
@Phoenix84 I already explained this above.  The point is to allow CubeBlocks to define the maximum, minimum and default field sizes.  Current backwards-compatibility code in SE hardcodes those ranges, which is not desirable. 

However, since I was up until 5:00AM trying to explain this and have apparently gotten nowhere, I'm going to just move ahead with my pre-Open-Source plan of using a binary patcher to let users apply the mods they want.  It's getting old.

PR #247, opened 2015-06-06
More types needed by SEServerExtender

dodexahedron:
These types are used by SESE.
Please make public so I can remove the reflection around them.

### DIFF
--- a/Sources/Sandbox.Game/Game/Multiplayer/MyPlayerCollection.cs
+++ b/Sources/Sandbox.Game/Game/Multiplayer/MyPlayerCollection.cs
@@ -70,7 +70,7 @@ namespace Sandbox.Game.Multiplayer
         }
 
         [MessageId(9, P2PMessageEnum.Reliable)]
-        struct ControlReleasedMsg : IEntityMessage
+        public struct ControlReleasedMsg : IEntityMessage
         {
             public long EntityId;
             public long GetEntityId() { return EntityId; }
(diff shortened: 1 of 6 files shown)


yajiedesign:
@dodexahedron  add  
InternalsVisibleTo["SEModAPI"]
InternalsVisibleTo["SEModAPIExtensions"]
InternalsVisibleTo["SEModAPIInternal"] 
etc
?

dodexahedron:
That would help in quite a few places.
Does it help for private members/methods as well as internal?

yajiedesign:
private members/methods  can't

dodexahedron:
Then there will still be plenty of members that need to be exposed, including a lot of private nested classes/structs.

yajiedesign:
you can use Mono Cecil ,auto modify dll file after each update.

dodexahedron:
My biggest problem with modifying game libraries is that it means that every single update to Space Engineers will break SESE.
Right now, unless there is a breaking API change in something that SESE depends on (which doesn't happen THAT often), users do not need to update to a version of SESE compiled against the current release of SE.
Modifying game libraries would result in very tight version coupling, which was something we had in the days of obfuscation and was very frustrating to server operators (and to me, as the maintainer of SESE, as it then placed a demand on me to deliver an updated version quickly).  In fact, it would be worse, because even the friday patch would break it, and that almost never had broken SESE, before.
It's a neat idea, but I don't think it's the appropriate way to go.

yajiedesign:
Yes, this is a problem.Because SE updates frequently ,There is no perfect solution now.

dodexahedron:
Right.
And I have no problem reporting the types that need more exposure. I can wait for the next release to reflect those changes. It's not like I'm going to be able to complete all of the work on SESE in one week anyway. :smile: 
It doesn't really hurt my users for it to take longer to have it done in the official release.  Sure, it might delay my work on new features by a couple of weeks, but that's not really a big deal. Stability and availability are what users seem to care about most, at this point.

Issue #208, opened 2015-05-31
Checkbox actions similar to On/Off button


mexmer:
well, idea behind this is to have _Check and _Uncheck action, like we have _On and _Off

there might be situations, where you just want to turn something off (uncheck), or on, and just don't want to bother about state.

like for example script that will turn off broadcasting on all antenas, or lets say enable check for humans on all sensors, and so on

concept of this can be found here
https://github.com/mexmer/SpaceEngineers/tree/terminalproperty-checkbox-actions

i did this i byproduct of another thing i was working on, and want to know, if it's even desired, or if you think it should be done different way.

please note, enabling check/uncheck actions also add this as toolbar action, which might not be desired (too long actions menu on blocks)
see http://i.imgur.com/Xg9rzEu.jpg

also note, only block (as example) where this is enabled is antena.
you can easilly do this for any block just add .EnableToggleActions(); after any .EnableAction() on checkbox construction.

pros
- you can have 2 buttons, one for off, one for on
- you can force on/off action regardless current state (eg. no need for state check)

cons
- it's redundant (at least from pb point of view), since we have setvalue/getvalue (see #207, which is also integrated in branch mentioned above)

lord-devious:
I don't think there should be enable/disable actions, it feels redundant. Your #207 is a far superior way to do this, standing alone. IMO _all_ terminal controls should derive from that value base class and have a representative terminal property. Much better than actions.

lord-devious:
One thing though, I think it would be a good idea to use the same suffixes as the OnOff action does. It is also used for doors (Open_On, Open_Off) so the checkboxes should use *_On and *_Off rather than *_Check and *_Uncheck. The fact that they are checkboxes in the UI is really irrelevant.

mexmer:
@lord-devious good point about On/Off and Check/Uncheck ... was not sure about it myself, then i gone with logical name ... since it's checkbox it's check uncheck, tho' it effectively turns something or on off

as for redudancy, there are 2 points of view.
i'm not sure, if it can be done way that for toolbar it will add on/off action, but not for PB .... terminalproperty seems little mysterious to me ;o)

lord-devious:
I think my current conclusion is that once you standardize the suffixes to what already exists this is a good addition.

mexmer:
@lord-devious done

mexmer:
i've added onoff to few checkboxes (that i think people might want to have on button panels). removed imo unneccesary show shipname.

[...]

PR #250, opened 2015-06-06
Feature: Terminalproperty checkbox actions

mexmer:
This PR changes checkbox to behave same way as ON/OFF button in terminal property.
This allows 2 things
1) add and use _On/_Off action in PB
2) add and use On/Off action on checkboxes in Toolbar

checkboxes that should have On/Off still need add
MyTerminalCheckboxControl.EnableOnOffActions() upon construction of control object (which i added to some on my own)

It will also allow to have MyTerminalControlOnOffSwitch and MyTerminalControlCheckbox somewhat interchangeable (they have same methods now) ... difference between them will be mainly visual representation.


mexmer:
this PR is implementation of  #208

mexmer:
synced with current master (rebase)

mexmer:
rebased and squashed

mexmer:
not needed, because SetValueBoolean does this (true for on, false for off)

PR #251, opened 2015-06-07
IMyAssembler: Added queue management functions

### LINKED ISSUE #96: Assembler queue access with ingame script

Sibz:
Added the following functions:
```c# 
bool AddQueueItem(string itemType, string subtypeName, int amount);
int GetQueueCount();
IMyAssemblerQueueItem GetQueueItemAt(int idx);
void RemoveQueueItemAt(int idx);
void ClearQueue();
int CountQueueItems(string itemType, string subtypeName);
```
Added interface IMyAssemblerQueueItem

Added class MyAssemblerQueueItem to MyAssembler

Uses existing functions that are sync friendly for modifying the queue.

Compiles and tested OK.
In game test code: http://pastebin.com/ybWuqxTw

Closes #96

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyAssembler.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyAssembler.cs
@@ -21,11 +21,12 @@ namespace Sandbox.ModAPI.Ingame
         /// <param name="itemType">type of item, i.e. 'MyObjectBuilder_Component'</param>
         /// <param name="subtypeName">Shortname of component, i.e. SteelPlate, Reactor, MetalGrid, etc.</param>
         /// <param name="amount">Quantity of specified item to add to queue</param>
+        /// <param name="index">Index/Position item should be added to queue<br />-1 (default) adds to the end of queue<br />0 is first slot, 1 is second slot etc.</param>
         /// <returns>
         /// true - item added successfully
         /// false - item not added - probably because itemType or subtypeName are invalid
         /// </returns>
-        bool AddQueueItem(string itemType, string subtypeName, int amount);
+        bool AddQueueItem(string itemType, string subtypeName, int amount, int index = -1);
 
         /// <summary>
         /// Retrieves the number of items in the assembler queue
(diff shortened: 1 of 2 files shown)


Souper07:
Long overdue. Nice work!

joemorin73:
What would you think about adding an overload for AddQueueItem?

bool AddQueueItem(string itemType, string subtypeName, int amount, int idx);

Allow insertion within the queue.

Sibz:
@joemorin73 Yes I was thinking about that, it would be easy to do. I'll add that in tonight when I get home.

Sibz:
@joemorin73 Added as a optional parameter, default is -1 (add to end of queue).

joemorin73:
Excellent.  This should bring in some impressive ingame production scripts.

How I suck.  Any thoughts on Disassembly or how the code works when it's in Disassembly mode?

mexmer:
@joemorin73 there is not much diffence between assemble mode and dissassemble mode.

in dissasemble mode it "produces" ingots from components/tools, that's about it.

joemorin73:
True.  But there is nothing in the interface to check which mode the assembler is in and change the mode.

mexmer:
@joemorin73 but there is
https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/207
this allows you to read state of any On/Off or checkbox control as bool or using SetValue to change it.

joemorin73:
Ahhh!  Then if AddQueue can also add to the disassembly queue, we are set!

Aleks976:
Is there any word on whether these changes will be merged with master or not?

Sibz:
I am still hoping, theres not been much merge activity lately, I assume it because they are working hard to introduce planets and/or netcode.

paradineai:
I am not sure if this question is in the correct location but: would it be possible to get the ability to add/remove items to/from the production queue in the assembler added to the api so we can use an in game script to control the production of items directly just like we are able to change the image or text on an lcd panel? This would allow the people who are righting the scripts to make a quota management script to make the assemblers keep a certain amount of items available in their cargo containers inventory. For example if I made a script to allow me to keep a minimum of 10k steel plates in my inventory it would be able to check all of the cargo containers on my ship/station and count the number of steel plates. Then if I have >= 10k steel plates total it does nothing but once I remove enough to get below 10k steel plates it tells my assembler to make the number needed to bring it back up to a total of 10k. If this question is in the wrong location then I ask that KSH move it to the correct one and let me know of the error. Also I ask these questions because I am not very good at programing myself and would rather leave it to someone whom knows more than I do.

zrisher:
@paradineai:

**In general**, report issues (either Bugs or Feature Requests) in the [Issues section](https://github.com/KeenSoftwareHouse/SpaceEngineers/issues). Discuss implementations of bug fixes and features in their respective PRs, like this page. In this case, Sibz mentioned the issue this is meant to fix in his initial comment - #96. I would look there to see if the issue he's attempting to fix is the same you have. If it's similar enough but you had additional requests you can add them there, otherwise you can open a new one.

**In this case**, it looks to me like what you're asking for is exactly what he's implemented.

Sibz:
Conflicts, will re-do request.

PR #252, opened 2015-06-07
MyDoor: Fixed OpenRatio > 1

### LINKED ISSUE #204: [Bug] IMyDoor.OpenRatio:Returns value higher than 1
Tested on standard door, returns 1.2

Introduced with the feature in
2073ef1743fca874b9ac748c0ee07677e151b208

Sibz:
MaxOpen could be greater that 1 and the default is 1.2. Therefore the
return value for OpenRatio could also be greater than 1.
This change makes OpenRatio return a percent.
Fixes #204


lord-devious:
:+1:

PR #254, opened 2015-06-07
Welder helper sync

mexmer:
sync for ShipWelder.HelpOthers checkbox in MP/DS
checbox state now preserved trough save
exposed HelpOthers value in script interface


mexmer:
rebased onto current master

PR #256, opened 2015-06-07
fixed oxygen generator autorefill sync

mexmer:
fixed oxygen generator not using sync object for autorefill checkbox
exposed Autorefill value to scripts

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMyOxygenGenerator.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyOxygenGenerator.cs
@@ -8,12 +8,8 @@ namespace Sandbox.ModAPI
     /// <summary>
     /// Oxygen generator interface
     /// </summary>
-    public interface IMyOxygenGenerator : Ingame.IMyFunctionalBlock, IMyFunctionalBlock
+    public interface IMyOxygenGenerator : Ingame.IMyOxygenGenerator, IMyFunctionalBlock
     {
-        /// <summary>
-        /// Autorefill enabled
-        /// </summary>
-        bool AutoRefill { get; set; }
         /// <summary>
         /// Increase/decrese O2 produced
         /// </summary>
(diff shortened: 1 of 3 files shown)


mexmer:
rebased onto current master

alexflorea:
If I understand correctly, m_wantAutoRefill is used as a guard on clients to prevent them from spamming the button and sending many messages. While that's fine, this flag will get out of sync when syncing and when loading (STR: flag is false on both server and client, client switches to true, now on the server m_wantAutoRefill is false, but m_autoRefill is true and the server can't change it anymore). There might be other cases where this gets out of sync.

Also, I've seen this pattern used in other PRs, like this one https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/254

mexmer:
i've seen this pattern used in KSW code, that's why i decided it to use also.
if this causes problem, i can change it to immediate change.
I asked on skype, which approach should i use, and didn't got clear answer.
there are 3 different approaches to syncing checkboxes trough code.
I can adapt any of them

alexflorea:
Hi. If you could please point me to where this is used in the codebase, I will take a look. I did a quick look and couldn't find any. An immediate change would be better for now.
Thank you.

mexmer:
it's used in antenna (broadcaster) related blocks
for example
https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/master/Sources/Sandbox.Game/Game/Entities/Blocks/MyRadioAntenna.cs

mexmer:
as for what i asked on skype, here is snip from it
------------------------------
i have noticed there are multiple approaches to this. so i want to know, which one is preffered.

would be nice, if setters and getters are unified.

sometimes getter uses private member, while other times it used public member, sometimes setter calls sync object while other times it calls public member, which call sync object upon set, or doesn't use sync object at all.

also some times setter directly modifies values, while other times it modifes value m_wantsXXX, and sync object sets m_XXX.

also there exists variant with Sync.IsServer and SyncProxy check .... and few others

PR #257, opened 2015-06-07
MyCharacter: Added a delay before damage from oxygen loss

Sibz:
Added a delay based on a setting to prevent damage from oxygen loss for
a number of seconds. Default number of seconds is 5.

This is to imitate the fact that in real life you wouldn't take damage from oxygen loss/depressurisation straight away. Usually about 10 seconds of feeling quite uncomfortable come before any 'damage' (unless subject attempts to hold breath then lungs could rupture).

On another note, it should also resolve the issue of taking damage when opening a door within a pressurised structure.

### DIFF
--- a/Sources/Sandbox.Game/Definitions/MyCharacterDefinition.cs
+++ b/Sources/Sandbox.Game/Definitions/MyCharacterDefinition.cs
@@ -40,6 +40,7 @@ namespace Sandbox.Definitions
         public bool NeedsOxygen;
         public float OxygenConsumption;
         public float PressureLevelForLowDamage;
+        public int SecondsBeforeOxygenLossDamage;
         public float DamageAmountAtZeroPressure;
         public float OxygenCapacity;
         public string HelmetVariation;
@@ -143,6 +144,7 @@ namespace Sandbox.Definitions
             NeedsOxygen = builder.NeedsOxygen;
             OxygenConsumption = builder.OxygenConsumption;
             PressureLevelForLowDamage = builder.PressureLevelForLowDamage;
+            SecondsBeforeOxygenLossDamage = builder.SecondsBeforeOxygenLossDamage;
             DamageAmountAtZeroPressure = builder.DamageAmountAtZeroPressure;
             RagdollDataFile = builder.RagdollDataFile;
             HelmetVariation = builder.HelmetVariation;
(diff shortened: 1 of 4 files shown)


mexmer:
hmm, not sure, if this will not cause problem with tracking door damage issue.

i mean ... now you receive damage immediately, when you open doors - this issue has still not been solved (in some cases block, with door acts like block without oxygen when opened, for short time of period), it's easy to spot.
with your change ... well ...

also, imo this aspect really should not be part of realism, i will stick with gamification.

zrisher:
If this truly is more realistic, :+1:, would be really nice for those times when you accidently press j.

Sibz:
@mexmer Yes I understand it may make the tracking of 'damage when opening a door' issue difficult. It isn't the intention to fix the issue and to work on tracking the issue, the setting can be set to 0.

As for realism, it still isn't real, its just a tiny bit closer. The 5 secs allows for 'oh shit' time, those times you open the door and depressurise the room without your helmet, or take your helmet off accidentally as @zrisher said.
It also allows modders to make resilient characters, perhaps an alien species that can survive minutes with no oxygen/pressure.

mexmer:
@sibz if you ever were doing diving, you should know, that with low pressure in oxy bottle, you have less income of oxygen, so you will feel haze and have other problems, before you run out of oxygen. it's quite unrealistic to have run oxygen tank dry without being already affected ... so better leave it as is, your attempt for realism has nothing with realism. Running oxy bottles dry works in sci-fi movies, but not in real pressure suit.

Sibz:
@mexmer This isn't a debate of realism, its a feature of not being affected by taking damage straight away.
I understand there is problems before your 'tank' runs out. But this is more about the instantaneous oxygen loss when opening door with helmet off, or taking your helmet off in space, or exiting a small ship with no helmet. The code doesn't differentiate between running low in oxygen suit and just hitting 0 oxygen. It does however do less damage from 0-50% pressure when running out of oxygen in a pressurised room.

I could add in low damage when suit oxygen gets low, but that would be a separate issue.

mexmer:
@Sibz well, you started with realism. if you stated, it's just "workaround" over door 0 oxygen bug ....

anyway, i think points are clear here ....

forget about all realism discussing ... just workaround ... tho' i think 2 sec should be enough for that.

Sibz:
@mexmer the original intention was to allow a few seconds to do something about sudden decompression, i.e. put helmet on, close the door, re-weld the block that you just ground down. Though I realised it would also circumvent the issue with opening doors inside.

But having discussed this, and thought about it a while, there is already a mechanism for this. The damage taken isn't fatal, so you do have time to do something before you die.
Therefore, it does become solely a work-around for the bug, not what I had in mind.
Modders can make resilient creatures by adding more health and lowering the damage from oxygen (DamageAmountAtZeroPressure).

I'll think I'll close this now.

mexmer:
well, if you can find source of decompression problem, would be nice ....

Sibz:
@mexmer haha, I've looked at the code for grid oxygen system - woah its a beast. but I'll give it a go :)

PR #258, opened 2015-06-07
Dx11 multidisplay check fix

mexmer:
fixes issue with game not displaying multiple displays connected to single adapter.

in rare cases (fullscreen only), mouse doesn't switch to new display - alt-tab fixes issue .... will post fix, when i find source of problem


planetenkiller:
I think the implementation is still broken because the 'adapter.Outputs` loop changes the `info` object created a few lines before the loop (`adaptersList` contains same `info`  multiple times).

mexmer:
@planetenkiller info reuse is correct, adapter remains same, output refers to monitor.

output loop adds same adapter to list but with different output, and different display modes
only connected monitors generate Output entry, so there is no problem with it.

loop preserves info, replaces output and display modes and stores it to adapter list.

also this way it was in version 1.0.84, where multimonitor was working - you can check commits to file to see, where it was changed.

main problem is with switchable graphic on notebook.
you have 2 graphic cards (2 adapters), but only one of them has output (notebook lcd), other one has no output. and therefore no display modes.
that was supposed to be fixed by fallback modes, but implementation of that case was incorrect also.

i have tested this changed code on my desktop with R7-260x and 2 monitors connected, also on notebook with GT750M + Intel HD4600 ... and found no issues. except that mouse problem, which doesn't appear always, but sometimes it does.

planetenkiller:
Thanks for the details. I found my mistake: I thought MyAdapterInfo is a class (it's a struct)

Phoenix84:
I saw a bug report on the forums and wondered if this would fix that.
http://forums.keenswh.com/threads/1-085-changing-display-to-secondary-display-in-dx9-than-changing-to-dx11-breaks-game.7361670/

Worst case, it's another test to try. :-)

mexmer:
yes, it should fix that. there is problem with display index, if you enable game on secondary display, your adapter + display combination get index 2, but due change in code, you have only 1 adapter + display in list - that result in out of index exception.

i have been notified tho' that issue was already solved in KSW code ... that's also reason, why this still was not merged. waiting for github update, and then we will see.

Phoenix84:
Nice. I do have multiple monitors, and while the game doesn't show both (however ME does), I haven't run into any issues myself.
So at least I'm a test control. :smile:

mexmer:
new update in master fixes issues with multimonitor, therefore close

PR #260, opened 2015-06-07
Remove '.' from faction name, and '-' from antenna if antenna name is blank

Phoenix84:
This cleans up the antenna HUD output if the antenna name is blank.

If ship name broadcasting is turned off, it will just show the faction name (eg. ~~PHX~~), not faction name and period (eg. ~~PHX.~~).

If ship name broadcasting is turned on, the '-' will only be shown if the antenna name is not (eg. ~~Ship name - Antenna~~). If blank, it will only show ship name (eg. ~~Ship name~~).


lord-devious:
:+1:

RossM:
LGTM

PR #261, opened 2015-06-08
Add Color name search

0xNeffarion:
Adds a function to retrieve a Color by name (just like System.Drawing.Color).


No-Brain:
This would be very slow

0xNeffarion:
See if its appropriate with new commit.

0xNeffarion:
i think that's it. can't think of any other way to get faster

PR #262, opened 2015-06-08
Character interface expansion

THDigi:
Exposed lots of useful fields to `IMyCharacter` interface, mostly read-only.

Also:
- Added setters for health, oxygen, battery, model&color, toggling jetpack, lights, broadcasting and dampeners which have the option to do it without notifications;
- Added a `Kill()` method which kills the character or can optionally ask to suicide;
- Added `IMySuitBattery` interface to allow more information to be gathered from the suit battery;
- Added `SwitchBroadcasting()` to `IMyControllableEntity`;
- Moved the `EnvironmentOxygenLevel` field to the `MyCharacter_ModAPI.cs` file to keep things consistent.


Spartan322:
The kill method should probably supply a damage type parameter, allowing the default to be DamageType.Suicide.

THDigi:
Good idea, but while I was doing that I also made Kill and DoDamage able to ignore the CanDie field by default with the option to obey it if anyone desires.

JimLess:
Digi can u merge this "Extending interfaces for Role Play mod. #269" or replace by you own strong code? I need some functionality to be able at client side. I just tried the to add the code, and you have most of the ready.

JimLess:
I've been using this interface in my assembly. until everything works fine! I hope very soon the request will merging the official branch.Really I am looking forward to this!

JimLess:
this interface must be merged as soon as posible.... please...

THDigi:
Hmm, fixed merge conflicts :}

IhanaMies:
What error do you get? If it stays on the small loading screen, just hover your mouse over one of the icons on toolbar and there is a small window saying "Your steam is offline" or something similar. Close it and the game should continue to load normally. Although you cant use workshop or blueprints. Dunno about mods

THDigi:
Ah, nevermind, something was wrong that caused errors on various files, cleaned and rebuilt and it's all ok.

Not sure about the new health system, if the current method is fine.

JimLess:
THDigi, i can't die in survival mode. What happens? I download SE repos and merge your changes. Please check, what cab be wrong.
Mode: Survival, Autoheal options is off

JimLess:
With this update i can die anymore or suicide. Digi can u approve this?
dedicated + survival

THDigi:
@JimLess I don't understand what you mean, can you rephrase that ? :P

JimLess:
Ok. I send u a message, that player can't die in survival game. I'm trying suicide my self... Not works any more. Maybe it because i merge your pull request in last source update... So it can be possible that your code is breaking game. Please, check it. Create a dedicate server survival mode and try to die... I hop i'm wrong, but i change only one thing from last update... Now u understand?

THDigi:
Ah I see, it seems to happen even on the master branch so it's not because of this PR.

I did some digging and I found the issue, damage is ignored if health is null and health is null when character is at full health. I'll submit a bug report :P

JimLess:
THDigi, thank u. I have no time to investigate such situations...

JimLess:
+1

THDigi:
Removed `DoDamage()` and `Kill()` methods because similar things already exist now in the IMyCharacter and IMyDestroyableObject.

paradineai:
This sounds close to a way of adding in a power pack that can be recharged in a reactor on a ship like a battery then moved to you suit inventory like the o2 bottle so it can recharge you suit power just like the o2 bottle recharges your suit o2. I may be wrong about that though. If I am correct then this could turn out to be a very useful thing for everyone and not just server admins.

PR #267, opened 2015-06-08
ShipWelder Performance

yajiedesign:
ShipWelder Performance .
cache MaxDeformation,calculation each farme.
cache FindProjectedBlocks. calculation  each 10 farme.


Sources/Sandbox.Game/Game/Entities/Cube/MySlimBlock.cs
```diff
@@ -193,12 +193,19 @@ public bool HasDeformation
                 return CubeGrid.Skeleton.IsDeformed(Position, 0.0f, CubeGrid, true);
             }
         }
-
+        private int m_lastTotalFrames = 0;
```
RossM: Adding fields to MySlimBlock makes me nervous - is there any way you could do this that wouldn't increase memory consumption?

yajiedesign:
@RossM put it in LRUCache ?

Sources/Sandbox.Game/Game/Weapons/Guns/MyShipWelder.cs
```diff
@@ -71,6 +72,7 @@ public override void Init(MyObjectBuilder_CubeBlock objectBuilder, MyCubeGrid cu
             base.Init(objectBuilder, cubeGrid);
 
             m_missingComponents = new Dictionary<string, int>();
+            this.NeedsUpdate |= VRage.ModAPI.MyEntityUpdateEnum.EACH_10TH_FRAME;
```
No-Brain: This would murder performance right there.

BasilHorizon:
We need it very much, please add this.
Btw how much it affects perfomance?

alexflorea:
Hi. Thank you for pointing out these issues. While we won't be using the PR directly, your suggestions helped us find and fix some of the performance problems with ship welders. The fixes should go out soon.

PR #269, opened 2015-06-09
Extending interfaces for Role Play mod.

JimLess:
+ ChangeModelAndColor(model, colorMaskHSV); - for changing spaceman
model by client side script.


JimLess:
I plan to expand the interface. Any objections?
situation: the player buys different types suit for game currency. (implemented)

mexmer:
when adding interfaces or fields or methods to interface, add also ///<summary> , so it will be reflected in documentation.

and yes, i'm aware that many existing interfaces doesn't have it, but it's common sense to add it anyway.

///<summary> is processed by sandcastle, which is used to create api documentation file

https://msdn.microsoft.com/en-us/library/2d6dt3kf.aspx

JimLess:
Ok. I do this soon.

mexmer:
thanks

JimLess:
Please check this out.

PR #272, opened 2015-06-10
Hostile Encounter Ships

CodeHamster:
Objective:

Exploration encounters currently leave a lot to be desired for the more adventurous space engineers.

This is the first in a line of updates aimed at improving the situation.

What this update provides:

1) A new check-box on the advanced world settings screen called 'Hostile Encounters'.

When checked.... 

2) all exploration encounter ships will spawn with their reactors ON.

3) all Gatling, missile and interior turrets will be turned ON.

4) A single antenna will be turned on and renamed to "Hostile"

5) All other antenna will be turned off.

6) The range of the 'Hostile' antenna will be set to broadcast it's position over 50Km for large grids and 5Km for small grids.

7) Each group of spawn ships will be assigned its own unique hostile faction ID, which all ships in the group share.  This means that they will appear hostile to players and will open fire on them when they get close.  It also means that players will not be able to simply walk into a derelict ship and take it.

What this update will NOT (or at least should not) do.

1) Change the way that the standard encounter system works.

2) Inventory management of the hostile ships ammo and fuel supplies.

3) Modify any other ship or block settings.

4) Change the spawn density, number or make-up of the ship spawn groups.

Providing this change is accepted, I intend, at some point soon, to add better spawn ship inventory management and balancing and will try to improve the way in which ships spawn groups are selected, so that they complement one another instead of being random.  

I may also add an additional 'Encounters' button to the advanced world settings screen which will open a new GUI screen to allow encounters (both hostile and non-hostile) to be configured globally by the player.

However; that is for the future and since I appreciate that Keen may well have their own plans for this area of the game, I would like to see how this change goes, before embarking on further and extended alterations.


kleshchevand:
>A single antenna will be turned on and renamed to "Hostile"
This will just make it easier to find ships. Small ships can't fight back and will be practically welcoming players to board them. Large ships might be a bit chalenging due to unknown design, but nothing really interesting since it is easy to find dead zones on unmoving ship (or create).
Unfortunately you need combat autopilot to make real 'Hostile Encounter'.

Tyrsis:
This should be a mod honestly.  Source changes should be made to help facilitate this as a mod, but shouldn't implement something that should not be part of the base game.

CodeHamster:
Yes, I agree, and I considered whether to do that or not when making the change quite carefully.  

Initially it was added for testing, however I left it in for the following reasons:

1) A lot of the small ships either don't have any antenna or the fuel to power them, so won't be easy to see anyway.  The large ships usually have defences of some kind.
2) The max range of small ship antennas is 5Km, so you still have to be pretty close.
2) It's too hard for new players to actually find the encounter ships (particularly the smaller ones).
3) It's so easy for experienced players to find them (using either camera zoom, scanner mods or SEToolbox) that intentionally hiding them is a bit pointless.
4) It brings encounter ships more on a par with cargo ships.
5) I don't feel that the game benefits greatly from hiding the content that people expressly wish to find, just for the sake of it.  It's not hide and seek.  However I do agree that because they are easy to find, steps should be taken to make them more dangerous and less game-breakingly valuable and that's what I'm working on now.

I also agree that encounter ships would be better with a dedicated combat AI and perhaps Keen, one of the other contributors or I will add that in at some point, however at the moment I can't contribute enough time to do that properly and so I am doing what I can to improve the situation in the mean time.

I have however been working on a new and separate GUI settings screen, dedicated to encounter configuration, with a number of options including the ability to specify a % chance of the ships being hostile (rather than the awkward on/off button included here), how many ships should appear in a group and whether they are perfect or damaged at spawn, amoung other things.

I will add an extra On/Off button to that screen, too allow the antenna to be turned off universally at spawn for those who prefer that.  That change should be complete within the next few days, depending on how much time I have available.

CodeHamster:
Tyrsis;
Perhaps, however I've seen no long term plan of what the base game will contain in the future or what Keen intends it to become and so I can only provide an option that I would enjoy using and hope for the best that Keen agree.  

Would I use the above as a mod.  No, I only play vanilla  So I didn't add the functionality to do that.

kleshchevand:
You can provide tools that will allow creation of active exploration vessels instead - expose some related logic, add couple new functions to in-game scripts and players will do lots of aggressive/challenging ships over time may be even some mini-missions and mini-games.
For example some kind of GetGeneratedShipInstance() that returns null/exploration/cargo classes and said classes have access to normally unavailable data like nearest player coordinates, some short-distance (antenna-distance) chat functions and everything that is necessary for custom autopilot to work and target ships/players.  Functions like getGeneratedParent() – was constructed/detached from exploration/cargo parent. 
In my opinion difference between exploration and cargo classes is that cargo might have destination/route and/or despawn timer, isAllowedToDespawn(), Despawn() functions. Exploration class might have access to Respawn() function that will either move/teleport+stop ship at spawn coordinates (may be restore cargo, but not repairs), isAllowedToRespawn(), GetSpawnCoordinates() functions. Exploration stations might even have some kind of ability to spawn smaller ships using projectors or assign, compile and run scripts at welded small ships. 

P.S. You can expose generation of 'exploration' vessels and 'cargo ships' to scripts so that modders would be able to add mods that can react to appearance of such ships and can control their behavior.

CodeHamster:
Indeed and they're all potentially good ideas, however without this change or something like it, your spawned ships will just fly straight by you without opening fire because they're the same faction as you or at least neutral.  

As mentioned in the title, this is a first step towards something more interesting.  There are things that I can and would like to see improved with it, but Keen did say that they want one feature at a time and so I would like to focus on this one enabling step to begin with.  

I will probably have to break the one feature rule, when I submit the GUI, because having a new configuration screen isn't worth adding when you have only one setting on there, but I will otherwise try to stick to the rules.

At the end of the day, if they don't like this, then they probably won't like any of the other changes that I have in mind either, so it's worth testing the waters and seeing whether it's acceptable before I continue.

kleshchevand:
-ships will just fly straight by you without opening fire 
Then why not let assigning of new owners to exploration ship from scripts? The only problem I see is that you will need script recompiled for new owner, but that can be part of 'reassign exploration ship' API.

CodeHamster:
Are you talking about in-game scripts or the modding interface?

CodeHamster:
Reading your earlier comments, I think it's the in-game scripts you're refrring too.  Spawning ships and changing factions from in-game scripts is against everything that I've seen Keen do with that over the last few months and I don't think they'd allow it.

kleshchevand:
Since in-game script will have to decide whether exploration ship is hostile or friendly to player it will also need functuinality to change owner without need to 'recompile' or with auto-recompile (only allowed for exploraton/cargo ships). This shouldn't be a mod, this should be a possibility for players to pre-program explorations/cargo ships to do ceratin actions normal in-game scripts aboard normal ships are not allowed to do.

CodeHamster:
OK, well I won't intentionally implement alterations that open up the possibility of exploits in multi-player.  This change, as it is, is something that players or the server admin chooses at the beginning and will remain that way.

However, you have a clear idea about what you'd like to see and so I suggest you get a copy of the code, make your changes and submit them.  At the end of the day that's what I did.

Tyrsis:
All I'm saying is that everything you've implemented can be done in a mod.  Adding it to source because "I only play vanilla" is not a real response. 

I'm only saying this, because the chances this gets merged is very very low due to the fact that this not only should be a mod, but is actually almost exactly like mods that already exist.

mccorkle:
A similar "radar" block was added to (the now deprecated) Sector 15 Mod,
but it did require that we first touch every enemy ship class, add in a
beacon, and put in enough Ur to power the beacon over a reasonable lifespan
of the enemy ship.

So yes, this can (and has been) done in mods, but it really added a HUGE
playability factor to the game and I encourage it to be added into the base
game to encourage more PVE in multiplayer and scenarios.

CodeHamster:
Tyrsis; once again, maybe you're right, I've not looked for similar mods, and won't, otherwise I could potentially be accused of stealing their ideas and presenting them as my own.   I note however that the GPS/way-points system was generally regarded as a mod that Keen then incorporated in-game, because it made a lot of sense.  

Too me, having the option of encounter ships in Vanilla that shoot at you when you try to steal them, makes a lot of sense.  Unless I've missed a post somewhere, Keen haven't said 'you will not develop such and such', at least when it comes to game-play options.  They pretty much said 'go and play with our code and if we like what you do we'll add it'.

While 'I only play Vanilla' may not be a good response, it is a real one and a true one.  I've spent a year playing vanilla, but know nothing about using or making mods.  It never appealed to me.  What right do I have to try to modify code that affects an area of the game that I don't use?  I can only add game-play options for which I have a fairly good understanding of the consequences, the same as everyone else.

MeusRex:
I like.

I think the encounters should have been hostile from the very beginning. Also, I'd say keen should have created a script to auto balance all the encounters' inventories...

CodeHamster:
Keen: if you are considering using this code, then you might want to wait.  I have more or less finished the new GUI configuration screen, which will include the above functionality in a more flexible and clear way, essentially making this obsolete.  I'm still finishing up and there's some testing to do, but it should be released on a different branch this evening.

CodeHamster:
This has now been made obsolete by the following change https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/288

Closing the pull request

PR #273, opened 2015-06-10
Add Inventory Mass

IhanaMies:
Add real time mass recalculation to ships when Inventory Mass is enabled from Advanced World Settings.

Only thing worrying me about this PR is how big impact it has on multiplayer. Since ships inventory masses are constantly being updated by refineries, assemblers and drills when mining, server has to sync the data between players creating somewhat unnecessary data. By how much? I have no clue and hope it is not a barrier.


mexmer:
if you are woried about impact, you might change your code, so value getter retrieves private variable, that is updated every 10th or 100th frame, eg. inside Update10 or Update100 function.

imo even update100 will still be enough, that is update like every 1.8 sec, but you might test it with Update10 and see, if it doesn't create too much overhead ...

Phoenix84:
Yeah, the inventory data will already be synced, so you can just calculate the mass locally. No need to send and receive network data for it.

OndrejPetrzilka:
We have some code for this feature already in Medieval Engineers, I'll check if we can use it first, because it's already working and tested. Thanks!

IhanaMies:
Cleaned up the code after testing it further. It now updates ships mass correctly. Before it did not include inventories mass before the ship was turned on.
Also ships center of mass is updated whenever mass is changed, although center of mass crosshair is not drawn correctly in DX 11. It is blocked by any object. This does not happen in DX 9.

Currently this doesn't take into account the mass of player and his inventory since I did not find a way to get it and where it should be pointed.

IhanaMies:
Inhouse progress :+1:

PR #276, opened 2015-06-11
fix for exploding warhead projections

mexmer:
disabled countdown and armed flag on warhead, when loaded in projector. (prevents exploding blueprints)
note - this is only for projector, to prevent "free" warhead explosions from projections.

if you save bp with wardhead countdown set, and then load it and paster into world, it will still explode.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyProjector.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyProjector.cs
@@ -735,6 +735,7 @@ namespace Sandbox.Game.Entities.Blocks
 
             foreach (var projectedBlock in ProjectedGrid.CubeBlocks)
             {
+                projectedBlock.UpdateForProjection();
                 Vector3 worldPosition = ProjectedGrid.GridIntegerToWorld(projectedBlock.Position);
                 Vector3I realPosition = CubeGrid.WorldToGridInteger(worldPosition);
                 var realBlock = CubeGrid.GetCubeBlock(realPosition);
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyWarhead.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyWarhead.cs
@@ -148,9 +148,7 @@ namespace Sandbox.Game.Entities.Cube
 
             m_countdownMs = ob.CountdownMs;
             m_isArmed = ob.IsArmed;
-
-            if (ob.IsCountingDown)
-                StartCountdown();
+            IsCountingDown = ob.IsCountingDown;
 
             this.IsWorkingChanged += MyWarhead_IsWorkingChanged;
         }
@@ -183,9 +181,9 @@ namespace Sandbox.Game.Entities.Cube
             {
                 StopCountdown();
             }
-            // restore countdown on finished block
-            if (IsWorking && IsCountingDown && !IsProjection)
+            else if (IsCountingDown && IsWorking && !IsProjection)
             {
+                IsCountingDown = false;
                 StartCountdown();
             }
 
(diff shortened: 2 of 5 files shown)


EnochMiningCorporation:
I strongly object to the inclusion of this feature, as I believe it breaks the existing, beneficial functionality:

_Warhead built from projector respects it's corresponding state set in the blueprint._

Without a means of blocking strangers from copying blueprints, the current functionality can be used to provide some forms of mitigation from 'blueprint theft' in multiplayer worlds.

Sibz:
@EnochMiningCorporation I think the current set-up doesn't do much to prevent blueprint 'theft', if the 'theif' got your blueprint, and tried to build it, found it had an exploding warhead, they could use the blueprint in a world with 'weapon damage' turned off, or edit the file manually.
The lack of a way of preventing strangers from copying blueprints is probably a separate issue.

@mexmer A question to answer here is, does the warhead retain its countdown once built? This is a useful feature for people making automatic missile builders/launchers.

mexmer:
@Sibz no it does not. It retains timer, but armed flag, and isticking (eg.countdown active), is cleared.  I will test again, if it's needed to clear IsArmed flag, might be, it is not required, so only one, that will be cleared is "isticking" eg. countdown active.

Function SetupForProjector (which is available on any objectbuilder, but so far used only for cleaning inventories, ammo and stuff from refineries and assemblers), is aplied when blueprint is loaded into projector.

I plan to do similar function, complementary to SetupForProjector(), that will restore states after block is finished, but first need to think out method, how to save the fields, that are cleared upon SetupForProjector().

If you created "torpedos" on merge block tho' you can set them trough pb, or manualy after they are finished.

@EnochMiningCorporation honestly, your argument is moot. Removing warhead from BP is less than 1 minute work. If someone has enough skill to steal BP, he has enough skill to search for potentials problems in BP. Also why should anyone steal your BP, if either he can download it from workshop (if it's public), and doesn't if contain exploding warheads will get lot of downvotes. Or make own BP from ship, just when he sees it ingame. This patch doesn't affect paste, only projection.

This patch is to prevent making bombs out of nothing. ATM. you can just sit 2km from enemy base, setup project, project 2km long block of AB, which has bunch of ticking warheads at end. you have almost zero cost bomb, not mentioning at "safe" distance. There were multiple examples of projector bomb on forum already. Function SetupForProjector is relatively new, and it was created for preventing transfering resources and components "for free" trough bp.

mexmer:
updated version.
i removed clearing countdown on blueprint load ... that had some other adverse effects, and most of all, it was not syncable state.

new implementation stops counter upon displaying projection ... this allow block to retain it's state, so when it's finished, countdown will automacally start.

Sibz:
:+1:  Excellent.
I see IsProjection isn't set back to false anywhere, is that because the actual welded block is created new?

mexmer:
Yes, welded block is actually cloned from projection and placed into world. Therefore it's not needed to set it back.
It might be needed later ... but not for now, it's not.

mexmer:
rebased onto current master

OndrejPetrzilka:
I'm not sure this is preferred approach, I don't like adding Fields into base classes because of new functionality, maybe there's better way or maybe not. I leave this to Alex (projector author).

mexmer:
thanks for response.
i was thinking on other way, but that is involving changes in prefab which on one hand will fix problem with exploding "projection" from CTRL+V (eg. my current fix only affects projection from projector, not "preview") but 
1) in survival you don't have CTRL+V
2) changing prefabs is even deeper change than this one.

i did change in base class, in case, it's needed for fixing something else, when projecting all blocks execute UpdateForProjection, so if there is any special condition, that needs to be set upon emiting projecting, block can just override this method.

alexflorea:
This issue was fixed internally in the warhead itself. Thank you for pointing it out and helping with the fix.

mexmer:
thanks for feedback ... closing then.

PR #277, opened 2015-06-12
Suspension additions fix and extra interface

THDigi:
Fixed the issue described here: https://github.com/KeenSoftwareHouse/SpaceEngineers/commit/b0f978dfe100bdf220b4031a17ab20f19001c23d#commitcomment-11657905

Also added ModAPI.IMyMotorSuspension interface mirror for modders to avoid the need to use `using Sandbox.ModAPI.Ingame` which causes conflicts with `using Sandbox.ModAPI`.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMyMotorSuspension.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyMotorSuspension.cs
@@ -1,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Sandbox.ModAPI
+{
+    public interface IMyMotorSuspension : Sandbox.ModAPI.Ingame.IMyMotorSuspension
+    {
+    }
+}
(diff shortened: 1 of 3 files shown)


plaYer2k:
Would you like to add one small addition too, a reverse acceleration checkbox like the reverse steer?
That would be very helpful for some projects and builds like a flipping car :-)

Other possible additions (which i only did locally myself without a pull request) is setting acceleration and targeted steer angle (the center to which the steering automatically returns, which is by default 0) so you can control suspension wheels through rotors and pistons with ingame programming.

Thanks for the work and good job there.

THDigi:
Well since you've already coded them, you should submit them :P

plaYer2k:
If i were home and had access to the code, then i would.
But thanks to me not having access to my desktop that can run SE aswell as the changes, that seems unlikely.

I might be back home in about a week or more, then i could submit a pull request after some more testing.

mexmer:
not sure about other thing, but reverse mode sounds like good adition.

PR #279, opened 2015-06-13
Add 'Audible Proximity Alert' option to Sensor

CodeHamster:
Added an option to allow the player to turn on/off the annoying click sound that plays when
something activates a sensor that has actions defined.  By default the
proximity sound will be on, as it may help the player with testing the
sensor.

The option is accessible from the settings of the sensor and will appear above the 'Detect Player' toggle.


mexmer:
detectPlayProximitySoundSwitch.EnableOnOffActions is not required.

keep in mind, that 
IMyTerminalControl.EnableAction
IMyTerminalControl.EnableActions
IMyTerminalControl.EnableToggleAction
IMyTerminalControl.EnableOnOffActions

add also actions for toolbox, so you effectively increased toolbox actions for sensor by 3 ... while toggle is more than enough. Or you can't opt to not use EnableToggleAction at all.

CodeHamster:
Thanks Mexmer, That was a good point.  Being able to set that setting remotely isn't really very useful at all and so I've removed both the toggle and the On/Off.  Thanks for looking.

mexmer:
which reminds me, i was thinking about modifying terminalcontrol, so you can effectively set, where the actions should appear. but i have so much things on list and don't know which one to do first.

functions i mentioned above add actions also for PB, and it might be usefull to have some action available for PB, while not have them for toolbox (to reduce toolbox options), or vice versa....

PR #280, opened 2015-06-13
Auto Pilot Fix

FlySniper:
Fixed an issue where the auto pilot would go to the first waypoint at
the wrong angle


alexflorea:
This issue was fixed internally. Thank you for pointing it out and helping with the fix.

PR #282, opened 2015-06-14
MyAdvancedDoor - Additions and Fixes

Eikester:
Added "Autoclose" Checkbox and "Autoclose Interval" Slider to Terminal (as per Request)
Fixed OpenRatio returns value > 1
Fixed Bug where Subparts will be added again when OnModelChange is called (i.e. when coloring the Block)


mexmer:
why you delete repo instead of rebase?

Eikester:
because i dont get it, i tried every guide i found but none worked, was easier this way

PR #284, opened 2015-06-15
Allow a powered connector to lock to an unpowered one

Knsgf:
This simple tweak allows a powered connector to establish a lock with an unpowered
one, as long as both connectors are switched on in the control panel.

Note that there several differences between this and initial (buggy) connector lock implementation:
1. The magnets disengage as soon as either connector is turned off. In contrast, the original implementation required BOTH connectors to be disabled in order to turn off magnetic attraction,
2. Only a powered connector may initiate a lock,
3. Finally, if both connectors lose power, the lock is immediately broken.

For detailed explanation on why this behaviour might be useful, see this thread on KSH forums:
http://forums.keenswh.com/threads/connect-to-unpowered-connectors-read-1st-post-before-voting.7361664/

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Cube/MyShipConnector.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Cube/MyShipConnector.cs
@@ -561,8 +561,9 @@ namespace Sandbox.Game.Entities.Cube
                             var otherPos = otherConnector.ConstraintPositionWorld();
                             float len = (otherPos - pos).LengthSquared();
 
-                            if (otherConnector.m_connectorMode == Mode.Connector && otherConnector.IsFunctional && (otherPos - pos).LengthSquared() < 0.35f)
+                            if (otherConnector.m_connectorMode == Mode.Connector && otherConnector.IsFunctional && otherConnector.Enabled && (otherPos - pos).LengthSquared() < 0.35f)
                             {
+                                MyHud.Notifications.Add(new MyHudNotificationDebug("Attach: " + CubeGrid.DisplayName));
                                 CreateConstraint(otherConnector);
                             }
                         }
@@ -580,8 +581,9 @@ namespace Sandbox.Game.Entities.Cube
             }
             else if (Sync.IsServer && !IsWorking)
             {
-                if (InConstraint && !Connected)
+                if (InConstraint && !Connected && (!IsFunctional || !Enabled || Enabled && !m_other.IsWorking))
                 {
+                    MyHud.Notifications.Add(new MyHudNotificationDebug("Detach: " + CubeGrid.DisplayName));
                     Detach();
                 }
             }
@@ -694,7 +696,7 @@ namespace Sandbox.Game.Entities.Cube
                 var connector = block.FatBlock as MyShipConnector;
                 if (connector.InConstraint) continue;
                 if (connector == thisConnector) continue;
-                if (!connector.IsWorking) continue;
+                if (!connector.IsFunctional || !connector.Enabled) continue;
                 if (!connector.FriendlyWithBlock(thisConnector)) continue;
 
                 m_tmpBlockSet.Clear();


lord-devious:
In other words: You can lock onto a connector which is powered down, but there will be no magnetic attraction in this case?

Knsgf:
Magnetic attraction will work normally. However, only a powered connector will be able to initiate a lock in this case; an unpowered connector's "Lock" button will remain greyed out.

Geneticus:
This might be against the intended function of connectors. It might be better to make this work as is by default, but allow this to be modded on a custom Connector.

Knsgf:
If you check solar panel array on Easy Start 1 platform, you'll find a couple of connectors near the top, built with sole purpose of transferring electricity, so I don't think that this ability is against intended purpose of connectors. At any rate, the only other ways to connect a powered grid to an unpowered one are either merge blocks -- which do not work for grids of different sizes and are more intended for permanent or semi-permanent connections -- or using rotor's attach/detach function, which isn't the primary intended purpose of rotors either.

lord-devious:
My opinion is that this is a base requirement, something that _should_ be vanilla. Something that can help you get rid of the ridiculous "startup reactor" for ships run purely by battery.

mexmer:
i believe there was discussion on forums regarding this.
more people were pro, than against, but if one connector is not on ... i'm not sure about attraction.
also i don't think, there was response to that thread from KSW staff

Geneticus:
Going from history... When Batteries were implemented in 01.039 we were able to connect to unpowered grids to charge batteries, but when they changed it, Phand said it was because connecting to unpowered grids was not intended:
(http://oldforum.keenswh.com/post/01-046-connectors-no-longer-connect-if-one-is-unpowered-7066848?highlight=connector). 
The fact that that "fix" negated  the examples in the Battery Patch Video (https://youtu.be/LmkMqpOEajw?t=68) was never addressed by Keen other than to tell us to use Merge blocks. I agree it should be doable in vanilla, but it would be more likely to get it through as an extension to the Modability of connectors rather than a reversion of a change Keen intentionally made.

Knsgf:
If we consider the official response in the bug report thread:

"Hi,

it was actually a bug. Now it is fixed. **It should work same as merge blocks. They do not connect if one is OFF.** Thank you for report though.",

we can clearly see that connectors were supposed to operate in a similar manner to merge blocks. Which, among other things, includes the ability to connect to an unpowered grid, as long as both connectors' control panel switches are in "On" position. This is basically what this PR implements, rather than going back to initial buggy behaviour.

P. S. Adding support of extra tag in CubeBlocks.sbc to disable this ability should be easy enough, but I have a hard time imagining a scenario where not being able to connect to a electrically dead ship or station would be useful.

OndrejPetrzilka:
Can you please make short summary what is changed?
Ideally:
original state is XXXX
new state is YYYY

plaYer2k:
I really like this fix too, at least from the textual description.

Having to power any construction just to engage a connection through connectors felt cumbersome for setups like modular cargo boxes which you want to drop solely through connectors and without having to fiddle with connector and merge block setups.

Thus this got my support from a design and gameplay point of view.

Knsgf:
@OndrejPetrzilka,

The change between current and proposed behaviour is highlighted in bold [point b)].

a) Both connectors are functional, switched "ON" and powered. In this case the magnetic attraction will work normally and the lock can be established from either side. This behaviour is the same for both current and proposed implementation,

***b) Both connectors are functional and switched "ON", but only one of them is powered. The magnetic attraction will still work normally, but only a live connector will be able to initiate a lock.
Under current system, no magnetic attraction occurs, and no connection can be made,***

c) Both connectors are functional and switched "ON", but both connectors are dead. There will be no magnetic attraction in this case nor any connection would be possible. Both current and new system behave the same way in this case,

d) One of the connectors is broken or switched "OFF". No magnetic attraction, not possible to lock regardless of power. Again, no difference between current and proposed solution,

e) Both connectors are broken or switched "OFF". Same as above.

MrFaul:
I think Ondrej meant something more like this:

### Current:
 **All blocks have to be functional!**

*Lockability* |  On  Power | !On  Power |  On !Power | !On !Power 
-------------- | -------------- | -------------- | -------------- | -------------- 
**On  Power**| **1** | 0 | 0 | 0
**!On  Power**| 0 | 0 | 0 | 0
**On !Power**| 0 | 0 | 0 | 0
**!On !Power**| 0 | 0 | 0 | 0

### New:
_**Note: Lock can only be initialized from a powered one**_

*Lockability* |  On  Power | !On  Power |  On !Power | !On !Power 
-------------- | -------------- | -------------- | -------------- | -------------- 
**On  Power**| **1** | 0 | **1** | 0
**!On  Power**| 0 | 0 | 0 | 0
**On !Power**| **1** | 0 | 0 | 0
**!On !Power**| 0 | 0 | 0 | 0

FavioGalvis:
Supporting too!

harag-on-steam:
:+1: 
 * quality-of-life improvement: allows to connect grids that don't have their own power source
 * no loss of functionality: turning a connector off still allows to deny locking
 * consistency: merge-blocks work the same

PR #287, opened 2015-06-16
Advanced Door Fixes and Changes

Eikester:
Added "Autoclose" Checkbox and "Autoclose Interval" Slider to Terminal
(as per Request)
Fixed OpenRatio returns value > 1
Fixed Bug where Subparts will be added again when OnModelChange is
called (i.e. when coloring the Block)


OndrejPetrzilka:
If you want to make sure auto-merges are fine, move added lines to xxxTexts file up several lines.
When your texts are added to end and we add some texts to end, git does not know which to include first. If you add it few lines above and not to the end, it should merge fine.

Eikester:
Reopened here #290 Any chance this could be added for tomorrows Hotfix since the code is now in the Game? Or should i issue a new PR with just the Fixes?

PR #288, opened 2015-06-17
FEATURE: Encounter configuration and selection GUI.

CodeHamster:
The new 'Encounter Configuration' GUI.
encounterconfiguration medium

The new 'Encounter Selection' GUI
encountershipselection medium

An example of ship damage applied to freshly generated encounter ships.
damage 1 medium

How the interior of a station looks with the damage applied.
damage 2 medium

A close-up of exterior damage to a ship.
damage 3 medium

<strong>Purpose:</strong>
Allows encounters to be fine tuned across your world, to suit your preferred play style.

<strong>Features:</strong>
Adds a new 'Encounter Configuration' GUI screen, accessible from the advanced world settings screen using the encounters 'Configure' button.

<strong>New!</strong>  Adds an 'Encounter Selection' screen which gives you total control over the ships/bases available for encounters. It is accessible from the 'Select Encounters' button at the bottom of the configuration screen. 

*  <strong>You can set the number of ships in each encounter group up to a maximum of 5.</strong>  The actual number of ships will be randomly determined as the group spawns, with the selected max as the cap, so even if you select 5 you'll probably get an even mix of small and large groups.

* <strong>You can set the % of hostile/non-hostile encounters in the game.</strong>  Hostile ship groups will shoot at you when you approach.  They will have all of their turrets enabled. Different groups will also shoot at other groups and cargo ships, but not at other ships in the same group.  Since they're static that isn't much of a feature. Yet.

* <strong>You can set the % of damaged/undamaged ships that spawn in the game. </strong> Damaged ships will look battered and beaten.  I've tried to make this look as natural and interesting as possible.

* <strong>For damaged ships you can choose the maximum damage level that will be applied.</strong>  The actual damage applied will be chosen randomly, using this setting as the cap.  So there will be ships ranging from brand new to what-ever your prefered level of damage is.

* <strong>You can also apply the same level of damage to all ships equally.</strong>  So if you want all of your encounter ships to look like they've been eaten by nano-bots, without any shiny, new ones in the game, then you can.

* <strong>You can choose the % of ships that begin with their reactors on.</strong> It's worth keeping in mind that the ship may be hostile, even with the reactors off.  Especially when you hit the on button.

* <strong>You can choose the % of ships that begin with their antennas active.</strong> For those ships that spawn without antennas active, all beacons and antenna will be turned off, to keep the element of surprise if the reactors are enabled.

* <strong>You can choose whether antenna use their maximum range or the default for the ship.</strong>  Some of the exploration prefabs have these set to lower than max limits, presumably to conserve power, ticking this forces their antenna to broadcast their position over a wider area, at the expense of their fuel stock-pile.

* <strong>You can choose the ratio of small ships to large</strong>  Groups the available encounter candidates into small and large ship/asteroid bases and randomly generates encounters based on the ratio of each that you chose.  Warning: selecting high large ship ratio's may cause performance problems, especially when used in conjunction with large encounter groups.

* <strong>Set large ships selection to be restricted to only those with turrets</strong>  If you want a challenge every time you find a hostile encounter ship then be sure to click this.  Restricts the large grid encounter candidates to those that have turrets on board.  Doesn't currently distinguish between gatling, missile and interior turrets. All count as armaments.

* <strong>Include and exclude specific ships for encounters using the encounter selection screen</strong>  Use the encounter selection screen to provide full control over the encounter candidates available in game, either using the flexible filter system or by clicking on each individually in the ships table.

* <strong>Review ship block statistics</strong>  Hover over the 'blocks' and 'turrets' columns in the encounter selection screen to see a detailed break-down of the encounter candidate blocks.

* <strong>Themed Presets allow you to quickly choose a set of options that suit a preferred game style. </strong>  Choose from Peace, Adventure, War and Ruins.  All settings will be modified to match, with only tweaking required.  The Random preset will set every setting randomly before closing the screen, so that you can't see what was selected.  Peace mode will replicate the settings used by the current vanilla default game mode.

This change includes and makes obsolete the changes for:
https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/272

The only difference between that change and this one, that isn't included above, is that the word 'hostile' has been removed from the antenna name; it looks better without and doesn't require translation.  Obviously there is also no need for the 'Hostile Encounters' checkbox any longer.

I have had some difficulty deciding on maximum values for the number of ships allowed in each group and the maximum amount of damage that can be applied, as both settings can affect performance.

For the max number of ships, the reason for the performance problems are obvious, it's the same reason that having loads of ships in your world causes it to slow down.

The performance hit for damage is less obvious.  I've set it up so that each damage severity level randomly affects a different number of blocks.  So the 'Almost new' setting might affect up to 2% of blocks, while the 'Heavily Damaged' option might affect up to 20%.  Non-functional blocks get destroyed completely (distorting the blocks around them) and functional blocks (thruster/solar panels/etc.)  are damaged so that they start their various damage effects (sparking/flame/gas, etc.).  Looks good.

At the lowest settings the ship looks like it's been hit by micro-meteorites, but is otherwise structurally sound, while at the mid-range settings, they look crumpled and distorted.  At the highest settings, the ship can start to break up (depending on the ship design) and that effectively adds more ship grids as each section separates.

I did have it so that  the highest setting completely gutted the ship/station, however the resulting loss of block cohesion, when applied to 5 large ships at a time resulted in hundreds of twisted ship sections floating off into space.  It looked incredible and is exactly what I'd like to see in the game, however I've had to reduce it as it caused too many resource problems.

With the maximum values available, both settings now seem stable and performant for me, on the 32-bit version, under high spawn density, and it may be possible to raise their cap to much higher values on the 64-bit versions of the game.

I've probably missed something obvious, so if you have any questions please let me know.


Geneticus:
Hmmm, only thing I can think to add would be:
Max Block count(or mass) for stations, 0  = no limit
Max Block count(or mass) for small ships, 0 = no limit
Max Block count(or mass) for large ships, 0 = no limit
ships vs stations to be encountered 
small vs large ships to be encountered

CodeHamster:
Geneticus;  thanks, I'd considered the last option, but not the others.  I may find a way to add those options, however so far I've managed to avoid modifying anything too fundamental about the way that the ships are spawned and adding in the options that you mentioned may change that.

If you leave the settings on default (as shown in the pic above) then the logic that is (or at least should be) executed will be exactly the same as the logic used by the game currently.

The system basically pulls a ship or two out of a hat and says 'OK, whatever I get I'll just add it in to the game as is.'  It then loads in the ship definition for each.  It doesn't do much/if anything with regards to trying to balance the ships that are chosen prior to that point, and I suspect that trying to select them according to pre-conditions may mean have to load up all of the prefab types into memory and then trawl through all of their grids to check whether they meet your criteria.

I may be able to pre-cache the prefabs in some way at the beginning of the game, I could then obtain the stats needed and use those to select the ships as each encounter is spawned, however that's a fairly drastic change and may force me to modify how the standard system works, which could reduce the chance that the change is incorporated into the game.

I'm not giving up on that idea though and if I can find a safe and performant way to do it then I will, as I'd like to be able to add an option to ensure that unarmed ships are always guarded by armed ships.  If I find a way to do that then I will include your options also, if you don't mind?

Until then, I may look at adding some preset buttons to the top of the page that sets the sliders to pre-chosen values, as it's tedious to go through all of the options on the screen and set them if you want something different to the norm.  Was thinking of something along the lines of labels like 'Peace/Default', 'Explorer', 'War!' and 'Post-Apocalypse'.

After that, I will see if I can do something about balancing the inventories.

zrisher:
I haven't looked through this PR, but I know that the vanilla encounter generation logic runs **after** all vanilla and modded prefabs have been loaded. So you shouldn't need to do any special caching.

CodeHamster:
Then good times are ahead. :)

CodeHamster:
@Geneticus 
I didn't have much time over the weekend to work on this, so I just tidied up what I'd already done and added the themed presets buttons mentioned earlier.

I've been looking at options to add the preferences that you mentioned and while I can do it in-game, I'm not happy with the results so far and so haven't committed them.  To be able to get at the grid size and block count, I'd have to load the grids for all of the ships every time an encouter spawns and while I've done it, it's not a very efficient way of making the changes needed.

I also think that having preferences for the max block counts alone isn't very intuitive, as you need a fairly intimate knowledge of the encounter ships in advance in order to use it and since most people won't have that, it will end up being an unsused preference.

Instead I propose having a list box on the new Encounters Configuration GUI, shown in the picture above, with a list of the encounter prefabs (by name) on it along with a few basic stats including the # blocks and the mass (if I can get to that easily).  You will then be able to filter out ships that you don't like using that screen by either selecting them manually or by applying filters such as those that you mentioned.

I think that I will also make the list visible, only if the player wishes to see it, as it could spoil the surprise if you're a new player and haven't seen many of the ships yet.

The other advantage is that it offers a point where other metadata stats (such as whether a ship is armed or not) can be farmed in advance, just once, without hitting the in-game performance every time an encounter spawn happens.

I've got the list box in place and have the logic need to fill it, however I've not had time to integrate them yet.  Hopefully I will be able to do so this evening; if not then I will probably wait until after the decision has been made about which changes will be merged by Keen this week, to avoid drastic (and half realised) changes to the interface while they're deciding.

Geneticus:
Honestly I'm not that invested in it. I was merely suggesting some options that seemed to be things that players would want to tweak. Don't put too much effort in on my account. There are probably easier targets for us to snipe at to make things more extensible.

CodeHamster:
@Geneticus   No worries, I'm enjoying myself.  It's quite fun to see what can be done, even if it never gets merged. :)

CodeHamster:
@Geneticus I managed to get the list box to fill with the ship details, however it was taking about 17 seconds to deserialise all of the ship data (using an SSD), even though I've trimmed down what was returned to the bare minimum needed.  While I personally, wouldn't mind waiting that long, I don't really see it as an option for release.

I could mess with the deserialisation routines or write my own light-weight reader, but that's a bit beyond the level that I want to go to at the moment.

About the best that I could do, without incurring the delay, would be to allow the player to remove ships from the spawn list by name and I may take another look at that again in future, although for the moment I'm not including it.

Geneticus:
Well, the vanilla ships are static content so you could just add an XML that already has the data in it and read it for checking the size. Mods could also add the info file. If they don't just default them to medium for large grids and small for small grids.

CodeHamster:
@Geneticus  If I don't find a better way of doing it then I'll give that a try.  For now, I'm going to put that bit on hold, move on to providing inventory balancing options, and will come back to it later.  It will give my brain time to work on the options. :)

CodeHamster:
@Geneticus 

I couldn't leave the idea of being able to choose your ships alone and so I've worked out a way to do it that is much faster than the method I had before.  Ended up writing my own xml reader.  I've added a couple of the prototype pictures below, as I've not released the code yet, as there's still a lot of tidying up and testing needed.

new encounter configuration medium

encounter ship selection 001 medium

CodeHamster:
Uploaded the new encounter selection GUI, which works almost as described above.  Changed the filtering system, as it was overly-complicated and didn't make sense. 

Note:
I added a MyPrefabProfileDefinition POCO, to hold profile information on the ships.  Wasn't entirely sure where to place it, so may need moving.

SpaceTexts will need to be added for the tooltip block name descriptions.

Will do the above and refactor, as and when the time is available.

Geneticus:
I was thinking about that. I'd have added tags to the SpawnGroups and read
them from there. Glad to see you worked it out. Would your XML stuff work
here <http://forums.keenswh.com/threads/7362537/>?  Maybe you have more
time than I do... :P

CodeHamster:
@Geneticus 

:) My girl-friend would probably disagree, I'm in serious quality time debt to her at the moment.  Part of the reason I'm publishing now, as I need to earn some major brownie points over the next couple of weeks and so won't have much time to do SE stuff.

Going to reply to your query on the forum, as I'd like to keep this thread oriented around this change.

CodeHamster:
@OndrejPetrzilka 

I appreciate that you're busy with planets and that this is a relatively big change to review, but I'd appreciate some feedback (encouragement or critique) on this change, please.  

It's taken a month of weekends and evenings to get the code this far and I could easily spend another month on it, improving it and adding all of the extended options that I have in mind.  

For instance, I still need to add inventory management options and I've been toying with the idea of having a 'specials' section that allows you to configure preset spawn scenarios for ships (so you could set up a scenario where random encounter ships have all of their forwards thrusters destroyed or where they transmit an S.O.S. message, etc.).

Unfortunately, my time is going to be limited over the next few weeks and so I would like to focus on changes which are going to be useful.  Even if the answer is 'don't bother with this, we've got our own version that we'll release in six months' then I can move onto something more useful in the mean time and won't need to keep applying your merge files every few days.

If you just need me to modify the way that the code is working in some way before it is acceptable, then I will happily do my best to comply.

I hope you find the time to respond and that you like the changes that I've made.

mexmer:
any reason to delete this?
it was not bad adition, and most likely cannot be done as mod.

CodeHamster:
Thanks Mexmer; spending way too much time on SE related activities and it's getting in the way of real life, so I'm backing off the game indefinitely and abandoning this project.  

There wasn't really much interest shown in this and, in truth, I don't think Keen will add it, but more than that I worry that if they do, then I'll spend the next six months with my head buried in the sand trying to fix any problems that arise from it and extending it.  Can't really afford that at the moment.

Even with the branch deleted, I believe that Keen can access and use the code, so if they or anyone else wants to do something with it or take it further, then they're very welcome too.  

Leaving the fixes that I added, as is, because they probably will be added at some point and will save someone some pain.

Have fun!

mexmer:
i'm not sure, if you can pull deleted branch, afaik you cannot.

as long, as you don't delete repo, it will most likely be possible to retrieve individual commits as patches, but not the branch itself.

zrisher:
@mexmer is correct, there's no way for us to use a branch that's been deleted, or even know what commit it pointed to to determine whether that commit exists on one of your other branches (i.e. `Encounter-Configuration`).

mexmer:
thanks

HashtagMC:
<blockquote>CodeHamster wrote:
There wasn't really much interest shown in this</blockquote>
However it's a useful extension and I'd be happy if this would be merged into Space Engineers. But of course real life has priority.

PR #290, opened 2015-06-18
Advanced Door additions and fixes

Eikester:
New PR because i made things worse in the last one trying to rebase it (meh)

Anyway:

Fixed OpenRatio returns values > 1
Fixed OnStateChange always adds new Subparts
Added requested feature AutoClose


kleshchevand:
Your AutocloseInterval doesn't look like something entirely necessary since same functionally can be done by scripts or timers, thus I'm not sure pull will be accepted. Beside some people need autoclose, others will want autoopen, auto-shutdown or something ridiculous (like stuck/broken door behavior). Implementing just one simple behavior makes little sense in my opinion.
Why not separate those two fixes from AutoClose and make AutoClose into independent and might be bigger feature?
P.S. May be something like SetOnOpenDelegate()/SetIdleDelegate() and link it with xml (sorry, not sure how SE works, so this is closer to wild guess) and provide 'example open/close timer' delegate that can be set from xml and/or code? If moder is not satisfied with existing options he can create own delegate and set it in block-related xml record as default behavior with his own parameters. This should be much more flexible in comparison to hardcoded close-timer - moder will be able to implement own autoclose/autoopen/shutdown timer, create 'hatch' that won't open unless both sides have/have no oxygen, some kind of 'detect owner' behavior, turn off door in the middle of opening, open half of the door e t c.
Actually feature that has higher chance of being accepted: UI flag to allow any player to open door (SE does this for buttons just fine, why not doors as well?).

Eikester:
#317 Contains only the Bugfixes without the Autoclose feature

PR #291, opened 2015-06-18
Fixed IMyGravityGeneratorSphere.Gravity's value

THDigi:
The `IMyGravityGeneratorSphere.Gravity` and `IMyGravityGenerator.Gravity` return different values for the same slider value. The flat one has 0.0 to 1.0 range while the spherical one has 0.0 to 9.81 range.

This PR modifies `IMyGravityGeneratorSphere.Gravity` to also return 0.0 to 1.0 value ranges just like the flat one (which was the first one), I also added summaries for the fields for clarification of returned values.


plaYer2k:
To be honest, i would prefer to use metric values consistently and thus 9.81 m/s instead of 1 G.
But i guess that is just a preference and others sure think differently.

Having a poll to decide whether to pick 1 G or 9.81 m/s might be the best, unless a keen member comes in and decides to pick one of them based on their guidelines.

Thanks for fixing that inconsistency though, i wasnt aware of that as i almost never use sphericals.

THDigi:
Alright, here: http://strawpoll.me/4686277 cast your vote guys :P

THDigi:
Seems the votes are leaning towards m/s.

@OndrejPetrzilka @No-Brain any preference from you guys ?

PR #295, opened 2015-06-20
Add InterlockDoor

yajiedesign:
Use this PR.the door can Interlock, One door open, other auto close.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyDoor.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyDoor.cs
@@ -12,6 +12,7 @@ using Sandbox.Engine.Utils;
 using Sandbox.Game.Entities.Cube;
 using Havok;
 using System.Reflection;
+using System.Threading;
 using Sandbox.Common;
 using Sandbox.Engine.Physics;
 using Sandbox.Game.World;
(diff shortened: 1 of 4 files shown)


THDigi:
From the code I see you're using GPS coordinates, wouldn't it be simplier to have a dropdown list or just type the name of the other door ?

yajiedesign:
@THDigi not GPS coordinates, it is ship block postion.it it private of the ship.i am only use GPS Parse.
 use name, name will repeat. 
in this rp.you choose one door.push copycoord, goto other ,paste it,set ok .
use dropdown  also can be.
I'm not sure that is better

yajiedesign:
@THDigi  i have a video  http://youtu.be/YCiWRNFME84

THDigi:
It's using GPS, you're copying and pasting coordinates.

Just use a name box and that's it, people can copy/paste the name instead which will be 3 buttons less for the UI which will also work if the ship is moving, while the GPS most likely does not.

You can further improve the name box with an autocomplete list that drops down showing all doors with the matching names as you type and clicking on one will paste that in.

yajiedesign:
@THDigi  it is not GPS coordinates.. only use LaserAntenna text.I can change text.
it is block position in grid, it not change when ship move.it also work if the ship is moving
i am get pos use MyDoor.Position,you can check it.
autocomplete  drops down list is good idea.but MyTerminalControlCombobox not autocomplete.I will try to do it self.

THDigi:
Ah, I see now, sorry for dragging that on xD

The delayed open/close could be a checkbox option, yes. Maybe even automatically set on both doors when you check/uncheck it from one of the doors.

PR #296, opened 2015-06-20
Fix. Command line argument transfer.

JimLess:
Fixed. Only first parameter transferred

### DIFF
--- a/Sources/VRage.Dedicated/Configurator/ConfigForm.cs
+++ b/Sources/VRage.Dedicated/Configurator/ConfigForm.cs
@@ -69,7 +69,7 @@ namespace VRage.Dedicated
             else // Local / Console
             {
                 string[] cmdLine = Environment.GetCommandLineArgs();
-                Process.Start(cmdLine[0], ((cmdLine.Length > 1) ? cmdLine[1] : "" ) + " -console -ignorelastsession");
+                Process.Start(cmdLine[0], ((cmdLine.Length > 1) ? String.Join(" ", cmdLine, 1, cmdLine.Length - 1 ) : "") + " -console -ignorelastsession");
                 Close();
             }
         }


Sources/VRage.Dedicated/Configurator/ConfigForm.cs
```diff
@@ -69,7 +69,7 @@ private void startButton_Click(object sender, EventArgs e)
             else // Local / Console
             {
                 string[] cmdLine = Environment.GetCommandLineArgs();
-                Process.Start(cmdLine[0], ((cmdLine.Length > 1) ? cmdLine[1] : "" ) + " -console -ignorelastsession");
+                Process.Start(cmdLine[0], ((cmdLine.Length > 1) ? String.Join(" ", cmdLine, 1, cmdLine.Length - 1 ) : "") + " -console -ignorelastsession");
```
harag-on-steam: According to [`ProcessStartInfo.Arguments`](https://msdn.microsoft.com/en-us/library/system.diagnostics.processstartinfo.arguments%28v=vs.110%29.aspx) arguments that contain strings need to be surrounded by quotes. And arguments containing quotes need to be escaped. 

Also, what happens if one of the `cmdLine` arguments is `-console` or `-ignorelastsession`? Can the started process handle duplicate command line arguments?

JimLess: "Also, what happens if one of the cmdLine arguments is -console or -ignorelastsession? "
nothing. because this stage skipped...
"Can the started process handle duplicate command line arguments?" - why?

"According to ProcessStartInfo.Arguments arguments that contain strings need to be surrounded by quotes. And arguments containing quotes need to be escaped." - Ok. yep if it need to run properly. u can modifiy code by yourself. 

but, this is a quick bugfix, not a entire refactoring

harag-on-steam: Well if `cmdLine` already contains `-console` it will be part of the `String.Join` and the `+" -console...` will concatenate it a second time. I'm just asking if that is ok for the new process.

And I'm pointing out that there is more than one bug here. So this line will need to be touched again.

JimLess: "+" -console... will concatenate it a second time. I'm just asking if that is ok for the new process."
-console option check by Contains method. so i see no problemo...

JimLess:
up-to-date

PR #299, opened 2015-06-20
API: IMyOxygenBottle

mexmer:
Interface to retrieve OxygenLevel and Capacity of oxygen container in inventory

```
 IMyInventoryOwner owner = (IMyInventoryOwner) blocks[0];
    IMyInventory inventory = (IMyInventory)owner.GetInventory(0);
    var items = inventory.GetItems();
    for(int i=0;i<items.Count;i++)
    {
        if (items[i].Content.TypeId == typeof(Sandbox.Common.ObjectBuilders.Definitions.MyObjectBuilder_OxygenContainerObject))
        {
            var bottle = items[i] as IMyOxygenBottle; 
            Echo ("Bottle found, fill level " + String.Format("{0:P2}", bottle.OxygenLevel) + ", capacity " + bottle.Capacity + " units");
        }
    }
```

capacity is important, because of modified bottles, level might not be enough to estimate, how much oxygen you have left (play time wise)

i've decided to not add "set" function for modapi for direct oxygen level modification trough mod script, because currently there is no simple way to invoke value synchronization ... this might change in future.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMyOxygenBottle.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyOxygenBottle.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Sandbox.ModAPI
+{
+    /// <summary>
+    /// Interface for accessing oxygen bottle properties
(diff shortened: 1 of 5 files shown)


Sibz:
:+1:

mexmer:

PR #301, opened 2015-06-21
Programmable block and grid program logic separation, proper constructor support, rare potential sync fix

lord-devious:
This is the first of two pull requests which aims to make it simpler to maintain the state of a program. 

This first branch separates the logic of grid program maintenance from the programmable block, making it easier to maintain. 
Coincidentally, it should also fix a potential bug where the storage did not always get synchronized properly in multiplayer.
Finally, and this is where the direct correlation with state management comes in: It is now possible to create a proper constructor in your scripts, which is invoked only once during the lifetime of a grid program.

```c#
IMyDoor _door;

Program()
{
    // Called only once after world load or rebuild, 
    // before the very first time Main is called
    _door = GridTerminalSystem.GetBlockWithName("MyDoor");
}

void Main()
{
    _door.ApplyAction("Open_On");
}
```


THDigi:
Hmm, does Program() execute only once during a grid's lifetime or on every run before Main() ?

That example is probably not a good one if it executes only once since MyDoor could be grinded down and moved so the existing _door pointer would be unusable or something.

lord-devious:
Read the comment :) I'm not trying to teach how to write safe scripts, only to show how the constructor works. Besides I have a billion scripts that would work very nearly like that (most of them, actually), because if a door was removed or destroyed, the script would lose its purpose anyway. I see no problem having to rebuild it after repairs.

THDigi:
The phrasing is confusing then, from the "and always before Main is called" I got that it was also executed along with Main(), which would be indeed pointless, but nevermind :P

lord-devious:
You have a very good point - rephrased. Better now?

lord-devious:
I will be rewriting this.

PR #305, opened 2015-06-23
Slave assembler pathfinding fix with six-sided conveyor blocks

phreakscp:
Slave assemblers would not pathfind through conveyor blocks since it is impossible to set ownership on conveyor blocks

This made it impossible to find their master
A workaround would be to set the slave assemblers to "Share with All", but some play styles may find this workaround unacceptable

This commit ignores ownership of conveyor blocks when pathfinding for the master assemblers from the slave.


mexmer:
you should rebase this. maybe squash also, but rebase for sure.

phreakscp:
Ok, done.

mexmer:
looks much more tidy now :+1:

OndrejPetrzilka:
Imho instead checking "is MyConveyor", there should be check if the block can not have ownership.

phreakscp:
I replaced the 'is' check with a property that'll return true if derived from MyTerminalBlock, false otherwise.  This should give KSH some more flexibility to expand upon that property.

Geneticus:
Does this incidentally also fix Assemblers not being able to see resources in your conveyor network for calculating the amount of ingot of material that you have available for assemblers to pull? I thought that bug was also related to conveyor pathfinding...

PR #308, opened 2015-06-24
FIX: UI icons affected by background transparency

mexmer:
UI icons texture is affected by background transparency, that makes icons like "filter storage" less visible than rest of UI controls.
This fixes this issue on icon based buttons in terminal GUI.


mexmer:
this fix is part of #311

PR #309, opened 2015-06-25
call BeforeMerge on both MergeBlocks

zrisher:
The `BeforeMerge` event, which seems to be only used by `ModAPI`, is triggered on `MyShipMergeBlock`s when one grid merges into another. Currently, `BeforeMerge` may be called on the grid that is receiving a merge OR the grid that is being merged. 

For my purposes as a mod author, this was insufficient - I needed to know if a grid was being merged into, and only sometimes would this event tell me. Thus whenever I received a BeforeMerge I would need to find the other merging grid and mark it, which is very inefficient considering the place where this event is raised already knows exactly what the other merge block is. It's hard for me to imagine this event was useful for mod authors when it will only be raised sometimes and you have no idea which side of the merge you're on.

This is a very simple fix - just raise the event on both MergeBlocks. Alternatively, I could raise separate events for each side or pass a boolean to the event representing if that block was the merger or mergee. Happy to code it up that way if Keen/others prefer. This is sufficient for my purposes.

There are two commits here - the first demonstrates the issue with some logging. Checkout that commit, go merge & unmerge some differently-sized nonstatic grids 5-10 times, and look at the log to see what I mean. The second commit removes the demonstration and brings this to its simplest form, which you see below. Happy to squash these into one commit if Keen prefers.


zrisher:
@No-Brain could you please consider this for inclusion this patch? Please let me know if any issues, happy to fix them. The players on my server would really appreciate it.

zrisher:
Rebased to lastest master, appears to still be needed.

PR #313, opened 2015-06-26
Production screen quantity dialog

phreakscp:
When right clicking blueprint during assembly/disassembly or the lower inventory grid during disassembly, the player will now be presented with a text box that allows them to enter a quantity of the item to be produced or disassembled.


FavioGalvis:
:+1:
* Quality of life PR, does not change actual performance or functionality.
* Supported!

Issue #292, opened 2015-06-19
Building on indestructible grids


devu:
It should not be allowed to build on grids you cannot destroy. Unless by admin of the server. In fact Indestructible nature of grid could be omitted if done by admin as well.

mexmer:
if you have complaint or suggestion about new feature ... use forums ... github is not for discussing user suggestions.

JimLess:
agree with devu, this is not suggestion, this is bug. u can grind indestructible blocks. u can weld it, u can build on it. it's bug...

devu:
Well, It worked that way only last Thursday.
Instead addressing it, they turned the grinder on for indestructible grids.
I really hope this will come back. One of most 'must-have' features for
public servers with some more game play.

mexmer:
it's not community made change.
if you implore it's bug, then report it on forums, that's one and only official way to report bugs.

devu:
Jump drives are not community made changes and you are still interested about it. I thought that would be something that community could pick up too.

JimLess:
i still wait this bugfix. so some must do this work...

Spartan322:
This isn't a github issue, this does not belong here, go to the forums.

devu:
But community could look into it. It's more asking for help rather
reporting. Keen wasn't able to look at this just yet ad it was reported
many times already.
If I understand correctly and everyone can see it, community made changes
sometimes getting adopted.
I do not understand why you were so quick to send me out with this
particular bit. While the other bits are being worked on.

Again, did Jump Drive cluster issue is a community made bug? No, but you
picked this up. I simply don't get it.
I can't go and report it to Keen because it was already done. Then I will
get banished for not reading already reported bugs :D

Maybe I started wrong. Let me try again.

Could community look into this problem please?

Spartan322:
Not every bug is to be reported here, most (if not all) go to the forums, also, you didn't even start off stating it is a bug, you stated a suggestion, this is not a suggestion box, please go to the forums.  Only code related suggestions go here (of which would be appropriate in the forums as well)

devu:
I see plenty of reports and suggestions here too. But ok, probably I am not
a part of a circle of trust here. Thanks for clarification.
This subject will stuck in the limbo then, because it was already reported
and nobody is willing to pick this up.

zrisher:
This is definitely a bug, but I think I see where they're drawing the line, maybe this is accurate and can help people understand what goes where:
[...]

PR #314, opened 2015-06-27
Damage Routing System.  Track, mitigate, amplify, and nullify damage to all things in game.

Tyrsis:
This adds the ability to hook into damage that is done on objects in game, and observe or modify it.  This also allows you to track if objects are destroyed.

Added a new game system called MyDamageSystem internally in Sandbox.Game.  This can be used internally if new objects are added and damage is done.

For mods, added IMyDamageSystem to IMySession.  This allows mods to hook into damage and modify it via:

MyAPIGateway.Session.DamageSystem.RegisterDestroyHandler(int, delegate);
MyAPIGateway.Session.DamageSystem.RegisterBeforeDamageHandler(int, delegate);
MyAPIGateway.Session.DamageSystem.RegisterAfterDamageHandler(int, delegate);

Destroy, and AfterDamage delegates are for information purposes.  BeforeDamage delegate allows you to modify damage before it's applied.  This allows you to mitigate, amplify, or nullify damage (including deformation).

Each function allows a priority.  The handlers with a lower priority handle damage first.

Test Map and Test Mod:
MOD: http://steamcommunity.com/sharedfiles/filedetails/?id=470287428
MAP: http://steamcommunity.com/sharedfiles/filedetails/?id=474190531

The mod basically outputs damage to chat and to a log.  It also puts a notification on your screen when something is destroyed.

You can also colour blocks to nullify, mitigate, and amplify damage as follows:
Color block 360, 0, 0 to amplify damage by 2
Color block 0, -100, 0 to mitigate damage by half
Color block 0, 0, -100 to nullify damage and make it invincible

This shows how easy it is to modify how damage is done in a very simple way.


JimLess:
Nice.

yajiedesign:
good work.

charlesdumas:
Good idea, can you also add some OnEntityDamaged and OnEntityDeformed events on Sandbox.Game.Entities.MyEntities similar to the existing OnEntityAdd and OnEntityRemove so we have a global hookup to catch damage? This would avoid having to subscribe to every object individually. Moders could then use game logic on the damageable entities that these events receive to filter out what they want.

Tyrsis:
@charlesdumas I talked to Ondrej about this, and he suggested something similar, in that there should be a SessionComponent for damage, similar to logic.  But they are currently overhauling components right now to properly aggregate when you want to attach multiple times to the same type.  

Right now I'm just making sure I'm testing all damage situations, and then implementation of how you can interact with damage will be looked on more.  There definitely should be a more "global" hookup as well as priorities as order that damage is applied is important.

charlesdumas:
That's great. I just thought it went well with the current events on MyEntities. Let me know if you need help testing or want to throw design ideas at someone. My steam user is "cdumas". I was developing something similar as a service to an asset protection mod to lock down scenario objects that shouldn't be destroyed (like spawn platforms) that would also allow anti-grief protection on public survival servers (ex. allow each player to lock down one single station object).

I don't suppose this would work on asteroid objects though. I've been making decorative asteroids that are only empty shells, they take a lot less system resources than full ones which allows very rich scenery.

Tyrsis:
This needed more than a branch update :P

This should work now, but please note I had to modify MyEntityStat slightly to pass along information required since it appears now an entity state script (???) control the health stat.  It may not be ideal though, but I tried to modify as little as I could.  

Stat modifications will sometimes require information to be attached to them (why they were modified and by what modified them).  Especially when it comes to health.  Right now it's passed as an object, but perhaps some sort of standard way in MyEntityStat should be created.  For now I pass it along with modifying the value.  This in turn gets passed back when the script kills the character.

devu:
Hi Tyrsis,

I was trying to make a case from some time on this subject:
https://github.com/KeenSoftwareHouse/SpaceEngineers/issues/292

You seems to tackle the problem from different angle and improve abilities
of modding API and your change was implemented in today's patch.
My question is, have you ever been concerned about extra overhead behind
this solution? It requires to fire up whole bunch of events no matter if
you using a mod or not that can take advantage of this.

My both servers I am running almost tripled average In after today's patch
and I wish the reason is somewhere else but looking at all those changes
today it seems to be the only reason.
I would rather see such mechanism to be built into the core of the game as
I wished for from the very beginning if that is the case.

Could you enlighten me if your change could be the reason? So far the game
on dual CPU xenon E5 is unplayable for more than 5 people with full
production on. When I have 2 refineries and 1 assembler limit set on my
worlds.

zrisher:
I can confirm that my host has seen increased CPU usage on all worlds since Thursday's patch, though  I have no idea if this was related.

Tyrsis:
With the source available, you can pretty much clearly see that these additions do almost nothing to vanilla.  The only thing ever added is the long added ON DAMAGE that could possibly sync along with the damage packet, which is not really a lot, and I do never resolve that long at any point, as it's up to the mods to use that data if they want (for example I never do a TryGetEntityById)

If you look at the source, all that is done is basically "Does this entity use damage system, if so raise events registered**" Important, because if no damage events are registered, this does NOTHING except looks at the event hashset, sees it's empty and returns.

devu:
Thanks for heads up. Don't get me wrong, it was simply a question to
clarify.

The fact remains that network lag is much greater after latest update and I
reacted on Friday when only your additions were added to the patch.
But I can be wrong on that and it's a bunch of internal bug fixes a day
before.

I officially giving up on trying to run server at sim speed 1.0 because
there is now way to put limits like half refinery and quarter assembler per
player.
I'm sure your addition comes handy in many cases and one could be to build
such anti-grief mod based on that.

However I'm still convinced that having it already in code (indestructible
grid) it would be much better to have it implemented on the core since it's
already there.
Having indestructible grid you can build on and grind (as a quick fix of
the problem) has no sense whatsoever. We ending up with parallel universes
not only in space ;)

PR #315, opened 2015-06-28
FIX: news "next button" not working

mexmer:
next button is overlapping with vertical scrollbar when news is too long.
this change adds to calculation also "height" of next button, so news text area ends right above "next" button and scrollbar doesn't catch mouse click that should belong to "next" button.


mexmer:
cleanup - no longer working on this - also it's fixed AFAIK

PR #316, opened 2015-06-28
Fix small source of oxygen loss when pressurizing/depressurizing.

JoeTheDestroyer:
Currently, when calculating how much oxygen an air vent can consume or produce, the amount of oxygen needed or remaining in the room the vent occupies is ultimately multiplied by a delta time factor. Because updates happen every 100 frames, this means these values are artificially increased by 67%  (i.e. *100/60). The result is that an air vent claims it can consume or produce more oxygen than allowed, resulting in either oxygen being lost (consuming) or being magicially created (producing).

This appears to be due to a typo, as only BlockDefinition.VentilationCapacityPerSecond needs to be multiplied by a time. This patch fixes this.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyAirVent.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyAirVent.cs
@@ -454,7 +454,7 @@ namespace Sandbox.Game.Entities.Blocks
                 neededOxygen = 0f;
             }
 
-            return Math.Min(neededOxygen, BlockDefinition.VentilationCapacityPerSecond) * deltaTime;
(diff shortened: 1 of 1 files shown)


JoeTheDestroyer:
Currently, when calculating how much oxygen an air vent can consume or produce, the amount of oxygen needed or remaining in the room the vent occupies is ultimately multiplied by a delta time factor. Because updates happen every 100 frames, this means these values are artificially increased by 67%  (i.e. *100/60). The result is that an air vent claims it can consume or produce more oxygen than allowed, resulting in either oxygen being lost (consuming) or being magicially created (producing).

This appears to be due to a typo, as only BlockDefinition.VentilationCapacityPerSecond needs to be multiplied by a time. This patch fixes this.

PR #320, opened 2015-06-30
Server Join from API

Tyrsis:
This allows scripts to have a player join another multiplayer server.  Same mechanism as if they did a +connect.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMyMultiplayer.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyMultiplayer.cs
@@ -25,6 +25,7 @@ namespace Sandbox.ModAPI
         bool SendMessageToOthers(ushort id, byte[] message, bool reliable = true);
         bool SendMessageTo(ushort id, byte[] message, ulong recipient, bool reliable = true);
 
+        void JoinServer(string address);
 
         void RegisterMessageHandler(ushort id, Action<byte[]> messageHandler);
         void UnregisterMessageHandler(ushort id, Action<byte[]> messageHandler);
--- a/Sources/Sandbox.Game/ModAPI/MyModAPIHelper_ModAPI.cs
+++ b/Sources/Sandbox.Game/ModAPI/MyModAPIHelper_ModAPI.cs
@@ -242,7 +242,20 @@ namespace Sandbox.ModAPI
                 SyncObject.SendMessageTo(id, message, recipient, reliable);
                 return true;
             }
-       
+
+            public void JoinServer(string address)
+            {
+                if (MySandboxGame.IsDedicated && IsServer)
+                    return;
+
+                System.Net.IPEndPoint endpoint;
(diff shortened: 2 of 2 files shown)


joemorin73:
What?  wait?  is this in reference to the "teleport" discussion on the forums?

Tyrsis:
Not sure what discussion you speak of, but this is a feature I've wanted for awhile.  It will allow a script controlled by the server to send a user to another server for load balance / wormhole style travel between multiple servers.

joemorin73:
Check here:
http://forums.keenswh.com/threads/teleportation-between-servers.7362708/

Tyrsis:
Yes it can be used for that.  But before this source mod there was no way to trigger a server join of another server via api.  This PR changes that.  The other stuff discussed in that thread are different issues completely, all this does make a user join a server.  What you do with that is up to the implementer.  I have 2 servers (Transcend) and I will use it to allow users to travel between the two servers (both servers have synchronized player lists, faction lists and mod lists, so it would be safe to do so).

joemorin73:
I had a similar idea.  The PR is definitely required, but there is a lot more work required.  I'm assuming you will be making an SESE plugin to manage transferring entities?

Tyrsis:
There is a lot more work required, but all that can be done on the server.  This part is a client requirement that had to be met before anything else can be done.

Qithe:
Whoa Whoa Whoa, hold my hoses. Are you seriusly working on a mod/script that would alow jumping betwen servers? 

(Yes, I'm the same guy from the forum. Im the OP)

devu:
Cross server join is indeed very needed in admin community. Thank you for
picking this up Tyrsis!.
But the ultimate goal would be cross server jump. Are you thinking about
this as well?

zrisher:
> (both servers have synchronized player lists, faction lists and mod lists, so it would be safe to do so).

Yes, this part would be great to have in Core too so that everyone can accomplish this "server jump"

Tyrsis:
I renamed it, as it seems to be a little bit misleading.  This is just a server join from API.  This doesn't include synchronization that would allow movement of objects from one server to another.  This just moves clients.

devu:
It wasn't misleading, totally got the idea and functionality behind. Just
saying is just a step closer to object movement across servers that would
be a ultimate goal.
But you kind of answered my question. However with player ID in sync across
different instances it would be much easier to create such mechanism anyway.

If you are focused on client join only, would be possible to join with a
custom flag? To be able to recognize if someone just joined instance
directly or joined from API?

IhanaMies:
With this commit is it possible to connect servers via GPS locations since jump drives were added?

joemorin73:
This commit adds the ability for to change servers via mods. It still requires mods to be made to use it.

JimLess:
now i can use it in my mod =)

joemorin73:
If the community merges are added in today's update.  If not, one week.

PR #322, opened 2015-06-30
Fix oxygen loss when room has multiple air vents

JoeTheDestroyer:
Problem: Significant oxygen loss can occur when multiple air vents are used to pressurize or depressurize a room.
http://forum.keenswh.com/threads/01-077-oxygen-loss-when-depressurize-a-room.7357431/

Cause: The current oxygen production/consumption algorithm queries air vents (all consumers/producers) without regard to which room they are attached to. Additionally, each air vent reports how much air it can use without regard to other vents attached to the same room. Finally, the possibility that an air vents requested oxygen amount could change as other vents are satisfied is neglected.

This leads to a problem when a room is near full (i.e. needs less than the air vents could supply in an update). If a room has X air vents and needs Y oxygen to be full, each of the air vents will request the full Y oxygen. A total of X\*Y oxygen is extracted from producers/tanks. When the fill process begins, the first air vent supplies Y oxygen to the room filling it. The remaining (X-1) air vents try to supply Y oxygen, but it is discarded in MyAirVent.Consume() because the room is already full, wasting (X-1)\*Y oxygen.

A similar event can happen when depressurizing (leading to the creation of oxygen).

Proposed solution: I have rewritten the consumption algorithm to query and satisfy consumers sequentially so that each consumer can accurately report the oxygen it can use (e.g., the remaining air vents in the example above would request 0 oxygen instead of Y).

One downside to this approach is that oxygen usage is no longer evenly balanced across all consumers for a given priority level (e.g., two equally sized, equally vented rooms can fill at different rates). I have attempted to mitigate this by randomizing the processing order within each priority level. The effect should be that usage will average out over time.

Testing performed: I have tested in a special creative world with 3 different room sizes from a single block to 7^3 blocks and with 1-10 air vents (where possible).  With the patch in place, oxygen can be cycled between the room and an oxygen tank without noticeable loss/creation (to 6 significant digits). In 1.088.012, the same setups show variation up and down of the oxygen volume, with only the first significant digit remaining constant. Various basic operations were tested on a single room (various combinations of filling/emptying rooms/tanks/generators).

### DIFF
--- a/Sources/Sandbox.Game/Game/GameSystems/MyGridOxygenSystem.cs
+++ b/Sources/Sandbox.Game/Game/GameSystems/MyGridOxygenSystem.cs
@@ -22,6 +22,7 @@ using VRage;
 using VRage.Components;
 using VRage.Input;
 using VRage.Library.Utils;
+using VRage.Utils;
 using VRageMath;
 using VRageRender;
 
@@ -539,6 +540,8 @@ namespace Sandbox.Game.GameSystems
                 }
             }
 
+            var consumerList = new List<IMyOxygenConsumer>();
+            var producerList = new List<IMyOxygenProducer>();
             foreach (var group in groups)
             {
                 if ((group.Consumers.Count == 0 || group.Producers.Count == 0) && group.Tanks.Count == 0)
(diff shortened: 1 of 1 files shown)


JoeTheDestroyer:
I only just started working with oxygen when I got hit by this bug, so I don't have any complex setups to test with. While I'm confident I fixed the bug in question, I'm less certain that I didn't introduce any new bugs.

If anyone can try testing this on their own oxygen setups, that help a lot.

JoeTheDestroyer:
Accidentally closed. Reopening.

JoeTheDestroyer:
Accidentally closed. Reopening.

JoeTheDestroyer:
Hmmm, I noticed another side effect. Due to the greedy consumption and randomized processing ordering, in an underserved (less production than desired consumption) system the air vents can oscillate between being supplied and not fairly rapidly (every 100 updates). This causes the visual air movement effect to appear and disappear as well.

Whether this should be considered a bug or not, I don't know. In the original code this doesn't happen because, in the underserved case, what production there is is spread evenly across all vents, thus all show the effect. On the other hand, because the effect doesn't vary with the amount of oxygen moved, this gives the impression that the system is working fully. The "puffs of air" that the proposed patch causes do give a visual indication that the system is struggling to meet demand.

JoeTheDestroyer:
I have added an alternate solution #327 that doesn't have the shortcomings discussed above. However, I still believe this original solution is the better as it is both less complicated and makes fewer assumptions on the functioning of producers and consumers.

Issue #213, opened 2015-06-02
MySmallGatlingGun is not DS safe in ammo consumption


joemorin73:
Ammo consumed is proportional to the amount of users online.  MySmallGatlingGun needs to use to implement SyncObject and only let the server fire.

PR #324, opened 2015-07-03
Ammo consumption fix for MySmallGatlingGun

joemorin73:
Fix for https://github.com/KeenSoftwareHouse/SpaceEngineers/issues/213

This corrects the ammo consumption in multiplayer environment by adding IsServer check.

### DIFF
--- a/Sources/Sandbox.Game/Game/Weapons/Guns/MySmallGatlingGun.cs
+++ b/Sources/Sandbox.Game/Game/Weapons/Guns/MySmallGatlingGun.cs
@@ -406,7 +406,7 @@ namespace Sandbox.Game.Weapons
 
             m_gunBase.Shoot(Parent.Physics.LinearVelocity);
 
-            if (!MySession.Static.CreativeMode)
+            if (Sync.IsServer && !MySession.Static.CreativeMode)
             {
                 m_gunBase.ConsumeAmmo();        
             }


zrisher:
Huge, huge issue in Multiplayer. I would really like to see this incorporated.

Tyrsis:
I've added about 3 reports and even videos showing this problem in closed testing.  And completely forgot about it once source was released, so great that you found it.  It's very obvious that ammo is just being double dipped now that I look at it.  Simple fix for a very annoying problem.

joemorin73:
It's one of those things that's obvious once you see it.  I only figured it out once I saw a correlation between ammo consumption and users.

zrisher:
Perhaps instead of adding this check to `MySmallGatlingGun.Shoot()` and `MyLargeTurretBase.RemoveAmmoPerShot()`, you should add it to `Sandbox.Game.Weapons.MyGunBase.ConsumeAmmo()` instead, which both functions call to consume the ammo.

image

That would allow you to:
* Also fix this bug for handheld weapons
* Unify how this check is done by removing the duplicate `Sync.IsServer` check in `MySmallMissileLauncher.Shoot`
* Ensure this fix is not dropped later when more weapons base types are added

joemorin73:
@#$%, you're right.  I was too busy focusing on the replicating MySmallMissileLauncher's code that was working that I didn't think further.  I'll be putting in a new pull request.

joemorin73:
Closing in favour of other PR

PR #325, opened 2015-07-03
Turret ammo fix for DS/MP

joemorin73:
Similar to issue https://github.com/KeenSoftwareHouse/SpaceEngineers/issues/213

This corrects the ammo consumption in multiplayer environment by adding IsServer check.

### DIFF
--- a/Sources/Sandbox.Game/Game/Weapons/Guns/MyLargeTurretBase.cs
+++ b/Sources/Sandbox.Game/Game/Weapons/Guns/MyLargeTurretBase.cs
@@ -2049,7 +2049,7 @@ namespace Sandbox.Game.Weapons
 
         public void RemoveAmmoPerShot()
         {
-            if ((Sync.IsServer && !MySession.Static.CreativeMode)
+            if ((Sync.IsServer && !MySession.Static.CreativeMode))
             {
                 m_gunBase.ConsumeAmmo();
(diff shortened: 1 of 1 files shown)


zrisher:
Big issue in Multiplayer. I would really like to see this incorporated.

joemorin73:
@mexmer, oops.  You're right.

joemorin73:
Removed the extra brackets.

Mrjinx453:
+1 Would love to see this fixed

mexmer:
i would like to see numbers, also impact on game response.
consider people using weapon groups, firing like 20 gatlings at once.

while i understand nature of the fix, i'm not sure, if this will not create some adverse effects (lagspikes because of tons of sync packets).

joemorin73:
I'm not sure there is traffic being generated from this scenario.  I'm adding a check for the game to validate the server is removing ammo.  If anything, it should be reducing traffic as all other machines won't be removing ammo and sending out messages about ammo removal to the other machines.

mexmer:
that's a good point, i didn't thought about that.

joemorin73:
It's an interesting thing to think about.  Is it possible there are other points in the code that are generating unnecessary traffic?

joemorin73:
Closing in favour of other PR

PR #327, opened 2015-07-04
 Fix oxygen loss when room has multiple air vents (alternate solution)

JoeTheDestroyer:
Problem: Significant oxygen loss can occur when multiple air vents are used to pressurize or depressurize a room.
http://forum.keenswh.com/threads/01-077-oxygen-loss-when-depressurize-a-room.7357431/

Cause: The current oxygen production/consumption algorithm queries air vents (all consumers/producers) without regard to which room they are attached to. Additionally, each air vent reports how much air it can use without regard to other vents attached to the same room. Finally, the possibility that an air vents requested oxygen amount could change as other vents are satisfied is neglected.

This leads to a problem when a room is near full (i.e. needs less than the air vents could supply in an update). If a room has X air vents and needs Y oxygen to be full, each of the air vents will request the full Y oxygen. A total of X\*Y oxygen is extracted from producers/tanks. When the fill process begins, the first air vent supplies Y oxygen to the room filling it. The remaining (X-1) air vents try to supply Y oxygen, but it is discarded in MyAirVent.Consume() because the room is already full, wasting (X-1)\*Y oxygen.

A similar event can happen when depressurizing (leading to the creation of oxygen).

Proposed solution: I have added the concept of a shared space (IMyOxygenShared) space from which producers pull and consumers add oxygen. I then reworked the oxygen consumption algorithm to sort consumers/producers into groups according what space they shared (if any), in addition to the pre-existing sorting by priority. Then it is ensured that producers never exceed the supply of and consumers never exceed the capacity of the space they share. Finally, if there is a production shortfall, the available supply is spread proportionally across all consumers in the priority level where supply has run out.

Although in some cases the structure of the code has be refactored (for DRY's sake), care has been taken to maintain the logic of the original algorithm wherever possible.

Testing performed: I have tested in a special creative world with 3 different room sizes from a single block to 7^3 blocks and with 1-10 air vents (where possible). With the patch in place, oxygen can be cycled between the room and an oxygen tank without noticeable loss/creation (to 6 significant digits). In 1.088.012, the same setups show variation up and down of the oxygen volume, with only the first significant digit remaining constant. Various basic operations were tested on a single room (various combinations of filling/emptying rooms/tanks/generators).

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyAirVent.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyAirVent.cs
@@ -486,6 +486,11 @@ namespace Sandbox.Game.Entities.Blocks
             }
         }
 
+        IMyOxygenSharedSpace IMyOxygenConsumer.GetSharedSpace()
+        {
+            return GetOxygenBlock().Room;
+        }
+
         int IMyOxygenProducer.GetPriority()
         {
             return 0;
@@ -556,6 +561,10 @@ namespace Sandbox.Game.Entities.Blocks
                 Debug.Assert(MyOxygenProviderSystem.GetOxygenInPoint(WorldMatrix.Translation) > 0f);
                 m_producedSinceLastUpdate = true;
             }
+        }
+        IMyOxygenSharedSpace IMyOxygenProducer.GetSharedSpace()
+        {
+            return GetOxygenBlock().Room;
         }
         #endregion
 
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyCockpit.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyCockpit.cs
@@ -1312,5 +1312,10 @@ namespace Sandbox.Game.Entities
                 m_oxygenLevel = 1f;
             }
         }
+
+        IMyOxygenSharedSpace IMyOxygenConsumer.GetSharedSpace()
+        {
+            return null;
+        }
     }
 }
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyOxygenFarm.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyOxygenFarm.cs
@@ -204,6 +204,11 @@ namespace Sandbox.Game.Entities.Blocks
             return 1;
         }
 
+        IMyOxygenSharedSpace IMyOxygenProducer.GetSharedSpace()
+        {
+            return null;
+        }
+
         bool IMyOxygenBlock.IsWorking()
         {
             return MySession.Static.Settings.EnableOxygen && PowerReceiver.IsPowered && IsWorking && IsFunctional;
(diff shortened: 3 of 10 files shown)


JoeTheDestroyer:
It should be noted that this pull request is an alternate solution to that presented in #322 . This solution attempts to more closely follow the logic of the original code, at the cost of being significantly more intrusive and complicated.

krypt-lynx:
I think IMyOxygenConsumer must to do push request if than can't consume oxygen, not just store it in nowhere.

JoeTheDestroyer:
It would be nice if it were that simple. Unfortunately, the oxygen system does not use push/pull requests (e.g., air vents have no inventory). In fact, the only way the oxygen system uses the conveyor system is to check if two blocks are connected, nothing else:
https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/master/Sources/Sandbox.Game/Game/GameSystems/MyGridOxygenSystem.cs#L529

So, the only solution is to make sure IMyOxygenConsumer is never given more oxygen than it can consume. (And never ask IMyOxygenProducer to give more oxygen than it can produce.)

krypt-lynx:
So we need to implement push/pull requests for oxygen system. And only when fix this issue :D

krypt-lynx:
@mexmer, do you have any plans to add oxygen system push/pull requests? Of course if you can share with information.

mexmer:
why you ask me?
i have not even started tampering with oxygen system unlike @JoeTheDestroyer 

if you wanted to ask developers, you need to direct your question to @OndrejPetrzilka or @No-Brain

krypt-lynx:
Sorry. Seems like can edit pull request's, so I supposed that you is an SE developer

mexmer:
everybody can edit own PRs and issues

JoeTheDestroyer:
IMHO, it's not likely. The dev's had the option of making more use of the conveyor system when they implemented oxygen in the first place, but they chose to make it separate.

Changing it now would require a significant rewrite, converting oxygen to an inventory item and adding inventories to all oxygen related blocks.

krypt-lynx:
> would require a significant rewrite
Not necessary. Yes, where work for day or two, but it is not full subsystem rewrite.

As option we can try to implement oxygen specific pull/push requests just inside MyOxygenProviderSystem

I can try to do it. But in fact, I'm too lazy :D

JoeTheDestroyer:
> Yes, there is work for day or two
Which is what I call significant. The solutions I've posted only took about an hour. Two days is more time than I'm willing to give away for free on a fix that's not significantly better than what I already have.

> but it is not full subsystem's rewrite.
The only part you could keep would be the part that checks whether rooms are sealed. The rest would have to be rewritten.

> As option we can try to implement oxygen specific pull/push requests just inside MyOxygenProviderSystem
In some sense, the system already works that way. It's just that the "pull" amount is not always honored by the resolution algorithm (or rather, the algorithm ignores the fact that the "pull" amount can change as other "pull" requests are satisfied).

krypt-lynx:
> In some sense, the system already works that way. 
Lack of time for proper implementation?

PR #328, opened 2015-07-05
FIX: game ui transparency settings reset

mexmer:
problem:
transparency slider is reset to default, if user doesn't do any change to it.


cubrr:
I noticed this bug - I hadn't played after the transparent UI update was rolled out and the sliders were all the way to the left when I opened the relevant options menu for the first time. Moving the bottom slider didn't do anything until I touched the top slider.

mexmer:
rebased to current source.

PR #332, opened 2015-07-06
FIX: projector detailed info update

mexmer:
ISSUE: when adding block that is not derived from terminal block (eg. armor, catwalk, and so on), detailed info is not updated, unless you exit and enter terminal, this affects also programmable block.
ISSUE2: when projector image is moved, detailed info is not updated on other clients after new offset is set, only projection is updated (in case of overlaying blocks, "welded" blocks are dedicted from projection, but text in projector stays unchanged)


mexmer:
rebased and fixed
bug is still present even in single player

mexmer:
fixed in release

PR #337, opened 2015-07-08
Gesture radial Menu

Eikester:
Allows to quickly select a Gesture from a radial Menu similar to the
Radio/Weapon selection in GTA V or other Games

Press and Hold Middle Mouse Button to show the Menu, hover over one of
the Gestures and release the Middle Mouse Button to play the selected
Animation

Screenshots:
http://steamcommunity.com/sharedfiles/filedetails/?id=477568662
http://steamcommunity.com/sharedfiles/filedetails/?id=477568697

### DIFF
--- a/Sources/Sandbox.Game/Game/Screens/MyGuiScreenGamePlay.cs
+++ b/Sources/Sandbox.Game/Game/Screens/MyGuiScreenGamePlay.cs
@@ -399,6 +399,15 @@ namespace Sandbox.Game.Gui
 
                     if (context == MySpaceBindingCreator.CX_CHARACTER || context == MySpaceBindingCreator.CX_SPACESHIP)
                     {
+                        if (MyToolbarComponent.CurrentToolbar.SelectedSlot == null && MyInput.Static.IsMiddleMousePressed())
+                        {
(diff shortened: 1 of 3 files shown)


CalJacobson:
Really like this idea.  Always felt that adding animations to the toolbar seemed awkward.

This would be very interesting if it could be contextual as well...say, if you're near a button panel then any of the set buttons could be selected without looking directly at the panel itself.  This would be the equivalent of having your "finger on the button", ready to press it, while allowing your attention to be focused elsewhere.

FavioGalvis:
+1 :+1:

mccorkle:
+1 for me too. I never use emotes in game, as I tend to voip with my teammates instead, but this could be huge for machinima in the game engine.

Spartan322:
+1

joemorin73:
+1.  Very awesome looking.

lord-devious:
Cool, but isn't the middle mouse button used for painting? That could quickly become a problem...

Eikester:
current implementation (which isn't perfect tbh) checks if the user has something (i.e. Block) selected (radial Menu can only be opened if the current selected Toolbar Slot is empty)

lord-devious:
:+1:

PR #338, opened 2015-07-08
Exposed IMyShipController inputs

plaYer2k:
The inputs of IMyShipController blocks can now be used with the ingame
programming API.
This allows for building your own ingame driver for thruster, gravity drives, robot arm remote controls, general 6-axis value inputs (i.e. slider for color/rotor velocity), custom build wheel and many more.

When a structure does not have power anymore, the old values will maintain.
The values however get updated even though the cockpit in question is not the main cockpit, ship controlls are disabled, there is no thruster or gyro grid system present etc.

Thus it can be used for pretty much everything as translation and rotation input from a cockpit.

Due to the "out of memory" issues with the source build i could not extensively test it but the normal KSH ships from the Easy Start worlds aswell as a few of my own ships behaved like before.

new properties:
- IMyShipController.Translation
- IMyShipController.Rotation

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyShipController.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyShipController.cs
@@ -15,5 +15,13 @@ namespace Sandbox.ModAPI.Ingame
         bool ControlThrusters { get; }
         bool HandBrake { get; }
         bool DampenersOverride { get; }
+        /// <summary>
+        /// Returns the current translational/movement input from this controller relative to itself.
+        /// </summary>
+        VRageMath.Vector3 Translation { get; }
+        /// <summary>
+        /// Returns the current rotational input from this controller relative to itself while taking the roll dampener into account.
+        /// </summary>
+        VRageMath.Vector3 Rotation { get; }
     }
 }
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyShipController.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyShipController.cs
@@ -83,6 +83,8 @@ namespace Sandbox.Game.Entities
 
         bool m_forcedFPS;
 
+        Vector3 m_localTranslation, m_localRotation;
+
         //        MyGunTypeEnum? m_selectedGunType;
         MyDefinitionId? m_selectedGunId;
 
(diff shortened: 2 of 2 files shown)


THDigi:
This looks great, simple and really effective.

JustBurn:
This is a nice and simple way to get analog controls... I'm curious to know how this one will play in MP :sweat_smile:

THDigi:
This needs an update :}

plaYer2k:
It sure does @THDigi but until the planets have been merged into the git and it is kept up to date i wont spend time any energy updating.
Especially because we git contributors here havent gotten any love whatsoever laterly for a few months now (with a very few exceptions).

So until keen gets active again, me and apparently almost every other contributor wont update the pull requests.

JustBurn:
I Did some tests with your PR and found out as the way it work right now it's pretty unpractical.
It only modify values when there's a change in the input in exception when the user release all axis, in this case the value doesn't change and remain the old value.

Translation: X:0.0; Y:0.0; Z:0.0
(User press W, 1 axis is being used)
Translation: X:0.0; Y:0.0; Z:-1.0
(User release W, no axis are being used)
Translation: X:0.0; Y:0.0; Z:-1.0 <- should be X:0.0; Y:0.0; Z:0.0
(User press D, 1 axis is being used)
Translation: X:1.0; Y:0.0; Z:0.0

You'll need to clear m_localTranslation and m_localRotation in ClearMovementControl() so this works as it should.

Here's a patch which works for me (I've tested it):

```cs
        private void ClearMovementControl()
        {
            if (!m_enableShipControl)
                return;

+            m_localTranslation = Vector3.Zero;
+            m_localRotation = Vector3.Zero;

            if (GridThrustSystem != null)
            {
                if (GridThrustSystem.ControlThrust != Vector3.Zero ||
```

I didn't tested this PR in MP myself but by looking at the code apparently shouldn't work online (probably except for the host), there some events in MySyncGrid but those deal with the thrusters and gyros directly.
If this is the case you'll need to syncronize m_localTranslation and m_localRotation using the Message system in MySyncShipController and figure out a way to only transmit when it's requested.

plaYer2k:
Thanks for checking.

Though that is odd because the tests i made never showed that issue aswell as the video i still got from 4 months ago.
Everything was performing right there. The values went to 1 when pressed and to 0 when released.

Thought i will redo this pull request once the source updated anyway.
I just didnt close it for the sake of discussions like that.

As for the syncing, it should be okay to ignore it imo as the players inputs are synchronized already.
Otherwise all ship movements would be async aswell.
But i couldnt do tests back then.

PR #339, opened 2015-07-09
Jump drive interface

THDigi:
Populated IMyJumpDrive with various fields and methods to read and control jump drives.


FavioGalvis:
:+1:

PR #340, opened 2015-07-09
FIX: IMyTextPanel.WritePrivateText

mexmer:
- due typo in code, variable was reset prior sending sync message, and text was reset, when using pb.


THDigi:
Hmm, PrivateTitle is also in need of this fix then.

mexmer:
@THDigi i see, you are right. will fix that later.
i don't use private text at all, just checked for this, because of report on forums.

mexmer:
rebased and added fix for title

mexmer:
could we get this merged soon, so we can finally use privatetext/privatetitle and not only publictext/publictitle?

LeonidasCZ:
merged with #394 thanks

mexmer:
yes, thanks

PR #341, opened 2015-07-10
Prevent cluster split when object moves too far

### LINKED ISSUE #158: Havok triggers entity removal in a limited world incorrectly
With a limited world size, if I change the position in ModAPI scripting on an entity to near the border (but still 10+km away from border), then havok triggers the HavokWorld_EntityLeftWorld event and deletes the entity.

This causes the entity to disappear, and the player left floating, however there's no HUD warning that you are near the border. You can still move away from the center much further before reaching the real end.

I'm attaching the world and my mod, which trigger it.
To reproduce, simply load the world with the mod, enter the cockpit, and press 1.

Phoenix84:
This prevents a cluster split if the distance is greater than the ideal cluster size.
Fixes #158.

I recommend a lot of testing with this fix. I'm not sure it's the ideal fix, but it works.

### DIFF
--- a/Sources/VRage.Math/Spatial/MyClusterTree.cs
+++ b/Sources/VRage.Math/Spatial/MyClusterTree.cs
@@ -653,10 +653,16 @@ namespace VRageMath.Spatial
                         {
 
                             System.Diagnostics.Debug.Assert(aabb0.Max.GetDim(longestAxis) - minimumCluster.Min.GetDim(longestAxis) > 0, "Invalid minimal cluster");
-                            isClusterSplittable = true;
-
-                            currentMax.SetDim(longestAxis, aabb0.Max.GetDim(longestAxis));
-
+                            if (aabb0.Max.GetDim(longestAxis) - minimumCluster.Min.GetDim(longestAxis) > IdealClusterSize.GetDim(longestAxis))
+                            {
+                                // Distance is further than a cluster size, IOW object was teleported (eg jump drive)
+                                isClusterSplittable = false;
+                            }
+                            else
+                            {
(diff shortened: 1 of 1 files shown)


jb-aero:
I don't suppose this is related to jumped ships disappearing?

Phoenix84:
It is.

zrisher:
I know for anyone who's been following this or #158 this is obvious, but just to reiterate this is intended to fix a large, very annoying bug with jumping in limited worlds that's been all over the servers since jump drive was introduced, as reported [on the forums](http://forum.keenswh.com/threads/1-090-jump-drive-loosing-entire-ship-and-view-left-distorted.7364254/page-2#post-1286801399). I hear from about 10 people a day that they lost a ship to this bug, even though we post info all over not to use the drive till it's fixed. KSH, please review and incorporate when you can.

Phoenix84:
I just want to make sure everyone is clear that this is a kludge, not a proper fix.
I think this bug is bad enough that if Keen has trouble with this one (and having scoured deeply in the code myself, I understand), then I recommend this only as a temporary measure.
I know, there's nothing like a permanent temporary fix. :wink: 

There are probably side effects to this fix, I just don't know what they are.
I didn't step through all the code to completion after the cluster split is prevented. I did some quick tests and was able to keep jumping for some time, and cluster reorg still happened elsewhere. So at least I'm confident it didn't break it in an obvious way.

In fact I'd probably put an assert in there too, to help catch that condition in the future.

As was recommended by @Tyrsis, the proper fix probably would be to create a new cluster every time a jump happens, like with how respawn ships are placed into the world.

I just don't know how to do that.

PR #342, opened 2015-07-11
Allow jump drive to jump floating objects and unseated players

Phoenix84:
This change allows all floating objects (ore, ingots, etc), as well as floating objects to be jumped with the ship.

It checks the bounding region of each attached grid and grabs everything or everyone inside.

Currently the jump positions are cached before countdown, so the last 10s of movement will be lost.

### DIFF
--- a/Sources/Sandbox.Game/Game/GameSystems/MyGridJumpDriveSystem.cs
+++ b/Sources/Sandbox.Game/Game/GameSystems/MyGridJumpDriveSystem.cs
@@ -30,7 +30,7 @@ namespace Sandbox.Game.GameSystems
         
         private HashSet<MyJumpDrive> m_jumpDrives = new HashSet<MyJumpDrive>();
         private HashSet<MyCubeGrid> m_connectedGrids = new HashSet<MyCubeGrid>();
-        private Dictionary<MyCubeGrid, Vector3D> m_shipInfo = new Dictionary<MyCubeGrid, Vector3D>();
+        private Dictionary<MyEntity, Vector3D> m_shipInfo = new Dictionary<MyEntity, Vector3D>();
 
         private List<MyEntity> m_entitiesInRange = new List<MyEntity>();
         private List<MyObjectSeed> m_objectsInRange = new List<MyObjectSeed>();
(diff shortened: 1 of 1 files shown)


Phoenix84:
I just realized that removing the cache is a exploit of the ship is moving while jumping.
This would allow a ship to start moving and break the 2km distance limit during the 10s countdown.

I can pull that commit out, I just didn't want player movement ignored during countdown.
Any recommendations?

JJRcop:
Are the floating objects counted for mass? If more objects enter during countdown and they cause the mass to become over the limit, what happens?

Phoenix84:
Mass of floating objects is included.
If objects wander into jump range, their mass will be included when the power is depleted.
The actual jump distance will also be reduced at jump time, although there will be no indication it happened.

Geneticus:
Yes!  One thing I always felt was "off" with the bounding box method was that a ridiculously large ship could have its pinky toe slightly in the box and jump with the much smaller vessel. Maybe a check to see if at least 80% of a (guest)grid is within the bounding box of the jumping vessel or it is ignored.

Phoenix84:
@Geneticus, while I agree that would be better, I'm not sure how that would affect performance. The changes I've made so far don't add many additional calculations.
I guess at least with the source code, I can profile it.
I'd probably do that in another PR though, once this is accepted.

zorg-kirill:
wow great job!

Phoenix84:
This is lame. The changes aren't going to merge cleanly, they changed that area of code as much as I have, and there are some incompatible changes.
I might have to go back to the drawing board with this one. :frowning:

paradineai:
A quick question. What about checking to see if the the outside objects (anything not part of the jumping ship) has a bounding box that is larger than the ship that is jumping? If it does then it is not included in the jump unless more than lets say 60% of it is inside the jumping ships bounding box. This would make it so that if for example you are preforming a jump and another ship comes close to join in the jump or just happens to get too close then all of the mass is added up and it would decrease the overall range of the jump or if it is over the limit then it would be able to post a message that says the mass is larger than the maximum and that you can only perform a minimum distance jump. This would require a full charge and use every drop of it. Also You would then be able to add in the ability to have the 2 ships power the jump drives and have them work together to increase their range. In doing this you could even add an assist or auxiliary jump function to the jump drives to allow you to use the jump drive of a larger ship to jump a smaller ship or assist the smaller ships jump drive to send it further away (like a jump gate would function). This would also allow a station to have a jump drive on it setup so it has a check box to select assist mode and a second check box for selecting auxiliary mode. By selecting assist mode it will assist any ships jump drives that are in range and by selecting auxiliary mode it would allow you to send a ship in the direction it is facing using the stations jump drive. (This would become very usable when planets are released to launch ships to space from the surface.)

PR #343, opened 2015-07-11
Symmetry Plane Transparency

aaraujo666:
This change add options under advanced world settings to change the alpha of the symmetry planes and also allows you to turn the symmetry planes on and off.


plaYer2k:
Why is that within the world settings and not within the game settings?
The symmetry scale should be adjustable per player and not per world in my opinion.

MeusRex:
Wait a second, wasn't that already done?

plaYer2k:
It was submitted and closed apparently (though i cant see the code there)

https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/263

lord-devious:
I have to agree, this needs to be an in-game option.

THDigi:
Or as someone suggested on the forums, when in the symmetry mode (M) use the mouse wheel to adjust the transparency, would be way easier to configure on the fly.

PR #344, opened 2015-07-12
Exposed IMyCubeGrid Mass and CenterOfMass

plaYer2k:
Exposed CenterOfMass in world coordinates and Mass of a cube grid.

This is especially useful with planets where mass and the point where the resulting weight under gravity affects the grid.

Furthermore it is useful for several other applications like custom dampener, levitating mechanisms etc.

(Ignore the multitude of comits and revers. That is me and the git tool not liking each other.)


JoeTheDestroyer:
How about moment of inertia, too?

There's a note here about only the inverse being correct:
https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/master/Sources/Sandbox.Game/Game/GameSystems/MyGridGyroSystem.cs#L180
Haven't looked any further into it, though, so I can't say why.

Of course, velocity/angular velocity would be nice too, but I expect that's a harder sell...

mexmer:
float should not return null.

plaYer2k:
@JoeTheDestroyer 
Right, the moment of inertia could be exposed aswell. I will fix that.
I initially wasnt sure about the velocities, but as they are present and the method to calculate them youself is rather simple, i see no reason not to expose them aswell to reduce the redundant calculations to manually do it youself.


@mexmer 
It returns null when there is no Physics. This should not happen but can.
As i am no friend of throwing an exception when some internal error happens, i prefer to give the user a simple null check over a costy try-catch construct.

Alternatively i could return a zero-vector and zero-mass, which makes it hard to distringuish between an errorous and working state.
(Sure it works for mass as there sure is no block that has 0kg mass due to them always requiring components, but it still aint elegant to return 0f)

So when you dont like this implementation, what is your suggestion for an improvement?

mexmer:
for vector it makes sense if it returns null, since vector is struct, and 0,0,0 vector position is center for coords, therefore valid value.
float is number for christ sake, you don't check float == null, you use float as given value.
i'm quite oposed to using nullable for any numeric variable when strict type is used.
returning  0 mass when physics is null is correct, because object without physics has zero mass (is virtual), but object with physics has non zero mass.

JoeTheDestroyer:
An object without physics is immovable, right? So wouldn't Single.PositiveInfinity be the correct thing to return (i.e., a=F/m, no matter what force is applied, acceleration is always zero)?

Zero is easier to check, though.

mexmer:
infinity would work also, just not NULL.
object without physics is for example also projection.
while there is currently no api to access grid from projector ingame, it might be later.

THDigi:
Or expose an in-game version of grid.Physics field that can be null if the object has no physics ?

mexmer:
i would rather not expose whole physics to ingame api.

THDigi:
Not the whole one, a special in-game scripting interface which can contain the physical related stuff that can and should be shared with PBs.

plaYer2k:
@mexmer 
See, while you are against using null for things that simply are not defined, i am against using real values for things that are not defined.
Giving any real value like 0 for a mass that is not defined at all is highly against anything i could accept.

@JoeTheDestroyer 
Well, not having physics doesnt mean that it is immoveable but more that it has no "real" or "physical" representation yet.

@mexmer @JoeTheDestroyer 
So while mass=0 might be easy to check etc, checking if float? is null is also easy to check and very clearly distinguishes between having a real grid or not.
So i personally couldnt accept returning a real value like 0 for that purpose.

float.NaN however is something i could accept. Not sure about Vector3.NaN though as it feels kinda alien aswell.
I actually am of the opinion that either returning null for it not being present or a real value is the best method.

Like mexmer pointed out there, projectors do that but also blocks/copies that are about to be placed.
We dont have ingame programming access to them now, but it makes sense to access a few of them.
We also dont know yet what the future will bring and which features come that may expose "non-real" grids.
Thus the current code has to be prepared for that.

@THDigi 
I like the idea of such a data type like IMyPhysics with properties for velocity, mass etc though i dont know how high the performance impact due to redundant and permanent checks and calculations.

Anyhow, i made a poll so people can vote for their favorite and we see what turns out to be the favored implementation for now.
http://strawpoll.me/4945413

mexmer:
@THDigi that might do the work, but if it's only for 2 values, it's just unnecessary.

@plaYer2k i'm not against using null, i'm against using null on basic strict types, like numeric types(either it's integer or float), or bools. For objects and structures, it's fine.
as for game defining interface, interface will not need to calculate anything, it will use already existing and computed values, there should be no impact on perfomance, unless you write it badly.

plaYer2k:
Alright i changed the return values for invalid bases from nullable types to
normal types with NaN and Zero based on the votes result.

Also added the linear/angular velocity/acceleration and tensors for the
moment of inertia.

The list of exposed informations now consists of:
```
        /// <summary>
        /// Get the center of mass position in world coordinates
        /// </summary>
        VRageMath.Vector3D CenterOfMassWorld { get; }
        /// <summary>
        /// Get the mass in kg
        /// </summary>
        float Mass { get; }
        /// <summary>
        /// The grids linear movement velocity.
        /// </summary>
        VRageMath.Vector3 LinearVelocity { get; }
        /// <summary>
        /// The grids angular movement velocity.
        /// </summary>
        VRageMath.Vector3 AngularVelocity { get; }
        /// <summary>
        /// The grids linear movement acceleration.
        /// </summary>
        VRageMath.Vector3 LinearAcceleration { get; }
        /// <summary>
        /// The grids angular movement acceleration.
        /// </summary>
        VRageMath.Vector3 AngularAcceleration { get; }
        /// <summary>
        /// The grids Moment of Inertia tensor.
        /// </summary>
        VRageMath.Matrix InertiaTensor { get; }
        /// <summary>
        /// The grids inverted Moment of Inertia tensor.
        /// </summary>
        VRageMath.Matrix InverseInertiaTensor { get; }
```

PR #349, opened 2015-07-15
Bugfix Programmable Block: Generics (List<MyCustomClass>), Accessibility, duplicate type in assembly

lord-devious:
Since the whitelist/instruction counter safety mechanism copied all nested types into the top level, it was very easy to end up with duplicate type names. This was specifically apparent when using the } exploit to create extension functions, and then using generic methods. This PR changes the type duplication method to retain the type hierarchy, so this exception is avoided. This should also hopefully fix the public/private accessibility issues, which was also caused by not rebuilding nesting properly.

Also, member replacement has been revamped to properly map generic method calls so that generic lists can be used.

Example of scripts that will not work before this fix is merged:
First:
```cs
void Main(string argument) {
    var list = new List<X>();
    list.Add(new X());
    X x = list[0];
}

class X {
}
```
Second:
```cs
void Main(string argument)
{
    C.Gen1<int>(this);
    C.Gen1<long>(this); 
    C.Gen1<X>(this);  
}

class X {
}

class C {
    public static void Gen1<T>(Program program) {
        program.Echo("Hello " + typeof(T).Name);
    }
}
```


lord-devious:
Hang on, found something else related to this...

lord-devious:
Yep, that should do it.

lord-devious:
Any tests using this branch on your most complicated scripts will be very appreciated.

lord-devious:
HOLD THIS ONE still. I'm onto something.

lord-devious:
I think I may have fixed the generic List<> problem. However since this is complicated, I would be very thankful for those of you who could find it in you to attempt to break this. 

Please see if you can find a piece of code that works in the public version that breaks with this branch. There's bound to be issues.

lord-devious:
A big thank you to @JoeTheDestroyer for testing this code for me and finding issues!

lord-devious:
This request should now be safe to merge.

stanhebben:
Does this also fix defining generic functions?

I tried something like this yesterday and it compiled but didn't run:

```C#
Main(string argument)
{
    var textPanel = FindBlock<IMyTextPanel>("[" + argument + "]");
    /* ... */
}

T FindBlock<T>(string tag) where T : class
{
    var blocks = new List<IMyTerminalBlock>();
    GridTerminalSystem.GetBlocksOfType<T>(blocks);
    for (int i = 0; i < blocks.Count; i++) {
        if (blocks[i].CustomName.Contains(tag))
            return blocks[i];
    }
    
    return null;
}
```

lord-devious:
@stanhebben Yes, it does. But your code snippet helped me find a bug in my code, so thank you very much.

lord-devious:
Hang on, fixing bug...

lord-devious:
Problem solved.

lord-devious:
By the way, @stanhebben, for a workaround until this PR is merged, I think it's enough for you to make sure all members are public. I.e. use public void main, public T FindBlock etc.

PR #353, opened 2015-07-16
Grammer

aaronfranke:
Fixed grammar in a quote

### DIFF
--- a/Sources/SpaceEngineers/Content/Data/Localization/MyTexts.resx
+++ b/Sources/SpaceEngineers/Content/Data/Localization/MyTexts.resx
@@ -1634,7 +1634,7 @@ See log for details.</value>
     <value>Newton's law of universal gravitation</value>
   </data>
   <data name="Quote69Text" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


zrisher:
"Point mass (pointlike mass) is the concept of either matter that is infinitely small, or an object which can be thought of as infinitely small." 
https://en.wikipedia.org/wiki/Point_particle#Point_mass

aaronfranke:
Oh... neat, I wasn't aware that it was actually a proper term. However, I'm not sure that "point mass" is the appropriate term to use here anyway, since Wikipedia says "When analyzing the gravitational forces in a system, it becomes impossible to account for every unit of mass individually. However, a spherically symmetric body affects external objects gravitationally as if all of its mass were concentrated at its center". This makes me think that "point mass" is a term for estimating, rather than an absolute thing in physics?

Anyway, if point mass is kept, at least add a note or something so people reviewing this quote know that this is actually the proper term, and no a grammar error.

aaronfranke:
Please archive this PR, thanks.

PR #354, opened 2015-07-16
FIX: missing mouse boundary check on click in UI control

mexmer:
gist of the fix is, that any control that is derived from control parent (which seems to be every grouped control, like tab, list, listbox, grid, and so on), has aditional check for mouse position, when mouse button is pressed (and only when mouse button is presed). if mouse is inside control area, it will allow to forward input to child controls, if it's not, it will exit from input control handling.

please note, this check is already present, on some controls (like scrollable list), but grouped controls (like tab or grid), are missing this check.

due this missing check, if you have for example grid, with lot of controls inside, that needs to be scrolled, you can click on any control of the grid (even outside "visible area")

this should fix problem with unclickable tab controls in terminal, when you scroll down for example inventory list (reported on forums)

please note i have no controller, but might be, that for controller it will need to add aditional check.

### DIFF
--- a/Sources/Sandbox.Graphics/Gui/MyGuiControlParent.cs
+++ b/Sources/Sandbox.Graphics/Gui/MyGuiControlParent.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Text;
 using VRage.Utils;
 using VRageMath;
+using VRage.Input;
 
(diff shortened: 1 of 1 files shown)


mexmer:
fixed in 1.096

PR #356, opened 2015-07-17
New ray function to ease the use and increase the performance

LynnuxX:
Added IMyEntities.ClosestRayIntersection a modified method of IMyEntities.IsRayCastBlocked which is easier to use in many applications - just one call in most cases - thereby increasing the performance.
Returns null if no intersection otherwise the location (Vector3D) of the
closest intersection of the ray.
Reduces Gamelogic time by up to 2ms when used by my rangefinder mod
compared to the current solution using several raycast methods on lotsa
different objects.

Change the name as you see fit.
First contribution on Github with problems, therefore the two reverts...


LynnuxX:
Read something about create a BRANCH and commit changes there.
Fail ! => closed
Sorry.

PR #360, opened 2015-07-18
Inventory Dupe

Aleks976:
When dragging and dropping an item in your inventory with a volume equal
to the space left your inventory, your character drops the item but it
stays in your inventory.

Example: Your inventory volume is 399/400. You have one computer (volume = 1) in one inventory slot. When you drag that one computer to other slots nothing in your inventory will change, but your character will drop one computer each time you try to drag the item to other slots.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Inventory/MyInventory.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Inventory/MyInventory.cs
@@ -193,7 +193,7 @@ namespace Sandbox.Game
             var adapter = MyInventoryItemAdapter.Static;
             adapter.Adapt(contentId);
 
-            if (MyPerGameSettings.ConstrainInventory() && (amount * adapter.Volume + m_currentVolume > MaxVolume) || (amount * adapter.Mass + m_currentMass > m_maxMass))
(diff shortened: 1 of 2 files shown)


Sources/Sandbox.Game/Game/Entities/Inventory/MyInventory.cs
```diff
@@ -193,7 +193,7 @@ public bool CanItemsBeAdded(MyFixedPoint amount, MyDefinitionId contentId)
             var adapter = MyInventoryItemAdapter.Static;
             adapter.Adapt(contentId);
 
-            if (MyPerGameSettings.ConstrainInventory() && (amount * adapter.Volume + m_currentVolume > MaxVolume) || (amount * adapter.Mass + m_currentMass > m_maxMass))
+            if (MyPerGameSettings.ConstrainInventory() && (amount * adapter.Volume + m_currentVolume >= MaxVolume) || (amount * adapter.Mass + m_currentMass > m_maxMass))
```
leftler: The same change should likely be applied to the mass check also

Aleks976: I don't think anything has a max mass yet, but I should probably change it anyways

mexmer: yes, atm. inventory has only volume limit, but it might change in feature. so that check might be worth adding.

PR #363, opened 2015-07-20
Ore Detector: Get Ore Locations

Rynchodon:
added GetOreLocations() methods for mod/ingame script use

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyOreDetector.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyOreDetector.cs
@@ -9,5 +9,11 @@ namespace Sandbox.ModAPI.Ingame
     {
         float Range {get;}
         bool BroadcastUsingAntennas {get;}
+
+        /// <summary>
+        /// Gets the world position and name of all the ores detected by this Ore Detector.
+        /// </summary>
+        /// <returns>A Dictionary with all the world positions and names of detected ores.</returns>
+        Dictionary<VRageMath.Vector3D, string> GetOreLocations();
     }
 }
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyOreDetector.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyOreDetector.cs
@@ -183,7 +183,19 @@ namespace Sandbox.Game.Entities.Cube
                 RaisePropertiesChanged();
             }
         }
+
+        #region IMyOreDetector Members
+
         bool IMyOreDetector.BroadcastUsingAntennas { get { return m_oreDetectorComponent.BroadcastUsingAntennas; } }
         float IMyOreDetector.Range { get { return Range; } }
+
+        /// <summary>
+        /// Gets the world position and name of all the ores detected by this Ore Detector.
+        /// </summary>
+        /// <returns>A Dictionary with all the world positions and names of detected ores.</returns>
+        System.Collections.Generic.Dictionary<Vector3D, string> IMyOreDetector.GetOreLocations()
+        { return m_oreDetectorComponent.GetOreLocations(PositionComp.GetPosition()); }
+
+        #endregion
     }
 }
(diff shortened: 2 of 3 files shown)


Rynchodon:
Over millions of iterations, I have seen median execution times vary from 35 ­µs to 70 µs, which seems perfectly acceptable to me. Setting a larger initial capacity for the dictionary does save some time, so I will add that.

mexmer:
you forgetting one important thing, it's called garbage collector.
and no. initial dictionary doesn't help that. your solution is suboptimal for use on multiplayer also.

you should use informations cached by game engine upon it's update ticks, and not add aditional workload, which your update does.

Rynchodon:
>you forgetting one important thing, it's called garbage collector.

I suppose it could take a dictionary, clear it, and use that, is that what you are trying to suggest?

>and no. initial dictionary doesn't help that.

The initial capacity for the dictionary reduces the number of times the dictionary needs to be resized.

>your solution is suboptimal for use on multiplayer also.

Can you explain this?

>you should use informations cached by game engine upon it's update ticks, and not add aditional workload, which your update does.

I did it this way to avoid doing any work unless GetOreLocations() is actually invoked. Even if a dictionary is maintained on update, it would still have to be cloned every time this function is invoked.

mexmer:
Ore markers on hud are relayed to other players whenever they are created or removed (they are shown only to friendly players, but still).
Your explicit call to update increases load to client and also to server.

If you remove that explicit update, it will work same way, because ores are updated in simulation cycle anyway, and there is no aditional load by calling unnecessary update.

You should not inquire update more offten than engine does itself.

Rynchodon:
>If you remove that explicit update, it will work same way, because ores are updated in simulation cycle anyway, and there is no aditional load by calling unnecessary update.

Ores are only updated when the ship is being controlled or "Broadcast using antennas" is enabled.

Rynchodon:
Invoking Update doesn't actually work.

mexmer:
Ores are updates always.
HUD markers are forwarded only when broadcast using antenas is enabled. Which option people use on scouting drones.

What is not updated by invoking "Update" is voxel map. Look at what Update does, and what it access.
Update only updates internals of detector. Voxel map is updated by different method, but that should not be invokable by player on demand anyway.

Rynchodon:
>Ores are updates always.

`MyOreDetector` invokes `Update` in `UpdateBeforeSimulation100` but `Update` checks for control:
```C#
if (!SetRelayedRequest && checkControl && !OnCheckControl())
{
    m_depositGroupsByEntity.Clear();
    return;
}
```
For ore detectors `OnCheckControl` points to `MyOreDetector.OnCheckControl`.

>What is not updated by invoking "Update" is voxel map. Look at what Update does, and what it access.
Update updates internals of detector and hudmarkers. Voxel map is updated by different method, but that should not be invokable by player on demand anyway.

`Update` starts `MyDepositQuery`:
`Update` -> `UpdateDeposits` -> `IssueQueries` -> `MyDepositQuery.Start`

mexmer:
deposit queries are not synchronous with Update call, you issue query, but it process, when it process. eg. you don't directly invoke voxel map update.

Rynchodon:
>voxel map queries are not synchronous with Update call, you issue query, but it process, when it process. eg. you don't directly invoke voxel map update.

That's why I wrote "starts".
This is the only way queries are actually started, so `Update` needs to be invoked to get all the ores but the ores will not be up-to-date immediately after `Update` runs.

I will rewrite this so that users register for an event that is triggered after all queries complete. Might as well throw in a read-only dictionary while I'm at it.

mexmer:
you can't and should not rewrite how engine update voxels deposit, that's why i said, you should rely on whatever engine has cached, and also why i wrote, your update call is in fact unnecessary.

either you will work, what engine gives your or you don't, but you should never add aditional workload to egine.

for fyour unction,using cached information should be enough, there is nothing wrong in your function copying existing GPS material list in new list, that is available for programable block. what is wrong, is you attept to force engine to update voxel and ore cache.

PR #323, opened 2015-07-02
Thruster Flame Color changeable in Terminal

Eikester:
This is a often requested feature which allows to change the Thruster Flame Color in the Terminal (K-Menu)

Less cluttered G-Menu
Less Blocks to load (no more x copies of the same Thruster just to have a different Flame Color)
Less work for Modders

PR #364, opened 2015-07-20
Thruster manual boost

Knsgf:
This pull request does 2 things:
* First, it extends the range of "Thrust override" slider, so that TurboBoost feature -- which is restricted to braking in vanilla version -- can be used for acceleration as well,
* Secondly, it adds a new "FlameBoostColor" tag to CubeBlocks.sbc. This tag allows to specify different flame colour for thruster operating in TurboBoost mode. If omitted, the tag defaults to "FlameFullColor" value.

### DIFF
--- a/Sources/Sandbox.Game/Definitions/MyThrustDefinition.cs
+++ b/Sources/Sandbox.Game/Definitions/MyThrustDefinition.cs
@@ -19,6 +19,7 @@ namespace Sandbox.Definitions
         public float FlameLengthScale;
         public Vector4 FlameFullColor;
         public Vector4 FlameIdleColor;
+        public Vector4 FlameBoostColor;
         public string FlamePointMaterial;
         public string FlameLengthMaterial;
         public string FlameGlareMaterial;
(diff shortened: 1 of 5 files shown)


MeusRex:
I don't think the first part is necessary... If you want faster thrusters, mod them. This will just add inconsistencies and unfair behaviour as people who know about this can suddenly gain more acceleration by putting an override in their toolbar or by PB.

Spartan322:
Isn't there already a PR for coloring thruster too?

Spartan322:
I'm pretty sure this is what mods are for, and I'm pretty sure that putting color in the sbc will be obsolete if #323 gets merged

Knsgf:
Why I didn't simply mod the thrusters? Because doing so will also increase WSAD control force (which is intended for performing fine movements and thus should remain relatively small) and TurboBoost strength -- things that this PR is designed to avoid.
Secondly, gyroscopes are boosted up in vanilla game when their override controls are switched on, so it would make sense for thrusters to do the same.
Thirdly, maneuvreing a ship by thrust override is more involved than doing the same with WSAD controls -- you need to reset the thrust manually and inertia dampeners are completely disabled for overridden thrust axis. And because this functionality is unconditionally available to all players, allowing thrust to be boosted via override is hardly unfair.

PR #367, opened 2015-07-23
Feature merge: 301 and 365 (depends on those PRs to be merged first)

lord-devious:
This pull request contains a combination of #301 and #365, and makes it possible for a functioning programmable block to start itself by enqueueing a run of itself in its constructor. This is the final stage of making a timer block completely redundant in relation to programmable blocks.

**A review and accept of #301 and #365 is required before this PR can be reviewed**. 
If either of those is rejected, then so must this one.

While this PR may look very large at a first glance, it actually only contains a few code lines. The largest changes are in those other PRs.


joemorin73:
Your evil plan is coming to fruition. ;)

lord-devious:
Since #301 is to be rewritten this one has no more meaning.

PR #369, opened 2015-07-24
FIX: Terminal property extension GetMinimum

mexmer:
Added missing GetMinimum to ITerminalProperty and TerminalPropertyExtension
Marked existing GetMininum as obsolete (not removing this function, since it might break existing scripts)


mexmer:
rebased

PR #373, opened 2015-07-28
Ore Detector - ore locations

Rynchodon:
Allows Mod A.P.I. access to ore locations


petie:
That is exactly what I was looking for to make fully autonomous drilling drones!

mze9412:
Would it be possible to do this also for the ingame scripts?

Rynchodon:
>Would it be possible to do this also for the ingame scripts?

It is possible but I have not come up with a good way of doing it.

PR #380, opened 2015-07-30
Recast gui ScreenTerminal for localisation

rem02:
Hi, I am french and i would like to show you my little contribution.

* More space for the info tab in GUI ScreenTerminal for localisation.
* Fix position of list when rename control is hide
* some tweaks


paradineai:
I am in the USA and like the newer look Would it be possible that you could take a look at the keensoftwarehouse forums pages to see if maybe you could do an upgrade to the rest of the menu screens described there. The link below is to the one I am referencing to specifically.
http://forums.keenswh.com/threads/a-bit-of-a-suggestion-and-request-for-ksh.7365229/

Issue #376, opened 2015-07-28
RequestShapeBlocking causing slowdown when player is distant from a mined asteroid


alex-davidson:
Reproduction of this bug seems intermittent but is generally correlated with mining from an inside surface of an asteroid, then moving >500m from the mined location.

Effect of the bug is regular 'hitching' of update speed. One or two times a second the update lag spikes considerably. Simulation speed drops to 50% or lower. Possibly related: https://github.com/KeenSoftwareHouse/SpaceEngineers/issues/362

I've applied a profiling tool to both the most recent and the immediate prior GitHub builds of SE and observed similar behaviour: http://i.imgur.com/t2rOfpu.png  I modified the source to log calls to CreateMesh (along with calling stack frame, duration, unique MyVoxelPhysicsBody instance id, triangle count, coord, etc).

One specific MyVoxelPhysicsBody instance, presumably the asteroid in question, dominates the log file.

At a certain point in time, strongly correlated with the start of the slowdown, CreateMesh starts being called almost exclusively from RequestShapeBlocking. Call duration is at least a millisecond.

When no slowdown is observed, the method takes less than a millisecond to return and more frequently returns null, ie. no mesh. Nulls are most frequently observed shortly before slowdown begins, ie. are correlated with distance up until the critical point.

Around the critical point, call rate for CreateMesh is very variable. Beyond a few seconds either side, it settles down to 100-180 per second. The slowdown appears to be caused by a) longer CreateMesh call duration and b) the fact that it happens inline with physics updates, on a callback from part of MyPhysics#Simulate.

RequestShapeBlocking gets called by Havok, presumably to satisfy a physics calculation (I'm not familiar with the Havok API or modern game engines). Could this be caused by Havok requesting meshes for things which the game has deemed 'out of range' and therefore not in need of precalc, forcing them to be generated on the update thread instead?

alex-davidson:
The increased rate and duration of RequestShapeBlocking calls starts immediately after m_aabbPhantom.CollidableRemoved fires and the number of 'nearby' entities falls to zero.

CodeHamster:
Thanks for the analysis, Ive had this problem in-game a lot recently and it's a killer.  Forces me to reload game after every mining mission as I'm heading back to base.  Once reloaded, everything settles down.  Would be very grateful for a fix.

alex-davidson:
[...]

PR #381, opened 2015-07-30
Retain voxel shape information while floating ore exists.

alex-davidson:
MyFloatingObject entities still collide with voxel objects when grids and
players are out of range. If shape information has been discarded this
will cause expensive RequestShapeBlocking calls to happen frequently on the
simulation thread, with a noticeable effect on sim speed.

refs issue #376

### DIFF
--- a/Sources/Sandbox.Game/Engine/Voxels/MyVoxelPhysicsBody.cs
+++ b/Sources/Sandbox.Game/Engine/Voxels/MyVoxelPhysicsBody.cs
@@ -33,6 +33,7 @@ namespace Sandbox.Engine.Voxels
         private bool m_needsShapeUpdate;
         private HkpAabbPhantom m_aabbPhantom;
         private readonly HashSet<IMyEntity> m_nearbyEntities = new HashSet<IMyEntity>();
+        private readonly HashSet<IMyEntity> m_nearbyFloatingEntities = new HashSet<IMyEntity>();
 
         /// <summary>
         /// Only locked in callbacks, since they can happen during multithreaded havok step.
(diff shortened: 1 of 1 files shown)


paradineai:
What will this do for processor, ram, and hdd use on a computer.

alex-davidson:
Should not affect HDD use.

A few bytes extra memory usage per asteroid, plus a couple of kilobytes (max) per floating object near that asteroid (for 256 object limit, upper limit should be 1-2KB per 'roid depending on 32-bit vs 64-bit, I think).

Asteroid shapes will be kept in memory longer than previously, ie. until all nearby floating objects have despawned. I don't know how much memory one of these consumes but it should only affect you if you've run into this bug. If you're not mining asteroids, there should be no floating objects, and the shape should be evicted as normal.

Processor usage under normal circumstances should not change. Processor usage when the player is out of range of an asteroid with floating objects nearby should be a lot lower and gameplay a lot smoother.

tl;dr: if this bug has never affected you, the game's resource usage should not change noticeably. If this bug did affect you, RAM usage might increase slightly.

MeusRex:
+1 (This is really necessary.)

After some heavy drilling on a remote asteroid my UPS drops from a constant 1.0 to 0.5-0.3 the moment I leave the asteroids immediate area. And once I create enough floating objects to force delete the ones close to the faraway asteroid the ups return to 1.0.

CheeseJedi:
Wow - if this works, then awesome work!

zrisher:
:+1:, keen please incorporate.

kurevin:
+1 to this one if still applicable

t3hk0d3:
+1

PR #383, opened 2015-07-31
FIX. Check availability "Convert To Ship" button.

JimLess:
Checks for grid ownership before displaying the 'Convert To Ship' button, similar to the 'Rename' button's behavior.


zrisher:
"Checks for grid ownership before displaying the 'Convert To Ship' button, similar to the 'Rename' button's behavior."

Pretty sure this is a highly requested feature. I know the majority of people on my server would like to have this additional safety mechanism.

JimLess:
zrisher, thanks.

mexmer:
OnRecreateControl is correct place for this check.

as for redrawing, that could be fixed also. i will check code and let you know. Possibly redraw can and should be triggered when ownership changes. Since usually you don't have lot of ownership changes in small timeframe

JimLess:
is up-to-date.
Please! Add this fix into next update!

PR #386, opened 2015-08-04
Feature: Terminalpropertyextension string

mexmer:
adds terminaproperty value of string to be retrieved or set
- IMyTerminaBlock.GetValue<string>(block, id) and IMyTerminaBlock<block>.GetValue<string>(id)
- IMyTerminaBlock.GetValue(block, id).AsString() and IMyTerminaBlock<block>.GetValue(id).AsString()
- IMyTerminaBlock.SetValue(block, id, string) and IMyTerminaBlock<block>.SetValue(id, string)
- IMyTerminalPropertyExtension.GetMinValue(), GetMaxValue(), GetDefaultValue() return empty string (eg.have no use on this property)

i'm currently investigating about adding posibility to retrieve maximum length of text for SetValue().

this can be used on textboxes (getvalue/setvalue) or labels (getvalue only)
from what i see there is not much blocks that will benefit from this, but i still plan to have all controls on block available for direct control (including lists and comboxes), to allow same control over block as from gui.

as a "sidefect" this exposes property field "Name" on each block, which equals to IMyTerminalBlock.CustomName

it can be also used to set argument on Programable block, then you can use Run action to apply.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Interfaces/ITerminalProperty.cs
+++ b/Sources/Sandbox.Common/ModAPI/Interfaces/ITerminalProperty.cs
@@ -58,6 +58,11 @@ namespace Sandbox.ModAPI.Interfaces
             return property.As<bool>();
         }
 
+        public static ITerminalProperty<string> AsString(this ITerminalProperty property)
+        {
+            return property.As<string>();
+        }
+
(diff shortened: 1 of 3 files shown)


mexmer:
closing, since there are only few blocks, that will benefit from this atm. most of text values is exposed directly in interfae. might return to this later

PR #387, opened 2015-08-05
Slightly more realistic gravity.

CrazyPotatoes:
Requesting Newtonian force model to enable fixed elliptical orbits. Updated force function to reflect that.

Fixed a little bug that presented when going underground. Once below the minimum hill radius, the gravity would instantly drop to half.  I simply doubled that attenuation.

Extended the field range to 4x the planet's radius instead of 2x. This makes high altitude orbits optional for low speed space station geeks.

Changed scaling to be slightly exponential. Realistic would be a cubic relationship. Used 1.5 instead. This way, a planet twice the diameter does not only have twice the gravity. Realistically it would have been 8x, but I made a change that makes it ~3.5x instead of 2x.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/MyPlanet.cs
+++ b/Sources/Sandbox.Game/Game/Entities/MyPlanet.cs
@@ -57,7 +57,7 @@ namespace Sandbox.Game.Entities
     public class MyPlanet : MyVoxelBase, IMyGravityProvider, IMyOxygenProvider
     {
         const int PHYSICS_SECTOR_SIZE_METERS = 2048;
-        const float DEFAULT_GRAVITY_RADIUS_KM = 50.0f;
+        //const float DEFAULT_GRAVITY_RADIUS_KM = 50.0f;
         const int ENVIROMENT_EXTEND = 1;
         const int ENVIROMENT_EXTEND_KEEP =  2*ENVIROMENT_EXTEND;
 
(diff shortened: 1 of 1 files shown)


plaYer2k:
While i like the step toward a more realistic gravity and thus large gravity fields and better fade, removing the customizable scales like *m_planetInitValues.GravityFalloff* might not be a good idea.

Also `* 2.52982f * (float)Math.Pow(10.0f, -7.0f)` should be a constant to reduce some processing power.

However i support all efforts to improve the gravity effects of planets toward a large and more realistic behavior.
Even though i highly assume that, as planets are the most insensely worked on feature right now, this PR will get merged as it either is obsolete or the working code at keen highly changed already so that it might be more a spiritual entry for keen and a useful one for everyone using the current source for planet exploration.

lord-devious:
Please allow Keen to *finish* a feature before trying to change it.

zrisher:
Well technically none of this is *finished* right? It's alpha and everything is changing rapidly? Most of the PRs and issues on this repo are probably never going to be implemented because they become obsolete or irrelevant, but having the info about people's requests and suggested implementation surely is still useful to Keen.

lord-devious:
True, but it seems rude to me to make changes to something that hasn't even been released yet.

plaYer2k:
Yeah i agree with @lord-devious there, hence why i highly doubt that it will get merged.
We dont know at all how the planet feature has developed for them internally.

krypt-lynx:
@CrazyPotatoes, from programmer to programmer:
Don't use magical numbers.

CrazyPotatoes:
Thanks for the feedback. This is my first time contributing anything code related to anyone.

I went back and reverted the expression to use the constant, although being new to programming, I think there must be a better way than using a constant? Like a variable that can be modified somewhere else? I don't remember how to write such a thing. 

The way that line works is if you set a constant to 25, then a planet with radius of 25km will have a surface gravity of 1.0.

Also I figured they've probably heavily modified or rewritten this script, but I'm hoping the model of gravity in the future matches with what I put up, because the orbits won't work otherwise.

koroden:
Few things:

-An orbit is possible with any kind of central force, and a steep decline probably makes better game play. But maybe a settable parameter would be nice.

-The surface gravity is linear in the radius (for the same density): It's proportional to mass/radius^2=rho*radius^3/radius^2=rho*radius. But this is just a game play choice since it would depend on the falloff exponent. The larger concern here is that with a steep power law exponent the range of planet radii with a "usable" surface gravity gets very slim.

CrazyPotatoes:
Orbits created by force that relates to the inverse of distance squared form the shape of a fixed non-moving ellipse about a focal point. Anything to lesser degree creates spinning elliptical shapes. Anything greater in degree degenerates into either an inward or outward spiral. The latter is the current model expressed in the original code, and testing verifies it's orbital degradation. That does not make better game play by any respect, but fair enough on the linear scaling.

I've set up a few orbits about the planets after modifying the code and they ran the same path over and over again. Using several beacons around the planet, I could see the min and max distances never changed as I made complete orbits. 

The problem with using whatever value Falloff was set to as an exponent, is that it resulted in orbital decay. Using the same speed, if I slow it down slightly, it just gets closer and closer as it circles the planet. If I sped the craft up a bit, it slowly got further and further away each pass. That is not game friendly. Spherical gravity generators had more "game" value than this.

If you have an object with a specific vector at a specific point near an object of gravitational force, it will follow a path that brings it right back where it started. That's the best model for "game" and I expect people will be really dissatisfied when they have to write scripts and cheats to maintain some form of orbit that doesn't result in destruction.

koroden:
Ups, my bad with the exponent, apparently  orbits with an exponent of <=-3 are unstable.

It would probably be better to fix the planetScale parameter to the maximum size of planets, so that, say, an orbit above the atmosphere around the largest planet is just possible with the standard velocity limit.

PR #389, opened 2015-08-06
Feature/Scripting: Access to RemoteControl autopilot features (basic)

mze9412:
Many people are trying to recreate the auto pilot with own ingame scripting code because it is not possible to access the ingame autopilot on a remote block. I extended the IMyRemoteControl interface with the necessary methods and made those either available (some only needed to be made public) or added them to the MyRemoteControl class.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyRemoteControl.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyRemoteControl.cs
@@ -2,11 +2,23 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using VRageMath;
 
 namespace Sandbox.ModAPI.Ingame
 {
     public interface IMyRemoteControl : IMyShipController
     {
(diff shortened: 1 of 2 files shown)


plaYer2k:
I just quickly checked the code, so forgive me if it is present, but for *RemoveWaypoint* to properly work there needs to be a method to get all waypoints imo. Otherwise you had to rely on the knowledge of previous additions in order to remove them.

Thus it needs a method to get all waypoints currently added.

Also please add null-checks for the parameters in AddWaypoint and RemoveWaypoint.
RemoveWaypoint(null) might work but AddWaypoint(null, null) might cause some unwanted behaviors.

mze9412:
Hi,

thanks for your feedback. I thought people should remember which waypoints they added ;) But a method would be good idea, you are right :D

About null checks: Yes and no. If I add them the methods should return a boolean to show success or failure or throw an exception if the value is null. Luckily Vector3 cannot be null (it is a struct) so it only affects the name in the add method.

PR #391, opened 2015-08-06
Don't check for moved door entities when grid isn't moving

zrisher:
`UpdateDoorPosition()` updates the positions of the various sub-entities (door parts) of an Airtight Door, be it a regular door, airtight hanger door, etc. This is called when the block initializes, opens/closes, and after every simulation frame. However it only needs to be called on simulation if moving. 

This PR does not change the current functionality in any way, just removes an unnecessary check when the grid is not moving. I've tested it with doors in game and the still have their position updated just fine.

This function is fairly expensive with a large number of doors. In my 5-week-old survival world we have 113 regular doors and 542 airtight hanger doors. When profiled, `UpdateDoorPosition()` took 5.08% sample time.

image
For comparison to the turret update function (which is only one part of the turrets' expensive updates), we have 116 turrets.

With this change, it doesn't even show up in `MyEntities.UpdateAfterSimulation`'s list.

image

Note all the numbers here a little lower because I spent less time in-game during that test.

**TLDR: Please merge this to improve door performance.**

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyAirtightDoorGeneric.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyAirtightDoorGeneric.cs
@@ -250,7 +250,7 @@ namespace Sandbox.Game.Entities.Blocks
         public override void UpdateAfterSimulation()
         {
             base.UpdateAfterSimulation();
-            if (CubeGrid.Physics == null)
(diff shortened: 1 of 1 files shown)


zrisher:
The additional check for `m_currSpeed != 0` that was added in upstream before `UpdateDoorPosition()` should fix this. Thanks!

PR #392, opened 2015-08-07
Fixed 587 "BlinkLenght" typos to "BlinkLength"

corey-:
I was trying to write a script in game one day and I used set value on "BlinkLength" and I kept getting a null reference exception so I investigated and found that everywhere "BlinkLength" should be it is misspelled as "BlinkLenght". The fields are the only cases where it affects players, but I figured they should all be changed anyways.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyLightingBlock.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyLightingBlock.cs
@@ -10,7 +10,7 @@ namespace Sandbox.ModAPI.Ingame
         float Radius{ get;}
         float Intensity{get;}
         float BlinkIntervalSeconds{get;}
-        float BlinkLenght{get;}
+        float BlinkLength{get;}
         float BlinkOffset{get;}
     }
 }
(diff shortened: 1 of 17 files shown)


lord-devious:
This will break existing worlds.

corey-:
@lord-devious Surely it would be simple to convert the world saves?

zrisher:
@Corey There's nothing particularly extensive built into the game that updates existing worlds to the newest code. That's why, for example, no existing cargo containers had their volumes changes by Thursday's patch. Because of that, we generally have to ensure backwards compatibility.

However, in this case it's an obvious typo, I'm sure people wouldn't mind adjusting their scripts for it. I wouldn't be surprised if significantly larger breaking changes were introduced into the ingame programming interface down the line.

THDigi:
How about not removing the incorrect word where it's needed for backwards compatibility and remove it further in the future when more important save-breaking features need to happen anyway (in case they do)... or just leave it like that forever xD

Spartan322:
This is a good example of why we need at least a simple deprecation system for scripts

lord-devious:
Deprecation system: Mark properties and methods with [Obsolete("Reason for deprecation here")] This will make it into the documentation at least.

@zrisher the problem isn't the scripts, it's the save games. Old save games won't work with this change whether you use scripts or not, unless @corey- actually writes conversion code to still take its value from the old name.

PR #395, opened 2015-08-08
[Bugfix] Fixed fields (Vector3D .X, .Y, .Z), which also stops very bad crashes in Win10. Fixed custom Enums.

lord-devious:
Before this fix, the following script crashes - sometimes safely, in Win10 sometimes even so hard there's no crash log.

This is now fixed, the vector X/Y/Z now return expected values - and no crashes.

```C#
void Main(string argument) 
{ 
    var pos = Me.GetPosition();
    Echo(pos.X + " " + pos.Y + " " + pos.Z);

    VRage.MyFixedPoint fp=(VRage.MyFixedPoint)10.12; 
 
  long rv=fp.RawValue; 
  Echo(rv.ToString()); 
 
  Echo(fp.RawValue.ToString()); 
 
//  Echo( ((VRage.MyFixedPoint)10.12).RawValue.ToString() ); 
}
```

Enums are also fixed. You can now do:
```C#
enum A {
    A,
    B,
    C
}

void Main() {
    Echo(A.B.ToString());
}
```


lord-devious:
Ok... Enums /work/ but their ToString is a bit wonky still. Working on it.

lord-devious:
Done!

Trojaner25:
+1

LynnuxX:
Does it only fix Vector3/3D or also other stuff like MyBlockOrientation ?
Using directions Forward and Up in programmable blocks seems to be bugged (wrong value returned by one of them)

lord-devious:
*Should* fix any and all field access. Vectors were only used as example.

LynnuxX:
Very nice !

lord-devious:
@Fujtajblik or whoever else in charge - a warning: I fixed the conflicts, but I was not able to run the game and test this after the fix. Is there a way I can get the current source to run?

lord-devious:
The important parts of this one was merged an alternate way.

PR #396, opened 2015-08-08
Added BlockDefinition to IMySlimBlock

THDigi:
Added `BlockDefinition` field for IMySlimBlock for both ingame and modding use, which is really useful to determine what the block actually is when it doesn't have a `FatBlock` available.

Also removed unnecessary fields in the modding interface because they already exist in the ingame one which is implemented by the modding one, therefore the fields are available there regardless.
With the exception of `Position` which is intentionally overwritten to provide the 'set' ability.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMySlimBlock.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMySlimBlock.cs
@@ -5,54 +5,35 @@ namespace Sandbox.ModAPI
 {
     public interface IMySlimBlock : Ingame.IMySlimBlock
     {
-        float AccumulatedDamage { get; }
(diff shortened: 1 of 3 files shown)


THDigi:
Rebased and fixed collissions.

LeonidasCZ:
Why so many deletions?

THDigi:
They're already defined in the Ingame.IMySlimBlock, less clutter and warnings for the project :}

It shouldn't break scripts or mods, I tested in my modding IDE with the compiled version of this and I was able to access the deleted things just fine from a ModAPI.IMySlimBlock reference.

I can undo that cleanup and leave only the added BlockDefinition field if you want.

PR #397, opened 2015-08-08
Bugfix/ilinjector short branches

alex-davidson:
Fixes remaining 'Illegal one-byte branch' error caused by Bgt_S and Bgt_Un_S opcodes.

### DIFF
--- a/Sources/VRage.Library/Compiler/IlInjector.cs
+++ b/Sources/VRage.Library/Compiler/IlInjector.cs
@@ -386,54 +386,25 @@ namespace VRage.Compiler
 
         private static System.Reflection.Emit.OpCode SwitchShortOpCodes(System.Reflection.Emit.OpCode code)       
         {
-            if (code == OpCodes.Bge_Un_S)
-            {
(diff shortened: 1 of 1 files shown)


alex-davidson:
Reproduction case for the bug:
```
void Main()
{
	var a = "A";
	var b = "B";
	var i = 1000;
	while(i-- > 0)
	{
		if(i > 500) continue;
		
		a = Method(a, b);
		b = Method(b, a);
		if(i > 500) continue;
		
		a = Method(a, b);
		b = Method(b, a);
		if(i > 500) continue;
		
		a = Method(a, b);
		b = Method(b, a);
		if(i > 500) continue;
		
		a = Method(a, b);
		b = Method(b, a);
		if(i > 500) continue;
		
		a = Method(a, b);
		b = Method(b, a);
		if(i > 500) continue;
		
		a = Method(a, b);
		b = Method(b, a);
	}	
}

string Method(string a, string b)
{
	return a + b + a;
}
```

lord-devious:
Good work! Nice to see somebody else working on the PB compiling issues as well.

Issue #398, opened 2015-08-08
IlInjector does not properly rewrite type usages


alex-davidson:
Consider the following:
```
public static B SelectSomething(A source) { return new B(); }
public struct B {}
public struct A {} 
 
void Main()
{
	var list = new List<A>();
	list.Add(new A());
	
	B[] result = System.Linq.Enumerable.ToArray(System.Linq.Enumerable.Select(list, SelectSomething));
}
```

Modifying MyProgrammableBlock.cs CompileAndCreateInstance to pass a final 'true' parameter to IlInjector.InjectCodeToAssembly will write out an IngameScript_safe.dll to the game's bin folder when the script is compiled. There will also be an IngameScript.dll present, which is highly suspicious since it's supposed to be in-memory only.

The reason why IngameScript.dll is present too is that it's ''referenced by'' IngameScript_safe.dll! The return type of SelectSomething will be Program.B from IngameScript rather than just B from IngameScript_safe.dll.

I'm not sure if this might be used to break the sandbox since the IL is still checked, but it does cause otherwise-valid programs to fail with obscure type-cast errors.

IlInjector.InjectTypes seems to assume that collecting types in inheritance order is sufficient, but it is not.

alex-davidson:
IlInjector is also intolerant of type dependencies. If class A has a field of type B, and B is a struct, B must be created before A.

alex-davidson:
I'm not entirely sure how generic methods ever worked here. I suspect the game always ended up calling the methods from the original binary, ignoring the injected version entirely and bypassing the instruction limits, etc.

alex-davidson:
Current working branch for this issue: https://github.com/alex-davidson/SpaceEngineers/tree/bugfix/ilinjector-member-rewrites

(It will be rebased off tool/script-checker prior to filing a PR)

alex-davidson:
I think this branch is moderately robust now.

It seems the the IngameScript.dll will be written out in this configuration even if it's not referenced, but Reflector/dotPeek reveal that my test scripts are no longer generating IngameScript_safe.dll binaries which reference the other.

Not yet handled (haven't even looked at it):
* Script-defined generic types
* Enums

Undefined behaviour (compiles, not tested it runs):
* Script-defined generic methods

alex-davidson:
Still got a problem around field members. Any value-typed field prevents the Program type being loaded.

alex-davidson:
[...]

PR #399, opened 2015-08-09
Bugfix/ilinjector member rewrites

alex-davidson:
Fixes multiple bugs in IlInjector around type rewriting.
- The _safe injected assembly no longer uses methods and types from the uninjected one.
  - Generic type parameters fixed, ie. List of A
  - Ref and array types fixed
  - Generic method calls fixed, eg. Linq
- Type dependencies are considered when finalising the injected module.
- Performance improvements.

### DIFF
--- a/Sources/VRage.Library/Compiler/IlInjector.cs
+++ b/Sources/VRage.Library/Compiler/IlInjector.cs
@@ -228,6 +228,7 @@ namespace VRage.Compiler
             if(typeLookup.TryGetValue(type, out replacementType)) return replacementType;
             return type;
         }
+
         private static void CopyFields(Dictionary<FieldInfo, FieldBuilder> createdFields, Type sourceType, TypeBuilder newType, Dictionary<Type, TypeBuilder> typeLookup)
         {
             var fields = sourceType.GetFields(BindingFlags.Static |BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.SetField | BindingFlags.GetField | BindingFlags.Instance);
@@ -327,16 +328,16 @@ namespace VRage.Compiler
                 switch (instruction.OpCode.OperandType)
                 {
                     case OperandType.InlineField:
-                        ResolveField(instruction.Operand as FieldInfo, createdFields, methodGenerator, code);
+                        ResolveField(instruction.Operand as FieldInfo, typeLookup, createdFields, methodGenerator, code);
                         break;
                     case OperandType.InlineMethod:
                         try
                         {
-                            ResolveMethodOrConstructor(methodGenerator, createdMethods, createdConstructors, instruction, code);
+                            ResolveMethodOrConstructor(methodGenerator, typeLookup, createdMethods, createdConstructors, instruction, code);
                         }
                         catch
                         {
-                            ResolveField(instruction.Operand as FieldInfo, createdFields, methodGenerator, code);
+                            ResolveField(instruction.Operand as FieldInfo, typeLookup, createdFields, methodGenerator, code);
                         }
                         break;
                     case OperandType.InlineTok:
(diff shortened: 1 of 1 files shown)


alex-davidson:
Refs issue #398

lord-devious:
You are doing duplicate work from stuff I have already done :smile:

lord-devious:
#349 #395 #301

alex-davidson:
Not necessarily a bad thing :) We seem to have approached the problem slightly differently, so the Keen devs can pick which solution they prefer.

Your fixes deal with some of the problems and certainly tackle the nesting issue. Also looks like you've dealt with generic constraints too, which I haven't paid much attention to. I'm not a particularly active member of the modding/scripting community so I don't know how much demand there is for defining generic types in scripts.

Have you dealt with the type usage rewrite problem? I'm sorry I don't have my development environment in front of me right now so I can't check, but it looks like the generated assembly will continue to reference the original for some script-defined types?

lord-devious:
Yes, you're right - I just wanted to make sure you were aware of it.

Generics are extremely important for scripting. Before my fix is merged you can't even use List<> with custom classes. As for the type rewrite I'm doing my best, I've captured a few, but who knows if I have them all. Probably not. This isn't exactly the easiest of areas, as I'm sure you've noticed  :smile:  also since the pause in merging, I'm starting to have the problem of stumbling over my own feet as it were, because the earlier fixes hasn't been merged yet, making it difficult to continue. I could make a single large PR, but Keen prefer smaller ones, because they are easier to review.

alex-davidson:
Heh, I know that feeling indeed :D

I've got quite a lot of experience of dealing with strange reflection-related bugs in .NET, so it's sort of 'my area' as it were, but still far from simple. Ideally I'd submit a single refactoring patch before trying any bugfixing in this code but, as you say, the merges are rare. Hard to iterate quickly.

My own scripts make a lot of use of Linq extension methods and generic CLR types too, and I'm pretty sure my own patches cover those as generally as possible, but it's script-defined ones I haven't really put any effort into yet.

Given the extremely tight restrictions on line and character count enforced by the game I haven't had need to define my own generic type yet, preferring to use Linq structures for data and keep any specialised accessor code in static methods. I'm hoping that your contributions towards supporting 'yielding' scripts which can spread their work over multiple updates will lead to those restrictions being relaxed slightly!

lord-devious:
Small world... I too have a lot of experience with reflection :smile:. Granted, I haven't had much experience with direct IL manipulation before this game, but having already had a few related bugfixes and features merged and released, I'm getting there.

When I say List<> with custom classes, I don't mean a custom generic class - I mean a custom _any_ class. A normal List<ScriptClass> will fail, you have to work around it by using List<object> or a struct wrapper, which for some reason does work.

I tend to avoid doing too much refactoring unless I feel it's absolutely necessary. Others' more purely "aesthetic" PRs have been rejected in the past, and the less changes there are for Keen to review, the more likely it is to me merged. Or so I believe, at least.

As to Linq - Keen frowns upon the use of Linq in their code, because it uses a lot of allocations. In a game of this complexity this is bad, because even though .NET handles the memory use, it makes the GC go ballistic and affect performance. This has caused me to avoid allocations in my own scripts, at least those who have a high repeat factor.

The character count restriction is there because the entire script gets transmitted in multiplayer, so they want to restrict its size. I doubt it will be increased, but hey, who knows.

alex-davidson:
I think this patch series fixes all generic constructions since they just require recursively rewriting the type usages. It's the defining of new generic types that I haven't looked into yet.

Hmm, allocations shouldn't be a big deal unless they survive a GC. I guess if there's a lot of them happening the 'critical lifespan' of an object is reduced, making it more likely to get promoted, but even a gen1 GC shouldn't be that big a deal. I agree that letting anything disposable go near gen2 would be a serious issue, though, even with .NET 4.5's 'low latency' GC options.

My standard pattern is to allocate any large structures once for the script, and try to ensure that repeat runs don't hang onto any references when they exit. On my to-do list is running a memory profiler against the game to get an idea of gen0 and gen1 collection frequency and then tune my scripts to yield within half a gen0 interval.

lord-devious:
Defining new generic types works great. I don't remember if they always have or if one of my bugfixes fixed it at one point... I was on board before the GitHub went public, on its trial run, so I can't actually remember what I've fixed and what they fixed themselves...

While I agree with you about allocations, it's their game, their policy - who am I to argue. I never went into the details on how the GC actually works, because I never had the cause to. I'm an application developer by occupation, not a game developer - so these kinds of concerns haven't really been relevant.

alex-davidson:
Heh, me too :) Last time I did anything akin to games dev was ten years ago, and that was C++, so working with the SE source has been quite fascinating. A change in perspective.

lord-devious:
Almost the same for me, I've been dabbling with various C# based game engines just for fun, but the last time I did anything semi-serious with games, was 10-15 years ago, in C++.

PR #400, opened 2015-08-09
Read only entity from sensor

mdfuego:
Restricted ingame scripts from modifying or applying actions on blocks detected through the ship's sensor(s). In other words, ingame scripts no longer have full control over any ship detected by sensors.  
An example of an ingame script that takes control of a detected ship is at http://steamcommunity.com/sharedfiles/filedetails/?id=404907524

This is my first pull request, and I'm still very new to the codebase, so if anyone can see a better way to do this, then I'm all ears.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyCubeBlock.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyCubeBlock.cs
@@ -19,6 +19,7 @@ namespace Sandbox.ModAPI.Ingame
         string GetOwnerFactionTag();
         Sandbox.Common.MyRelationsBetweenPlayerAndBlock GetPlayerRelationToOwner();
         Sandbox.Common.MyRelationsBetweenPlayerAndBlock GetUserRelationToOwner(long playerId);
+        bool isControllableForProgrammableBlock { get; }
         bool IsBeingHacked { get; }
         bool IsFunctional { get; }
         bool IsWorking { get; }
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyProgrammableBlock.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyProgrammableBlock.cs
@@ -248,6 +248,8 @@ namespace Sandbox.Game.Entities.Blocks
             var gridGroup = MyCubeGridGroups.Static.Logical.GetGroup(CubeGrid);
             var terminalSystem = gridGroup.GroupData.TerminalSystem;
             terminalSystem.UpdateGridBlocksOwnership(this.OwnerId);
+            terminalSystem.updateControllingProgrammingBlock(this.SlimBlock);
+
             m_instance.GridTerminalSystem = terminalSystem;
 
             m_isRunning = true;
--- a/Sources/Sandbox.Game/Game/Entities/Cube/MyCubeBlock.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Cube/MyCubeBlock.cs
@@ -114,6 +114,11 @@ namespace Sandbox.Game.Entities
             get { return MySandboxGame.TotalTimeInMilliseconds - HackAttemptTime < MyGridConstants.HACKING_ATTEMPT_TIME_MS; }
         }
 
+        public bool isControllableForProgrammableBlock
+        {
+            get { return CubeGrid.isProgrammaticallyControllable(); }
+        }
+
         public MyCubeBlockDefinition BlockDefinition { get { return SlimBlock.BlockDefinition; } }
         public Vector3I Min { get { return SlimBlock.Min; } }
         public Vector3I Max { get { return SlimBlock.Max; } }
--- a/Sources/Sandbox.Game/Game/Entities/Cube/MyCubeGrid.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Cube/MyCubeGrid.cs
@@ -298,6 +298,14 @@ namespace Sandbox.Game.Entities
             MainCockpit = cockpit;
         }
 
+        // The block that is accessing this MyCubeGrid - usually the currently running programming block
+        public MySlimBlock programmaticallyControllingBlock;
+
+        public bool isProgrammaticallyControllable()
+        {
+            return CubeBlocks.Contains(programmaticallyControllingBlock);
+        }
+
         public MyCubeGrid() :
             this(MyCubeSize.Large)
         {
--- a/Sources/Sandbox.Game/Game/GameSystems/MyGridTerminalSystem.cs
+++ b/Sources/Sandbox.Game/Game/GameSystems/MyGridTerminalSystem.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Text;
+using System.Linq;
 using System.Collections.Generic;
 using Sandbox.Game.Entities.Cube;
 using VRage.Collections;
@@ -148,6 +149,23 @@ namespace Sandbox.Game.GameSystems
             {
                 block.IsAccessibleForProgrammableBlock = block.HasPlayerAccess(ownerID);
             }
+        }
+
+        // sets the programming block that is accessing the cubeGrid in order to enforce "read-only" access to detected blocks
+        // that are not part of the programming block's cubeGrid
+        public void updateControllingProgrammingBlock(MySlimBlock programmableBlock)
+        {
+            // currently only sensors and landing gears return "detected" blocks
+            // because landing gears have a physical connection to the attached cubegrid they may be considered able to have "write" access.
+
+            // TODO: make cubeGrids returned by landing gears respect block ownership - possibly through the UpdateGridBlocksOwnership() function
+            // alternatively, add any cubeGrids they return to the detectedCubeGrids function
+            IEnumerable<Sandbox.Game.Entities.MyCubeGrid> detectedCubeGrids = Blocks
+                .Where(terminalBlock => (terminalBlock is Sandbox.ModAPI.Ingame.IMySensorBlock) &&
+                    ((Sandbox.ModAPI.Ingame.IMySensorBlock)terminalBlock).LastDetectedEntity is Sandbox.Game.Entities.MyCubeGrid)
+                .Select(sensorBlock => ((Sandbox.ModAPI.Ingame.IMySensorBlock)sensorBlock).LastDetectedEntity as Sandbox.Game.Entities.MyCubeGrid);
+
+            detectedCubeGrids.ForEach(detectedCubeGrid => detectedCubeGrid.programmaticallyControllingBlock = programmableBlock);
         }
     }
 }
--- a/Sources/Sandbox.Game/ModAPI/Blocks/MyFunctionalBlock_ModAPI.cs
+++ b/Sources/Sandbox.Game/ModAPI/Blocks/MyFunctionalBlock_ModAPI.cs
@@ -17,5 +17,11 @@ namespace Sandbox.Game.Entities.Cube
             add { EnabledChanged += GetDelegate(value); }
             remove { EnabledChanged -= GetDelegate(value); }
         }
+
+        void Sandbox.ModAPI.Ingame.IMyFunctionalBlock.RequestEnable(bool enable)
+        {
+            if (this.isControllableForProgrammableBlock)
+                RequestEnable(enable);
+        }
     }
 }
--- a/Sources/Sandbox.Game/ModAPI/Blocks/MyTerminalAction_ModAPI.cs
+++ b/Sources/Sandbox.Game/ModAPI/Blocks/MyTerminalAction_ModAPI.cs
@@ -10,21 +10,22 @@ namespace Sandbox.Game.Gui
 {
     public partial class MyTerminalAction<TBlock>
     {
+
         void ModAPI.Interfaces.ITerminalAction.Apply(ModAPI.Ingame.IMyCubeBlock block)
         {
-            if (block is TBlock)
+            if (block is TBlock && block.isControllableForProgrammableBlock)
                 Apply(block as MyTerminalBlock);
         }
 
         void ModAPI.Interfaces.ITerminalAction.Apply(ModAPI.Ingame.IMyCubeBlock block, ListReader<TerminalActionParameter> parameters)
         {
-            if (block is TBlock)
+            if (block is TBlock && block.isControllableForProgrammableBlock)
                 Apply(block as MyTerminalBlock, parameters);
         }
 
         void ModAPI.Interfaces.ITerminalAction.WriteValue(ModAPI.Ingame.IMyCubeBlock block, StringBuilder appendTo)
         {
-            if (block is TBlock)
+            if (block is TBlock && block.isControllableForProgrammableBlock)
                 WriteValue(block as MyTerminalBlock, appendTo);
         }
 
--- a/Sources/Sandbox.Game/ModAPI/Blocks/MyTerminalBlock_ModAPI.cs
+++ b/Sources/Sandbox.Game/ModAPI/Blocks/MyTerminalBlock_ModAPI.cs
@@ -97,5 +97,17 @@ namespace Sandbox.Game.Entities.Cube
         {
             (MyTerminalControlFactoryHelper.Static as IMyTerminalActionsHelper).GetProperties(this.GetType(), resultList, collect);
         }
+
+        void ModAPI.Ingame.IMyTerminalBlock.SetCustomName(string text)
+        {
+            if (isControllableForProgrammableBlock)
+                SetCustomName(text);
+        }
+
+        void ModAPI.Ingame.IMyTerminalBlock.SetCustomName(StringBuilder text)
+        {
+            if (isControllableForProgrammableBlock)
+                SetCustomName(text);
+        }
     }
 }
--- a/Sources/Sandbox.Game/ModAPI/Blocks/MyTextPanel_ModAPI.cs
+++ b/Sources/Sandbox.Game/ModAPI/Blocks/MyTextPanel_ModAPI.cs
@@ -16,27 +16,31 @@ namespace Sandbox.Game.Entities.Blocks
 
         void ModAPI.Ingame.IMyTextPanel.ShowPublicTextOnScreen()
         {
-            SyncObject.SendShowOnScreenChangeRequest((byte)ShowTextOnScreenFlag.PUBLIC);
+            if (isControllableForProgrammableBlock)
+                SyncObject.SendShowOnScreenChangeRequest((byte)ShowTextOnScreenFlag.PUBLIC);
         }
 
         void ModAPI.Ingame.IMyTextPanel.ShowPrivateTextOnScreen()
         {
-            SyncObject.SendShowOnScreenChangeRequest((byte)ShowTextOnScreenFlag.PRIVATE);
+            if (isControllableForProgrammableBlock)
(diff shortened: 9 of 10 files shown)


MeusRex:
Why does the IMySensor even allow access to an IMyEntity? That's pretty much always an illegal action.
I think the simplest solution would just be to remove that entry. (And if anything of it is needed allow readonly access to that specific field.)

lord-devious:
Welcome aboard! :smile:

You should make sure you follow Keen's coding standard. You are currently deviating significantly, it is very unlikely that Keen will merge this as it is.

Also, food for thought: The problem you are currently trying to fix is part of a much larger problem, which most likely needs a common solution for all instances.

mdfuego:
MeusRex, 
I believe the sensor originally didn't include it, but then it was specifically added to allow scripts to get info on what was detected to be more useful, but by so doing a whole can of worms got opened up.

lord-devious,
Could you link to (or otherwise direct me to) Keen's coding standard? This is definitely a first-draft sort of fix (both in the super ugly code and in the implementation itself).  I mostly wanted to get feedback on a possible way to fix this the "right" way - I know that what I'm doing works for the easiest cases, but likely fails many corner cases.

lord-devious:
Please do not misunderstand me. This might be useful indeed.

There is no document outlining the standard. You need to derive that by reading their code, seeing how they do things. They themselves vary their standards slightly from file to file (since they have more than one developer :wink:, but they do have an overall style.

I can give you a jump-start, however. For instance, you seem to use java-esque casing, where all your members are camelCased. Keen (and the C# community in general) uses PascalCasing for all public members. camelCasing is reserved for private and protected fields, variables and method arguments. Keen also prefix their class member fields with m_ - a practice I disapprove of, but my personal opinion in this matter is irrelevant :wink:

Finally, you should avoid exposing class fields directly outside classes (except in structs). They _should_ always be wrapped in a property.

mdfuego:
Before I clean up the code, I want to ask about a concern I have with how I'm implementing this.  I mimicked the way that ownership is dealt with in ingame scripts (ie IsAccessibleForProgrammableBlock being set within the actual block). We are modifying the blocks themselves, and we are essentially saying "hey whatever ingame script is running got clearance to do this action before it ran, and I'm assuming that whatever is calling this function now is that script".

My main concern is that I have to assume then that ingame scripts are run in sequence, and never in parallel as otherwise this would break if there were two programming blocks owned by different factions on the same ship (or two sensors that detect the same ship).  I can see why there may be reasons not to have concurrent programs running on the same ship, but this limits the entire game world to only running one script at a time. 

In the case that ingame scripts can only run one at a time, wouldn't it be better to have a variable that tracks the source of the ingame script (ie which programming block) that is currently running?  That way any functions that require a friendly faction or other relationship with the block (like being part of the same cubegrid) can reference that variable.  I don't see why we have to compute that beforehand for each block regardless of whether the program even calls functions on those blocks.

lord-devious,
I appreciate the input! Thanks for the tips.

lord-devious:
You're welcome :smile:

Scripts can indeed only be run one at a time, never parallell.

mdfuego:
Ok, I've come up with a solution that I like much better.  Getting access to a block via an ingame script by any means now respects both friendliness and connectivity (ie part of the same TerminalSystem) to the programming block running the script. Because of this, the following is now true:
* If a block is stored as a member variable while it is friendly, then the block's ownership changes to an enemy, the script can still "read" data from the block, but it can no longer run actions or "write" to the block.  Previously, the script would have full control over the block.
* If an ingame script gets a reference to a block via sensors, landing gears, or any other means that is currently available or one day will be available, but that block is not "friendly" or is not part of the programming block's TerminalSystem, then the script will only have "read" access to the block.
* The "gold source" for if a block is accessible to the script no longer resides within the block itself, but in the running script.

There's still some minor code cleanup required on this, but I just want to see what people think.

JoeTheDestroyer:
I glanced over your code and my first comment would be that I don't think failing silently is the right thing to do. Although I'm not a fan of exceptions in general, in this case I think you should throw some kind of "access denied" exception to notify and make people have to address their illegal access. Next, I think you missed some write-type functions: for example, the inventory transfer functions are not something an enemy should have access to. Also, how does your changes impact the ModAPI? At a glance, it looks to me like a mod would generate null-pointer exceptions if it tried to change text panel related items, since IsAccessibleByProgram assumes m_runningProgramBlock is non-null.

As @lord-devious alluded to, you appear to be somewhat addressing the so-called merge block exploit. That is, the fact that references to blocks can long outlive when they were nominally accessible. The problem I have with your approach is that I don't think merely restricting to read-only access is good enough. The main thing preventing us from adding PB access to the antenna network is this exploit and your approach doesn't fix it, it merely narrows it's scope. What is needed is to completely deny access to inaccessible blocks (e.g., when antennas go out of range). While you could accomplish this by modifying all the read-only properties as well, this would be a lot of work and may have an impact on ModAPI performance (unnecessary permissions checks on property access).

As w/ @lord-devious, I've been thinking about this but I haven't come up w/ any good solutions either (the best so far would be PB-specific facades for all block types.)

mdfuego:
@JoeTheDestroyer,

> I don't think failing silently is the right thing to do. Although I'm not a fan of exceptions in general, in this case I think you should throw some kind of "access denied" exception to notify and make people have to address their illegal access.

I also am not a huge fan of throwing exceptions if not needed, but I see your point. I'll add it to the "to do" list

> Next, I think you missed some write-type functions: for example, the inventory transfer functions are not something an enemy should have access to.

Thanks for the heads up.  I'll go through the available functions in more detail after I resolve some of the other things you've brought up

> Also, how does your changes impact the ModAPI? At a glance, it looks to me like a mod would generate null-pointer exceptions if it tried to change text panel related items, since IsAccessibleByProgram assumes m_runningProgramBlock is non-null.

I've not dealt with the ModAPI much, so I coded a bit "fast and loose" with that bit.  Are ModAPI functions that don't require permissions checks called exclusively outside of ingame scripts, or can they be triggered while a script is running?  If they are mutually exclusive, then I can just do a null check on the RunningProgramBlock and allow access if it's null.  That way it solves both issues in one blow.  If ModAPI calls can occur within ingame scripts (and should not use the permissions granted by the PB), then I'll have to think on this some more.

> The problem I have with your approach is that I don't think merely restricting to read-only access is good enough. The main thing preventing us from adding PB access to the antenna network is this exploit and your approach doesn't fix it, it merely narrows it's scope. What is needed is to completely deny access to inaccessible blocks (e.g., when antennas go out of range). 

I could set up permissions as an enum (noAccess, readOnly, and readWrite).  In the case of sensors there wouldn't be much point in the LastDetectedEntity if you can't perform the read functions, but you are correct about the need to deny all access to antennas going out of range, ships that have left sensor range, etc.  The troublesome part of that is that I would have to check for any blocks that can return other blocks, and add them to a readOnly List (I think that only sensors, landing gears, and connectors do this, but there may be more).  Another problem is that if some other block comes along that returns "read only" type blocks in the future, then those would have to be added into the logic that builds the readOnly list.

> While you could accomplish this by modifying all the read-only properties as well, this would be a lot of work and may have an impact on ModAPI performance (unnecessary permissions checks on property access).

Yeah, I'm not a fan of requiring each method call to check for permissions either - it's easy for a function to be missed, or require the wrong level of permissions, and every new function that gets written needs to include it.  As much as I hate this way of doing it, I don't see any other way around it other than the PB-specific facades you mentioned, and I'm guessing that has its own issues, or that would have been implemented already. If I am able to do a simple null check on the running PB, then the performance impact should be minimal for the ModAPI.

I'll be working on some of these problems, but if anyone sees huge flaws in my plans, then please let me know.

mexmer:
as for modapi scripts access, i commented multiple times, and thing is, that for owner ship related issues responsibility lies on mod/modapi script author.

there are lot of interfaces, that allow you to access any and every entity in your world (including characters). therefore having part of api, that will restrict you on acl basis, while other doesn't will only create inconsistency.

block reference inside must be handled entirely different way, and on block reference base, not on function level. essentialy, whener you use block object/block reference in pb, it must go trough "translate with access check", and then produce either functional interface, on fake interface, that does only minimum functions.

JoeTheDestroyer:
>  Are ModAPI functions that don't require permissions checks called exclusively outside of ingame scripts, or can they be triggered while a script is running?
The in-game API is basically a subset (whitelist) of the ModAPI. So any in-game API function could be run by a mod (outside the in-game scripting environment). As @mexmer said, the expectation in the ModAPI is that everything is accessible, regardless of permissions/etc, and it is up to the mod author to ensure they are doing the right thing.

In that light, yes, I think a simple null check would be sufficient.

> The troublesome part of that is that...
Indeed, I think many of us see this as a fundamental flaw in the design of in-game scripting itself, making solving it a huge problem. That's probably why you're the first brave soul to tackle it...

> I don't see any other way around it other than the PB-specific facades you mentioned, and I'm guessing that has its own issues
Many. I didn't bring them up because I didn't want clutter your PR with discussion of a different solution.

> If I am able to do a simple null check on the running PB, then the performance impact should be minimal for the ModAPI.
Yes, I agree. I mostly brought it up as an FYI, that any checks need an early-out in case they are run outside the in-game scripting environment.

mdfuego:
So, after deciding to use the three permissions levels (noAccess, readOnly, and readWrite) I've been thinking of the two big problems left that I don't like about this solution, namely the following:
* Each getter, setter and method exposed in the Sandbox.ModAPI.Ingame namespace must be modified to check for the access level of the running script.
* I have to loop through the blocks in the PB's terminalSystem and check against a hard coded list of blocktypes that return CubeGrids, or CubeBlocks, and add those blocks to the list that grants "readonly" access. Each time a new block is exposed in this way (like some new sensor type, or whatever) this code would have to be updated.

I think I can make a *fairly* elegant solution to this if I can find a way to do the following:

* Use some sort of code injection, or other methodology that throws an exception if the PB doesn't have at least "readonly" access to the block before any of its getters, setters, or methods are called.  This way only setters and "write" type methods need to have code written to check for that level of access.

* use reflection to get a list of all Sandbox.ModAPI.Ingame interfaces (and their associated properties / methods) that return an IMyEntity object.  This way the code can ensure that it checks for all possible instances of "readonly" type blocks.

Unfortunately, I've never really done anything with reflection or code injection.  I'm not even sure if C# allows for code injection (I know that Java does that sort of thing regularly, but I've never done it myself).  If anyone can point me in the right direction or tell me it's impossible I would really appreciate it.  Also, I'm not against someone cloning my branch and doing a pull request with the implementation.

Lastly, to everyone who has provided feedback and help already - thank you so much!  This being my first pull request, I've had a great introduction to what I feel like is a pretty awesome community here. I feel like I started with a very basic and badly implemented fix, and with the help I've gotten along the way it's starting to turn into something I'm proud of regardless if keen decides to merge it.

mexmer:
which is why i said, this needs to be done on block/entity level, and not on function level (you will need to modify any and every function in exposed interfaces).
we discussed already i think with @lord-devious posibilities of block proxy interface, but it's easier said than done.

mdfuego:
@mexmer,

> this needs to be done on block/entity level, and not on function level

Ideally, yes, if I knew a good way to deny access to blocks with no access at the block/entity level, then that would be great.  If someone lets me know of a good way to do this, then I'm all ears. That being said, the "readonly" / "write" access will need to be done at the function level anyway, as some functions should be available for use when detected outside of the PB's terminal system, and others should not.  Because I don't know how to do block level access restrictions (as they may store a reference to the block when it's valid to do so, and then later should be denied access), and the "readonly" / "write" access needs the function level access anyway I figure I'd just do it all at the function level.

Again, if anyone has a good way to restrict access at the block level, then I would be more than happy to do that for when there should be no access to the block.

mze9412:
Why not simply denying access to any blocks that are not owned by the PB owner? This could be achieved by not exposing any method that can return blocks from different owners.

For sensors I would propose to replace the DetectedEntity with a special type which is read only (for example IMyDetectedEntity) and contains specific information (Location, type of the entity, maybe info if enemy, neutral, own, etc). 

This way it should become impossible to access enemy blocks and such a overly complicated rights architecture is not necessary.

mdfuego:
@mze9412,

> Why not simply denying access to any blocks that are not owned by the PB owner?

That is what currently occurs (though it's currently limited to only the PB's home GridTerminalSystem) - the problem with that is when the script stores the block as a member variable when it does have valid access (ie when the ship is docked, before an enemy claims the block, etc.) then something occurs that would restrict access (ie the ship undocks, or an enemy claims the block).  In that case, the script retains full access to the block(s) due to the fact that they have a reference to the block. From what I understand, this is known as the "merge block exploit".

> For sensors I would propose to replace the DetectedEntity with a special type which is read only (for example IMyDetectedEntity) and contains specific information (Location, type of the entity, maybe info if enemy, neutral, own, etc).

The only way I can think to do this is by writing a "Detected" version of every ingame accessible interface, and I'm sure there are other issues with this as well.  This may be feasible if there is a way to auto-generate those classes (just as I'm trying to inject the permissions checks), but that is something I don't know how to do.  Again, this may be feasible, and perhaps even a better solution, but I don't know how I would actually implement it.

mze9412:
The only thing a sensor can detect are entities, so I only see one interface for that part of the issue.

I did not think about the claiming of blocks by the enemy :\

mdfuego:
>The only thing a sensor can detect are entities

Yes, but many things inherit from entity, so I can say 

````
if (lastDetectedEntity is ImyCubeGrid)
{
ImyCubeBlock blockOfDoom = ((ImyCubeGrid)lastDetectedEntity)
                .GetCubeBlock(placeWhereThereIsABlock)
                .FatBlock;
   doStuffIShouldNotBeAbleToWithABlockIAmNotConnectedTo(blockOfDoom)
}

````

and that block can be any block. that, in turn can be cast as a ImyFunctionalBlock, etc.  also, landing gears and connectors also return entities / blocks.

mze9412:
That's way I said that "LastDetectedEntity" should not be there for ingame scripting. Ingame scripting needs a different method which returns an object which is not part of the normal hierarchy but is specifically created with read only information. That way it would be impossible to get access to a grid via a sensor, which is already have the problem.

lord-devious:
Actually it's a symptom of an even greater problem, if you can access blocks you don't have access to by using GetCubeBlock in IMyCubeGrid.

I think the entire programmable block API needs to be redesigned, even if it means breaking existing scripts. It needs completely sandboxed interfaces, which has no way at all to access the general API.

mze9412:
That might be a good idea anyway, yeah :)

mdfuego:
>Actually it's a symptom of an even greater problem, if you can access blocks you don't have access to by using GetCubeBlock in IMyCubeGrid.

yup, that's what I'm trying to address

> I think the entire programmable block API needs to be redesigned

well yeah, sure, but unless keen says that they are planning on this in a blog post, I'm going to continue to try it this way.

zrisher:
> I think the entire programmable block API needs to be redesigned

> > well yeah, sure, but unless keen says that they are planning on this in a blog post, I'm going to continue to try it this way.

They haven't, but there are a lot of posts on here with various improvements to the PB interface. If a team of you guys interested in improving the PB interface were to get together to incorporate all that work into a single new interface solution, it would be awesome. It seems like a redesign is going to be a lot of effort and probably break some things, so probably easiest to get Keen to accept if done at once.

But I understand you're probably hesitant to put that much work in until accepting PRs becomes a greater priority for them.

lord-devious:
I have a working concept. Or two, actually. One I can almost do from within the whitelisting of a programmable block, so it's simple but loses existing interface extensions - but they can be replaced. There are ways around that though. The other means creating a separate proxy for each and every block, which means a whole lot more work.

But no, while Keen has a low priority of merging, this won't happen by me. I mean, I already have 6 PRs waiting, at least one that fixes a very serious bug.

mdfuego:
So, I haven't implemented any of the "check for read only access" stuff, but I have added most of the "check for write access" stuff in a way that I like.  I still need to do a thorough check through all the ingame interfaces to be sure to get to all of them, and throw errors if a script tries to perform an illegal action on a block.

t3hk0d3:
Imho C# was a poor choice for PB for several reasons. This issue is one of the reasons - you can't completely control sandbox, and there is possibility PB can affect not just enemy ships, but possibly server management code, for example - hi-jack servers, banning administrators, etc.

Even more C# itself is not good for purpose of PB because of its static-typed nature.

Thats why i came with idea of using completely sandboxed programming language without even access to server code directly. All interactions should be made thru wrappers. Something like IronPython, IronRuby, Boo, Lua, etc.

lord-devious:
I'm sorry, but that is simply not correct. Purely technically speaking, C# is the perfect choice because it is the quickest and easiest to implement, because the game itself is made in C# and the C# compiler is available by default in .NET. This means that you get no overhead whatsoever when running the scripts. And you can create as much of a sandbox in C# as in any other language, claiming anything else is just silly. They have simply not done a good enough job of it - yet. Their main error was to give us the actual blocks rather than a sandboxed reference. Any system requiring an extra layer between the game and the script is bound to be bug prone, not to mention the extra overhead you will unavoidably get.

t3hk0d3:
@lord-devious С# is not good for scripting itself. And sharing same VM as rest of server is bad idea as well. Perfomance and Implementation simplicity is not proper concern for this topic.

lord-devious:
I do agree that the _language_ C# is not a good scripting language, but only because it's a "professional" programming language and not the easiest to learn. But one does not need to use a non-IL language to change that. Boo, for instance, already compile to IL I believe. But using an extra non-IL layer between adds perfectly unnecessary complication. You do _not_ get any kind of advantage you can't achieve using pure .NET. You most certainly do not have to run in a "different memory space" to be perfectly safe, namespace has nothing at all to do with this but I'm assuming that's not exactly what you meant so let's ignore that. If one insists on a complete separation, you can even do that within the .NET system itself. You don't need to add a non-IL system to do that. But again, it's simply not needed. 

The ingame scripting system is already whitelisted. You can't get "server object references" - provided I understand what you mean by that. It's perfectly safe, it can't be misused in a really dangerous way. The only misuse that can be done today is getting hold of in-game block references you shouldn't, but that can be remedied easily enough. If you're talking about the ingame hacking incidences, that's the mod API, not the in-game API. That can also be remedied, we know why and how that is possible. But none of these APIs give you access to any really dangerous objects, you don't have access to anything outside of the game itself.

 This is a game. You need to pull as much performance as possible. A programming block can be run every single tick, every single frame. You don't have much to go on there especially since you can assume that the scripters aren't exactly performance aware. Second, Keen is a fast-paced development company. You bet implementation simplicity and implementation speed is a proper concern.

t3hk0d3:
@lord-devious you really underestimate how cunning humans can be, and seriously underestimate power of reflection api.

lord-devious:
I have many years of experience with .Net. I'm fairly sure I know everything that can be done with reflection. But: You don't have access to reflection in neither ingame scripts nor the api scripts.

t3hk0d3:
@lord-devious You can't guarantee there is no sandbox escape exploit existing, will exist in future version of .NET, or such possibility will be added accidentally (or not) by some mod maker, running priviliged code.

There is tons of such exploits for Java, for example.

MeusRex:
Nore does an additional layer. (As blizzard proofed with a wc3 bug that persisted for years and enabled coders to run machine code on a player's pc.)

Absolute security is a myth.

lord-devious:
@t3hk0d3 Mod makers can't run privileged code either, so they can't provide access to privileged code. The only mod makers that can do that are those who run illegitimate versions (source code modifications) of the game, and those who play on those can only blame themselves.

I feel rather confident in claiming that a .NET based sandbox will be every bit as safe as any non-.NET based sandbox, yes. Absolutely.

PR #407, opened 2015-08-25
Added method to IMyBatteryBlock to set current stored power

Jimmacle:
This feature allows modders to set a battery's current stored power through script mods via IMyBatteryBlock.SetCurrentStoredPower(float power). It will not allow the charge to be set less than 0 and will cap at the battery's maximum capacity. It adds an IMyBatteryBlock interface in Sandbox.Common.ModAPI while keeping the original interface in Sandbox.Common.ModAPI.Ingame.

If anyone's interested, I did this to enable Darth Biomech's Mass Driver mod that I'm programming to use a charge mechanic. This change would also be useful for shield mods now that the damage system is accessible.


Jimmacle:
Closing this because I messed up my repo, will open a new PR later

Issue #411, opened 2015-09-04
Loading worlds with large mods can be very slow


alex-davidson:
This issue came to my attention via https://www.reddit.com/r/spaceengineers/comments/3jf20e/why_does_se_reads_320gb_of_data_during_loading_a/

The test world was published here: http://steamcommunity.com/sharedfiles/filedetails/?id=511660817

The I/O rate mentioned in the Reddit thread does not appear to relate to actual disk access; the operation is not even remotely bottlenecked on the disk. However a CPU core is completely saturated for a very long period of time, with high page fault delta and 'I/O read bytes' (presumably hitting disk cache or another device?)

I have reproduced the problem using the current retail build of the game (1.098) which took > 40 mins of CPU time to start. I then modified the version constant of the GitHub source (MyFinalBuildConstants.cs, line 18) to 01094012 in order to load the game world, and tested a 'Release' build. Given the long load times I did not wait for it to complete but the I/O and CPU patterns matched those seen before, so I have made the assumption that the codebase behaviour is sufficiently similar for an analysis to be valid.

Over a 20 minute period, a profiler attached to a Debug build of the GitHub code revealed the following:
* ~7200 seconds wall time in 'waiting for synchronisation' within ParallelTasks.FixedPriorityScheduler, presumably idle task threads
* ~1100 seconds wall time in RenderThreadStart, probably updating the 'loading' screen.
* ~1100 seconds wall time in MyDefinitionManager.LoadDefinitions, of which:
  * ~870 seconds in LoadPhase3
  * ~170 seconds in LoadPhase5

LoadPhase5 seems to spend a lot of time in XML-handling code. This is probably skewed by the profiler and the Debug build.

LoadPhase3 spends a long time in MakeBlueprintFromComponentStack (~520 sec) and eventually MyDefinitionManager.ProcessField (~520 sec), and also ToDefinitions (~350 sec) leading again to ProcessField (~350 sec).

These two ProcessField invocations consume about 350 sec and 65 sec respectively (~415 sec, ie. about 40% of the load time).
According to a call graph for this method, 66% is a Count() invocation and 33% is a DirectoryExists check. There's another '15%' which is recursion.

[...]

PR #414, opened 2015-09-08
Avoid unnecessary filesystem tests on the hot path.

alex-davidson:
When loading mods, we need only check file existence if we did a fixup on
the file path eg. applying a file extension when one is missing.

refs #411


alex-davidson:
Filename minus extension might not exist. This patch is invalid.

PR #419, opened 2015-09-27
Add limited programmability to IMyJumpDrive

leftler:
This PR adds the ability for programming blocks to do the following things to a jump drive

**Read-only properties and methods:**
* `CanJump` - Get the status if the jumpdrive in a state that would allow a jump.
* `IsFull` - Get the status if the jumpdrive is full of energy.
* `IsJumping` - Gets if the drive is currently in the process of jumping.
* `JumpDistanceRatio` - Get the % of power the jumpdrive should jump.
* `PowerNeededForJump` - Get the amount of power that is required to perform a jump.
* `StoredPower` - Get the amount of power stored in the jumpdrive.
* `RechargeTimeRemaining` -Get the number of seconds remaining in the recharge time.
* `Recharging` - Get if the jump drive allows recharging.
* `ComputeMaxDistance()` - Get the maximum distance the jumpdrive will jump with the current JumpDistanceRatio.
* `CanJumpAndHasAccess(long userId)` - Get if the drive is ready to jump and could be jumped by the specified user.

**Writeable methods**
* `RequestJumpDistanceRatio(float ratio)` - Set the % of power the jumpdrive should jump.
* `RequestRecharging(bool enabled)` - Sets the value of Recharging.
* `RemoveSelected()` - Deselects the jump waypoint if one was set.
* `SetTarget(Vector3D cords, string name)` - Sets the jumpdrive to jump to the passed Vector3D, it uses `name` to set what will be displayed on the GUI.

This ***does not*** allow a user to perform a jump programmatically, it only allows for exposing or modifying information that was already available through the K menu.

PR #420, opened 2015-09-28
Npcs can perform a Jump on a Jump Drive.

leftler:
This PR is a superset of the changes in PR #419.

This adds a `RequestJump(IMyJumpDrive jumpDrive)` to `IMyRemoteControl` in similar fashion to `GetNearestPlayer(`. This method will only succeed if called from a remote control block that is owned by a NPC.

It also adds `GetJumpCoords(IMyShipController shipController)` to `IMyJumpDrive` to find the target location of the jump, it uses the `shipController` to determine direction of a bind jump. 

This PR will likely need to be updated once the changes to `MyRemoteControl` which are live but not here on the repository, but I wanted to put this out there now in case people wanted to comment on it.


leftler:
I messed up my branch when trying to get it up to date. I am just going to remake this with the latest.

PR #421, opened 2015-10-04
MyFixedPoint.cs improvments

DepthOfSight:
1. Fixed rounding operation of MyFixedPoint.cs

2. Reduced the number of operations to perform multiplication
- the fixed point value is basically a float value shifted by some number of decimal places( * 10^X).  
A = f1 * 10^X
B = f2 * 10^X
A * B = (f1 * 10^X) *(f2 * 10^X) = (f1 * f2) * 10^X * 10^X
As the 10^X is only to meant to shift the value into the integer range, there should only be one in the result.  So you can just multiply the two integers and divide by the Divider value
A* B / 10^X = (f1 * 10^X) *(f2 * 10^X)  / 10^X= (f1 * f2) * 10^X * 10^X  / 10^X = (f1 * f2) * 10^X 
"(f1 * f2) * 10^X" is the desired result as it is the float value shifted by the desired number of decimal places.

So MyFixedPoint multiplication should be (A.rawValue * B.rawValue) / Divider


DepthOfSight:
I'm closing this pull request as my "improvement" of the multiplication operation of the FixedPoint class will overflow in large value cases.

PR #426, opened 2015-10-08
NPC jumpable drives

leftler:
This request gives remote control blocks the ability to initiate a jump. This function can only be called by remote control blocks owned by NPC's, the same way GetNearestPlayer can only be called by a NPC.

This pull request depends on changes made in #425

MyGridJumpDriveSystem changes:
- Refactored logic out of RequestJump( into RequestJumpCalculation(.
- Added RequestJumpWithoutPrompt( which tries to perform a jump request and returns a boolean if the request was sucessful or not.

IMyRemoteControl and MyRemoteControl changes:
- Added RequestJump( which tries to initiate a jump sequence, this method can only be called remote control blocks owned by NPCs.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyJumpDrive.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyJumpDrive.cs
@@ -2,10 +2,93 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using VRageMath;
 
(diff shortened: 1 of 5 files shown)


leftler:
Crap, I found a error that snuck in when I merged in changes to get the latest version. I am going to cancel this PR for now and re-open it once the source is compileable again.

leftler:
It was easier to fix that I first feared, I just had to put the gravity check in both request jump methods.

PR #438, opened 2015-10-17
[Feature] The Programmable Block Storage Property is a Json Token

lord-devious:
* NOTE * This should now be ready, but the "files changed" reported is just nuts. Please verify this before doing any kind of merge! I didn't delete those files... 

Thanks to @Chrison for this idea - I can't *believe* I didn't think of this myself. It's so simple.

This feature changes the Storage property into a JToken by Newtonsoft (http://www.newtonsoft.com/json). You no longer need to do any kind of serialization yourself, and the game is capable of serializing your data on demand and only when actually needed as opposed to every time your program runs.

http://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Linq_JToken.htm

This pull request is *slightly* breaking, but to make any existing script working you should only need to change any read references from this.Storage to (string)this.Storage and it should work nicely.

```cs
    // Change
    var myStoredString = this.Storage;
    // to
    var myStoredString = (string)this.Storage;
```

In many situations, however, it would be far more efficient to initialize your storage and use it as specified on Newtonsoft's site.

```cs
    if (!(this.Storage is JObject)) {
        this.Storage = new JObject();
    }
   this.Storage["MyProperty"] = "My String";
   this.Storage["MyProperty2"] = 15;

   var str = (string)this.Storage["MyProperty"];
   var no = (int)this.Storage["MyProperty2"];
```

Again, check Newtonsoft's site for more advanced features. Let me know if there is something that doesn't work. Some of the features may need to be excluded on purpose though.


leftler:
Very cool! I did a quick double check of all the new whitelisted namespaces, there is nothing in there that can break out of the sandbox (No constructors that will generate a Reader or Writer from a file path, everything takes in a existing `TextReader` or `TextWriter` only). I hope they put it in.

lord-devious:
@leftler Thanks, a double check never hurts :smile:

JCMais:
I really hope this is merged when Keen get back to the source code, since looks like they forget it for awhile when they are finishing planets.

JustBurn:
Instead of breaking the current implementation which could cause issues why not just expand it?

```cs
var myStoredString = this.Storage;      // Always string type (i assume?)
var myStoredJSON = this.StorageJSON;    // Always JObject type

myStoredString = "MyLegacyHelloWorld !";
myStoredJSON["MyProperty"] = "My String";
```

This way there's no need for sanity check " if (!(this.Storage is JObject)) { " either.

lord-devious:
Because it's better to replace it, leaving a single point of access and cleaning up the garbage code. Because it will force people to consider updating to a far more performant way of storage, because they no longer need to serialize stuff each and every run. Because this is still an early access game and such small breaking is no big deal. Because the storage string hasn't really been much in use until now. How do I know? Nobody has complained much about a bad bug which made the storage string reset to its last saved state every time one manually compiled the program. That has now finally been fixed though.



Because I believe and hope the programmable block will be rewritten in a way that will break far more scripts anyway, because it's not sandboxed enough.

JustBurn:
To tell the trust i never heard of "Storage" and "Me" until i actually started messing with the source code to implement the features I've wanted long time ago (Started doing this recently).

Available documentation has no reference to either of those and any talk about them is very deep buried in the forums (really hard to find).

In fact I've even wrote " Implement 'self' in programmable block " and " Special variables that get loaded/saved in world " in the list of features that time, I'm glad that most of them were implemented by the community and most are in the official game.

There still 2 features that weren't implemented by the community which I've implemented on my private build, i've been using one of them all the time, the other one is equally useful for buildings that aren't just flying ships... like a crane, airplane, mecha or something like that.
They aren't public yet so far... I don't know if someone interested anyway because looking back in the suggestion posts, @lord-devious was the only one that did reply in one of them (by the time Keen didn't had released the source code yet)... anyway i think I'm getting too much off-topic so sorry about that ;)

lord-devious:
I *am* lord-devious and the author of most of the extra PB features, except Storage which is Keen.

JustBurn:
i've wrote @ lord-devious, it may show up differently for you since it's referencing you ;)

lord-devious:
(shrug) No matter. 

Anyway, my experience has taught me that legacy code often causes *a lot* of problems. Which is why I prefer to clean up when I can. And I've been doing this for a very long time - I have learned to trust my instincts by now :wink:

In the end it's Keen's decision. If they want me to make this 100% backwards compatible, then I will do so - with regrets.

JustBurn:
Whatever it's decided I'll rather manipulate variables in nodes rather than packing/unpacking data into a single string, so i clearly support this pull. :+1:

Even through i rarely play the official build anymore (I can't live without the features I've added in my build), I'll still patch on my personal build, so, many thanks :smile:

Evrey:
Interesting feature. Combined with PB-to-PB communication (on which I'm currently working on) this enables players to build some kind of simple data server, enabling multiple PBs to solve one big problem cooperatively.

lord-devious:
@Evrey I'm sorry, but not really. The Storage property already exists. It's an internal feature for each programmable block. It's not accessible outside of that single program. All I'm doing here is change it to a JToken to make its use easier and more performant. Honestly I'd prefer it to be a preinitialized JObject but I fear that would be _too_ breaking.

Actually it should be a property on the encapsulating programmable block. That would be interesting and could indeed be used the way you suggest.



Hm...

Evrey:
@lord-devious I think, you misunderstood me. What I was thinking of is something like this:
`int sender_id;`
`string message = GetPbMessage(out sender_id);`
`if(message != null) {`
`  SendPbMessage(sender_id, Storage[message]);`
`}`

So the idea is not to randomly access the internal storage of other blocks at any time but to request data transfer by sending and receiving messages. This also gives "server blocks" a chance to reject undesired requests.

lord-devious:
@Evrey If so then you can already do it, the Storage property already exist as a string. You just need to do the serialization yourself.

Evrey:
That's true, but the point I wanted to make is that your feature here significantly eases doing such things, making them simpler and more robust.

lord-devious:
@Evrey Point taken :smile: that was indeed it's purpose.

lord-devious:
Stand by; making an improvement as requested...

lord-devious:
Ready but the commits must be verified because they look just weird.

LeonidasCZ:
Yes please commit again without the deletions in Effects2 (ignore localy if they keep bugging in)

lord-devious:
Finally managed to clean up the repository so it reflects actual changes. However while I was able to make the corrections, I am not in a position to test whether these fixes broke something until later this evening (it's now about 9am, at the time of writing), so I guess I'll have to wait another turn.

Phoenix84:
I don't suppose you could take this and apply it to all functional blocks?
With the custom terminal controls coming, this would be a great way to make sure custom controls can be saved with the world with minimal effort on the modder.

lord-devious:
@Phoenix84 Not in this first iteration at least, I don't want to add any more complexity to this one. Good idea though! It should be fairly simple to extend to other blocks once this one is in and the standards accepted. One could perhaps consider to put the PublicStorage all the way down into the functional block base class to be inherited by all, this would be incredibly useful for ingame script metadata as well.

By the way - it's tested and ready to be merged now :smile:

lord-devious:
To be readded later

PR #448, opened 2015-11-06
[FIX-LOW] Allow larger scale explosions

justburn:
Bugfix, Low priority - *(Cannot be exploit in vanilla.)*

The original code generate an integer overflow if the volume result is over 1290^3 causing the value to be negative and the program taking the wrong path that freeze the game for several minutes or hours, the patch will expand this limit to 1,664M ^ 3 ( This should obliterate the entire universe xD ).

Additionally, this patch increases the physics raycast recursion count limit (from 10 to 100) in case the huge explosion ray cross several levels of layers in grids. This was only tested with an explosion as big as an asteroid on the platform base so i'm not 100% sure if the value 100 is big enough for colossal planetary explosions (Death Star superlaser, anyone?).

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Cube/MyCubeGrid.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Cube/MyCubeGrid.cs
@@ -5693,7 +5693,7 @@ namespace Sandbox.Game.Entities
             var localSphere2 = new BoundingSphereD(center, sphere2.Radius);
             var localSphere3 = new BoundingSphereD(center, sphere3.Radius);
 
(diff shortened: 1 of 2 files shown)


JustBurn:
Tested today explosions on alpha planets with x64 build. Results of the experience weren't good.
A volumetric explosion may require allocation of memory depending of the object it intersect, If they are asteroids then it's okay, if is large planets then is not okay.
If the area of the explosion is really big on a big planet surface, the working buffer (in volumetric size) can get really huge, OS or .NET doesn't like to give the game a single massive block of several gigabytes so it results in OutOfMemory easily (alloc of 1.5GB can throw it even if the system has over 4GB free).

Colossal non-volumetric explosions do work (despite being oddly slower than volumetric)... so... there still a possibility from there.

PR #449, opened 2015-11-09
Added GraphicsRenderer field for IMyConfig

THDigi:
This populates the graphics renderer (DX9, DX11 or NONE) read-only through *MyAPIGateway.Session.Config.GraphicsRenderer*.

Right now it's impossible for a script to know what renderer the client is using.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/IMyConfig.cs
+++ b/Sources/Sandbox.Common/ModAPI/IMyConfig.cs
@@ -2,6 +2,7 @@
 using VRage;
 using VRage.Serialization;
 using VRageRender;
+using VRage.Utils;
 
 namespace Sandbox.ModAPI
 {
@@ -32,6 +33,7 @@ namespace Sandbox.ModAPI
         int RefreshRate { get; }
         bool RenderInterpolation { get; }
         MyRenderQualityEnum? RenderQuality { get; }
+        MyGraphicsRenderer GraphicsRenderer { get; }
         bool RotationHints { get; }
         int? ScreenHeight { get; }
         int? ScreenWidth { get; }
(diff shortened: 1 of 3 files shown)


leftler:
I would make the fallback state "Unknown" instead of "NULL"

harag-on-steam:
Is returning `NULL` or `Unknown` really better than making the property optional? Other properties in this class do exactly that. See `RenderQuality` directly above your new property for an example.

On the other hand, why plan for this eventuality at all? What is the caller supposed to do if the return value is null? It is highly unlikely that the renderer will ever be unknown. You could even call it an error if it ever is.

mloftis:
I'm coming from left field knowing little to nothing at this point...but the renderer might not get set if you inquire about it in say a dedicated server.  That's where I could see a NULL personally.

harag-on-steam:
I guess `NONE` would be less ambiguous to describe that situation.

Spartan322:
`NONE` would be more appropriate for a system missing a renderer, cause `NULL` leaves ambiguity, as does `UNKNOWN`, it all cases, Space Engineers only supports `DX9`, `DX11`, or `NONE`, so the renderer should never be treated as if unknown.

THDigi:
Alright, `NONE` it is.

Spartan322:
Sorry to be picky, but it appears that to be consistent, enums need an `Enum` suffix.

mexmer:
@harag-on-steam null renderer is returned on server, although there are better methods to check, if you are on server

LeonidasCZ:
Sorry it didnt conflict but still didnt compile against the new code can you update and reopen? Thank you :)

THDigi:
@LeonidasCZ Sure, fixed and PR'd https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/479

PR #453, opened 2015-11-18
Radio Antenna data transfer using Programmable Blocks

Evrey:
# What is this all about?

Greetings, guys!

This is a rather big patch I've been working on. It adds a lot of new functions to the existing `IMyRadioAntenna` interface, which implement the following features:

- Programmable Blocks can use the `IMyRadioAntenna` interface to transfer a pack of data from one ship to an other. Obviously, received data packs can be queried using the very same interface. 

- Programmable Blocks have to enable this feature per antenna, so that antennas not making use of this patch have no runtime or memory penalties.

- The whole antenna patch is balanced and safe. You can only query or send stuff within radio range. You cannot take control over the target antenna's blocks or ships by using the `IMyRadioAntenna` interface.

- Depending on the distance to the target antenna, you can query different kinds of information about the other antenna's ship, like it's mass, position, and in case that showing the ship's name is enabled this very name.

- All pending messages in antenna data queues are being saved, so that no data is lost when loading/saving your game.

- You can transfer and query data between antennas that have broadcasting disabled. Thus, you can disable broadcasting in your complicated sattelite network without breaking data connections, to free the screen from gazillions of HUD markers. To be "invisible" for antennas, you'll have to power-off your antenna completely.

- This *should* work bugfree in multiplayer, too. Tell me, if you experience any bugs.

- You can inspect the play time at which a message has been sent, to e.g. drop messages that are too old to be of any relevance or to measure the time between sending a message, waiting for it to be handled, and receiving the corresponding answer.

# The Code and an Example

The code is well documented and I did my best to mimic Keen's programming style. Tell me, if any piece of documentation is unclear. I'd consider this being a bug to fix. Here is a very short example of how this new API works:

```c#
var antenna = (IMyRadioAntenna) GridTerminalSystem.GetBlockWithName("STEVE");
antenna.DataTransferEnabled = true;

// Send data to an antenna with the ID 42.
bool success = antenna.SendToNearbyAntenna(42,"Hi there!");

// Send data to every antenna within your radio range.
antenna.BroadcastToNearbyAntennas("SOS!");

// Receive data. The sender's ID will be stored in `sender`
long sender;
string msg = antenna.GetReceivedData(out sender);
if(msg != null) {
  // Handle received message.
}

// Receive data with a time stamp.
TimeSpan dispatchPlayTime;
msg = antenna.GetReceivedMessage(out sender, out dispatchPlayTime);
if(msg != null) {
  var dt = MyAPIGateway.Session.ElapsedPlayTime - dispatchPlayTime;
  // Handle received message and "message handling lag" `dt`.
}

// Query the nearby antenna's position, if in range.
var targetPosition = antenna.GetNearbyAntennaPosition(42) ?? Vector3D.Zero;
```

# Why have such a Feature?

This is just one key piece of a larger patch including many features. All together they form the basis to create a fully automated fleet, where some computer ships can send instructions to nearby drones and query reports from those drones when they come back after work. One or even 8 players are just not enough to manage a whole fleet in real-time. ;3

Someone in the forums created a script that transfers strings using... laser antennas, I think. However, this script has some major drawbacks in comparison to this patch:

- Data transfer is very slow, as there are long pauses between transfers. Transfer speed in my patch only depends on how fast you can repeatedly run your programmable blocks.
- It exposes the transferred data through the antenna's name. Privacy, anyone?
- It is a dirty hack requiring quite some complicated code in your programmable blocks. However, I do admit that there is quite some ingeniouity behing this hack.
- You cannot query additional information about your communication partner, which would be essential for an automated fleet.

For these reasons I believe, that this patch would be an awesome and very useful addition to vanilla Space Engineers.

# Final Words

If you spot any bugs, please tell me. Also tell me if something is unclear or if something in the code - e.g. naming schemes - are inconsistent. You may also suggest features.


SigmaStrain:
Hello! I'm the user from the forums who wrote the "dirty hack" as you called it. Great code. I would hope they add something like this soon as it would make for much better and more fluid automation within the game. Keep up the good work.

MeusRex:
Maybe add the ability to run a specific PB that is located on the same ship as a specific antenna. (With the msg you send as the argument for the specified PB. This would remove the requirement for a null timer to promptly catch all messages. But be wary of the ownership.)

Evrey:
Greetings!

@SigmaStrain Wait until I tested all my patches. Automation will never have been easier. °^° I'm still testing PB messaging and PBs inspecting an ore detector's findings.

@MeusRex By the time I started writing my answer, I summed up quite a lot of arguments against such a design, but now I'm getting an idea of an elegant solution to both, your request and my complaints. This could result in completely different but possibly more efficient code. One PB will have to register itself as an antenna's backend and will be called with the incoming message as string. I'd have to put a getter somewhere to receive the sender's ID (no useless stringly typing/parsing) or `null` if not being called from an antenna.

Evrey:
@MeusRex I gave it a quick thought and tested something similar. I came to the conclusion that it's still best to poll messages using a "null timer", and I have several reasons for that:

- Someone here worked on a self-repeatable PB patch, which - in my opinion - does too much, but it works. I, too, made such a patch, which is way simpler in it's design, but is not fully tested yet. Both spares us the "null timer".
- Many of us at least use this inventory status script from the workshop, which, too, requires that "null timer". You can easily make this timer loop more PBs, which still requires just one timer.
- The coupling between PB and antenna would become too tight, IMHO.
- If a received message automatically results in calling a PB, I'd have to delay this call in a way that distributes the execution load across multiple frames. Otherwise my patch will cause heavy frame drops when sending a lot of data in lots of messages. By polling, however, the PB user has to think about distributing the load by himself. Otherwise, his function will abort due to a timeout exception.
- And that's the main reason: Calling a PB per message gives each call the whole time slot, whereas a single looping PB gets only one time slot for as many messages as the PB user is willing to handle.
- And even if I - say - delay each call by one frame... how do I handle pending messages before a PB came by saying "Hey, dude, tell me what ya've got!", or how do I handle PBs that have been shut down for a while? Drop old messages? How many? Yeah, a PB user could make early returns, but this wouldn't stop the one-frame-delay-per-I-don't-care-message. And who am I to decide which messages to drop before the PB is called?

However, I just found a possible exploit in my design as it exists right now, and I'm not sure whether it is sensible to fix this. It's in fact a design flaw of the entire sandbox API. And here's the thing:

- Build a computer ship with... merge blocks. Or use the screw-away-that-connecting-pipe-trick.
- On this connected grid, build an antenna.
- Grab the antenna by script.
- Disconnect the grids and drop the antenna somewhere.
- Et voilá, you possibly have an antenna with data connection enabled, and you can receive messages within this antenna's range, no matter how far away it is from the actual PB handling the messaging.
- This should work with all kinds of blocks.

So, here's the problem. Should I fix this? The fix would be ugly static functions taking the ID of the antenna you want to work with.

But there is one thing I will do: Add timestamps to the messages. Just got the idea writing this comment.

LeonidasCZ:
GetBlockWithName should return you wrapped object that is invalidated when it leaves your grid, not done atm and will be probably solved inhouse.

Can you update?

foobarto:
This feature would be a game changer! Please consider merging this PR.

Icecream-burglar:
I would prefer an API that mimics actual data transfer. Like mirror a stream or the system.net.sockets.tcp sockets.

PR #478, opened 2016-02-27
Programmable Block Assembler Control API

computerneek:
Hello, KSWH Developers.  I was rather irritated when I found I had to have something like 40 assemblers working in concert to be able to have programmables ensure the perfect counts of everything, with assembly & disassembly as needed.  So I wrote this:  An API for programmable blocks to access the assemblers.  The programmables refer to everything by name, and cannot change any blueprints or anything; they are basically given the same controls as can be found in the production tab, with the added bonus of being able to search for available resources.  However, due to how little I know about Space Engineers code, there are a few modifications you may wish to make before incorporating this into the game:
1.  MyAssembler.BlueprintName()   Perhaps there's a better way to get the name of the blueprint that might return something like "Assembler", "AirtightHangarDoor", "Component/SteelPlate", "Component/LargeSteelTube", etc. for the associated objects?
2.  MyAssembler.BlueprintByName(), MyAssembler.ComponentByName() and MyAssembler.ToolByName()   There's probably a better way to get the components and blueprints by name- goes in conjunction with BlueprintName().
3.  MyAssembler implementations of IMyAssembler.GetResources() and IMyAssembler.GetAvailableResources()    I don't know how to check if the assembler is actually set to use conveyors before looking down them.  It only looks; as it is, the programmable block will always be told it has all resources the player could conveyor-send in, but the assembler will not go fetch them if it's not set to use conveyors.
4.  MyAssembler.GetBlueprintResources() and MyAssembler implementations of IMyAssembler.GetResources() and IMyAssembler.GetAvailableResources()    The three functions above use MyDefinitionId.ToString() to report resource names to clients.  Perhaps something else that just returns "ingot/iron", "ingot/gold", etc?
5.  When I was designing the original API, there was a little bit in the refinery too- not control, just status- like what ore is used to make an ingot (and vice versa), how long it takes to burn X amount of ore or make X amount of ingot, how much ingot is made by X ore (or ore used by X ingot), and what resources the ore in the refinery will produce.  I was not able to find ways to do that, but I know there is one.  I also had the MyAssembler implementation of IMyAssembler.GetAvailableResources() using such functions to add future refinery outputs to the availability list.


Spartan322:
This is C#, so why create argument-less getter functions when you can use field getting instead?

computerneek:
@Spartan322, thank you for your comments.
For the first one, about the getters, it is probably because I'm used to Java- which simply doesn't have properties.
For the second, I have it using lists of names because the actual object is not whitelisted for scripting.  If I used the actual object, the ingame script using the API would not be able to call these functions at all.
For the third, this is more documentation than I usually do, and less than I like.  It's so documented here because I designed the functionality before I had even started installing Microsoft Visual Studio, let alone downloaded the source to design the API- thence, I started with only documentation.  As for the function newlines, that's just my style.
I'll fix the getter-vs-property thing and the newlines thing right away, but the object-vs-name thing is a game limitation, and I honestly don't see a _problem_ with documentation...  It just helps someone understand- like the KSWH developer who goes in to tune my code down the road, or the confused ingame-script-writer looking to automate their assemblers.  I can remove it, though, if you can convince me to...

Phoenix84:
RemoveQueueItem is server-side only. You need to use RemoveQueueItemRequest.

Aside, I am already adding this functionality myself (merge pending). However yours adds much more information, such as blueprints and component lists.
Mine just adds the ability to queue and dequeue items, and see the queue.

computerneek:
Interesting comments.......    The script also runs server-side, meaning all these functions will only ever be called server-side.  I tested this code; if I don't add the OnQueueChanged() calls to notify the client of the changes, the client thinks nothing changed!

Spartan322:
Opps, forgot about that limitation, honestly, I find it a tiny bit dumb to have to use strings, but good enough here.

mexmer:
@Spartan322 is indeed correct, your comment will not be processed by sandcastle, which is used for producing documentation. Use <sumary> notation for commenting interface.
see https://msdn.microsoft.com/en-us/library/z04awywx.aspx for details regarding how to use microsoft xml documentation

Phoenix84:
Remember, anything added to the PB is also accessible in the ModAPI, which _can_ be run on the client. :smile:

computerneek:
@mexmer I was thinking more about someone reading the source code, but I will adjust it to work with this 'sandcastle', thank you.

@Phoenix84 Thanks, I wasn't aware of that...   I'll have to look into the sync thing and adjust all queue modifications to go through them.

Phoenix84:
@computerneek The minimal sandcastle implementation is easy:

In VS, above the method, property, or whatever, type three /
`///`
With VS autocomplete, it'll automatically add an XML summary tag. In there, just write a brief description of what it does (one line). Fill in any parameters if any. Ideally the return too, if there is one.
More detailed data can be added in a <remarks> tag, but that isn't automatically added.

computerneek:
@Phoenix84 Thank you, I wasn't aware.  The find-and-replace feature was able to turn my comments into a basic rendition via <remarks> tags for me- before I saw your post.

Anyways, I have completed the multiplayer/sync modification...   However, because Keen's sources won't compile for me (a couple symbol-not-found errors in a couple of the VRage projects?) I was unable to test it.

mexmer:
using <summary> is for microsoft XML documentation feature.
it's processed not only by xmldoc processor (part of build), but also by visual studio intelli sense.
output of xmldoc can be used by external docummentation processors (like sandcastle, which is used by KSW to produce programming api help document)

LeonidasCZ:
Hi, computerneek thanks for the PR, we are actually working on this internally and this helps us improve the interface to contain more of access required by community.

Just for your future reference there is couple of advice.
Passing strings around might be quite inefficient, consider passing whole MyDefinitionId if you can or MyStringId/MyStringHash located in VRage.Utils, this will make checks on scripts side much easier and faster. Second is making new list in lot of methods, when you can just let the caller pass you the list he wants to fill like in GetQueue. In case its internal list consider making it private member reusing it in all of the functions.

Thanks for your contribution and hope to see more :)

PR #479, opened 2016-02-27
Added GraphicsRenderer field for IMyConfig again

THDigi:
Fixed and re-added as requested.
https://github.com/KeenSoftwareHouse/SpaceEngineers/pull/449


znk:
@THDigi Marek said, on his blog[1], that KSH is gonna drop Dx9 support by the end of March for SpaceEngineers. 

[1]: http://blog.marekrosa.org/2016/02/space-engineers-news-full-source-code_26.html

THDigi:
I know, but if they implement this before they end support then it can be useful for mods that chose to support older versions to determine what renderer they're using.

mexmer:
like i pointed out in #449, it's still needed, because DS doesn't have renderer. there are other methods to detect you running mod on server, but still if you base checking what you do, just on renderer presence, this will help.

PR #483, opened 2016-02-28
Fix: Loading random blueprint crashes the game if no blueprints match the search pattern

Jimmacle:
This PR fixes the IMyProjector.LoadRandomBlueprint(string searchPattern) method throwing a divide by zero exception when no files match the search pattern by checking if the list of files is longer than 0 before picking the random index and loading a blueprint.


LeonidasCZ:
Already fixed on trunk thank you

PR #486, opened 2016-03-03
Grid physics info for ingame scripts

stanhebben:
Extends the ingame IMyCubeGrid interface to provide physics information. This makes it much easier to write PB scripts to accurately control ship motion - such as cruise control, automining, autodocking...

The following properties are available: mass, center of mass, linear velocity, angular velocity, linear acceleration and angular acceleration. Each of these properties are available both relative to the grid and relative to the world.


Phoenix84:
I think it would be better if an IMyGridPhysics interface was created and MyGridPhysics implemented that, then add the getters there. IMyCubeGrid.Physics could then be of type IMyGridPhysics instead.

It keeps the division between grid and physics, user can still check for null themselves, instead of you adding it in each property.

lord-devious:
I agree with @Phoenix84. Should be simple enough to provide a wrapper.

That said, good work, @stanhebben! Very useful.

stanhebben:
An IMyGridPhysics interface makes sense. In fact, I'd rather call it an IMyPhysics object. Grids aren't the only things with physics, and perhaps at some point we might like to know the physics of an entity, a block...

stanhebben:
You mean having both an ingame and a modapi variant, even if both would provide just the same functionality?

mexmer:
you don't need to define both as whole, inheritance FTW.

stanhebben:
Yeah, sure, it is easy to make. I was just wondering if it would have any meaning at all. But thinking about it, it makes perfect sense with the structure as we have it with ingame / modapi, so I'll add it.

Phoenix84:
@mexmer Keen themselves don't even do that, though I'm not opposed.
@stanhebben if you do that, it can just be an empty interface.

Sources/Sandbox.Game/ModAPI/Blocks/MyCubeGrid_ModAPI.cs
```diff
@@ -259,5 +259,135 @@ Sandbox.ModAPI.Ingame.IMySlimBlock Sandbox.ModAPI.Ingame.IMyCubeGrid.GetCubeBloc
...
+
+    internal class GridPhysics : ModAPI.Ingame.IMyPhysics
+    {
+        private MyCubeGrid grid;
+        private MyGridPhysics physics;
+
+        public GridPhysics(MyCubeGrid grid)
+        {
+            this.grid = grid;
+            this.physics = grid.Physics;
```
Spartan322: In this case, saving physics might be pointless.

stanhebben:
Investigating the code, I can't simply make a ModAPI version of IMyPhysics. Sure, I could make a subinterface of the ingame IMyPhysics, but I'd have to provide a Physics property on ModAPI.IMyCubeGrid - and it already has one from IMyEntity.

For the ingame version, this doesn't matter, since those Physics (and IMyEntity) aren't accessible for ingame scripts. However, doing this on the ModAPI class would likely break some existing mods (and perhaps even the game).

Alternatively, would could pick a different name for the property. But since the current Physics property is already available, this seems unnecessary to me.

mexmer:
@Phoenix84 when exactly KSW did add any new interface to PB?
if you look at modapi/ingame, there is no ingame interface, that has not it's modapi part.
modapi part sometimes have same functionality, sometimes has setters (where pb has only getters), and lot more functionality. so you are wrong about "keen itself is not doing that", they did it before.

also i explained that before to others, it's also about having all in same namespace and extensibility.
not to mention, it's makes reusing pb scripts in mods much easier.

@stanhebben if there is same field in modapi, then you just override definition from ingame, that's again inheritance, it allows you to have in child class redefinition of field from parent.

Phoenix84:
@mexmer 
IMyAssembler
IMyTextPanel
IMyShipGrinder
IMyCockpit

I could go on.
Please don't tell me I'm wrong unless you know for a fact.

stanhebben:
@mexmer I know I can override properties in a child interface. The problem is that everyone so far expects IMyCubeGrid.Physics to be a MyPhysicsComponentBase. If that suddenly becomes an IMyPhysics, things will break.

In fact, my own mod would break as well.

Phoenix84:
@stanhebben It looks like MyPhysicsComponentBase declares a few of the properties you want.
If it has all of them, you could make that itself inherit from the interface.

mexmer:
@Phoenix84 ok, my appologies, indeed there are some missing i didn't check every interface.
some of them are community additions

anyway reasons for adding modapi interface are still valid.

stanhebben:
MyPhysicsComponentBase contains everything I need, but it is part of VRage. I can't (and shouldn't) be using types from Sandbox.Common in there - that would break their architecture: Sandbox.Common depends on VRage, but not the other way around.

I could make ModAPI.IMyPhysics part of VRage, but then Ingame.IMyPhysics would have to be part of VRage as well (in order for one to extend the other). However, there is no ingame API in VRage, and I am not sure if they plan to.

Another alternative would be to have MyGridPhysics implement IMyPhysics. But then IMyCubeGrid would have to return a MyGridPhysics object instead of MyPhysicsComponentBase - but MyGridPhysics isn't whitelisted.

Another, simpler alternative would be to create a property with a different name, such as PhysicsData, Physics2 or PhysicsInfo. That would avoid these problems altogether, but it would be less pretty.

From a software architecture point of view, exposing physics as IMyPhysics instead of MyPhysicsComponentBase would make the mod API cleaner and provide more control about what is and isn't available to mods. But it would break compatibility and this doesn't just seem worth it.

So either we create a new property with a different name, or simply have no ModAPI implementation for now.

PR #487, opened 2016-03-06
New Turret Type for Modding - Large Generic Turret

CMDBob:
Adds a new type of turret, LargeGenericTurret. It's a simple interior style turret, but with support for conveyor networks. Perfect for turrets that aren't gatling guns or rocket launchers, such as autocannons, tank guns and such.

Internally, the models use the same structure as the interior turrets, but named differently. The subpart for the turret body is GenericTurretBase1 and the subpart for the barrel(s) of the turret is GenericTurretBase2. Everything else behaves as a Space Engineers turret behaves. The main change was setting the base class from LargeTurretBase to LargeConveyorTurretBase. It's been tested and works.

It's more of a modding turret, as I was fed up with the limitations of the existing turrets. Gatling turrets require an extra model that you don't see with fixed barrel turrets, and the aim points for the turrets jitter; Rocket turrets have the "Reloading..." thing which isn't useful for large magazine, rapid fire turrets, and interior turrets, which are most useful from an actual shooting point of view can't link to conveyor systems. 

A perfect demo for this turret would be a converted version of the [Large Ship Cannon turret](http://steamcommunity.com/sharedfiles/filedetails/?id=402952773) example mod (which needs converting to DX11 anyway).

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyLargeGenericTurret.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyLargeGenericTurret.cs
@@ -1,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Sandbox.ModAPI.Ingame
+{
+    public interface IMyLargeGenericTurret : IMyLargeConveyorTurretBase
+    {
+    }
+}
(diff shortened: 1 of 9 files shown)


CMDBob:
Closing this, as I've realised that it was obsolete before I even wrote it, as missile turrets already do what I want.

PR #488, opened 2016-03-06
Soundblock Categories

Eikester:
This adds a second Listbox in the Soundblock Terminal which lists
Categories by either the SubtypeId or Displayname to help with huge
Soundlists. Only Sounds of the selected Category will be displayed in the "Sounds List" Listbox

[Screenshot ](http://steamcommunity.com/sharedfiles/filedetails/?id=639401144)

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MySoundBlock.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MySoundBlock.cs
@@ -38,6 +38,7 @@ namespace Sandbox.Game.Entities.Blocks
         private readonly Sync<float> m_volume;
         private readonly Sync<MyCueId> m_cueId;
         private readonly Sync<float> m_loopPeriod;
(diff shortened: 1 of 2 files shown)


mexmer:
missing "All" just in case.... and maybe use "Built-In" instead of "Default" for sounds, that are in game from start.

Eikester:
"All" is a good point, "Default" is part of the Sound Category SubtypeId, because nobody used the Property "DisplayName" in their Sound Collections i've used the SubtypeId for the Category name as fallback ( <SubtypeId>DefaultCategory</SubtypeId> ). another way would be to put all Sounds into an "All" Category and introduce a new Parameter for the SoundCategoryDefinition i.e. "Group" or similar. Would allow multiple Mods adding Sounds to the same Group

Eikester:
imo the best solution would be a Treeview like control but for now it is better than before

mexmer:
i see, didn't know about default, it makes sense then.

Issue #490, opened 2016-03-10
Expose a projector's loaded blueprint to ModAPI


zrisher:
Currently, there seems to be no way to access a player's blueprints via ModAPI. 

This functionality would be very useful for a mod that wants to create ships from player's BPs or calculate the component cost of a player BP.

There may be some privacy issues to address here - perhaps a player doesn't want a server-side mod to have access to all their BPs. 

There are also data transfer issues - a full list of local BPs is a lot of data to serialize and send over the network.

So, as a first step, what if we allowed ModAPI to access a BP that has been loaded into a projector? That way the player still has some control over what's available to a mod, and we don't need to add any messaging because this data is already available to the server.

zrisher:
Implemented in #491, feedback is welcome!

Issue #489, opened 2016-03-10
Unable to compile/build from source


zrisher:
Following the build guide provided in [the Readme's quickstart section](https://github.com/KeenSoftwareHouse/SpaceEngineers#quickstart), I am presented with the following build errors with the latest version of master:

```
Error	4995	The type or namespace name 'StartupResultEnum' could not be found (are you missing a using directive or an assembly reference?)	C:\Sites\zrisher\spaceengineers\Sources\VRage.Network\MyRakNetStartupException.cs	11	25	VRage.Network
Error	4997	The type or namespace name 'StartupResultEnum' could not be found (are you missing a using directive or an assembly reference?)	C:\Sites\zrisher\spaceengineers\Sources\VRage.Network\MyRakNetStartupException.cs	13	57	VRage.Network
Error	4994	The type or namespace name 'ConnectionAttemptResultEnum' could not be found (are you missing a using directive or an assembly reference?)	C:\Sites\zrisher\spaceengineers\Sources\VRage.Network\MyRakNetConnectionException.cs	10	25	VRage.Network
Error	4996	The type or namespace name 'ConnectionAttemptResultEnum' could not be found (are you missing a using directive or an assembly reference?)	C:\Sites\zrisher\spaceengineers\Sources\VRage.Network\MyRakNetConnectionException.cs	12	60	VRage.Network
Error	5177	Metadata file 'C:\Sites\zrisher\spaceengineers\Sources\VRage.Network\bin\x64\Release\VRage.Network.dll' could not be found	C:\Sites\zrisher\spaceengineers\Sources\Sandbox.Game\CSC	Sandbox.Game
Error	5178	Metadata file 'C:\Sites\zrisher\spaceengineers\Sources\VRage.Network\bin\x64\Release\VRage.Network.dll' could not be found	C:\Sites\zrisher\spaceengineers\Sources\SpaceEngineers.Game\CSC	SpaceEngineers.Game
Error	5000	Metadata file 'C:\Sites\zrisher\spaceengineers\Sources\SpaceEngineers.ObjectBuilders\bin\x64\Release\SpaceEngineers.ObjectBuilders.dll' could not be found	C:\Sites\zrisher\spaceengineers\Sources\Sandbox.Common\CSC	Sandbox.Common
[...]

PR #491, opened 2016-03-10
expose projecter loaded BP to modapi

zrisher:
Implements #490:

* Add a method to get the currently loaded blueprint from a projector to ModAPI
* Add a projector interface to ModAPI inheriting from the ModAPI.Ingame version (so we have a place to expose the above)

I haven't submitted a PR to this repo or even worked with C# for a few months, and I can't build the project due to #489. So this probably has a bug or two. Please let me know and I'm happy to fix!


Phoenix84:
I like this.
My only other suggestion would be to also add a LoadBlueprint method which accepts a MyObjectBuilder_CubeGrid. This would allow mods to load custom grids programmatically that aren't saved as a blueprint.

PR #493, opened 2016-03-10
Added Properties to IMyPlayer and IMyIdentity

Rynchodon:
Makes it easier for mods to move between IMyPlayer, IMyIdentity, and IMyCharacter.
Adds tests for whether a player is a human or a bot to IMyPlayer.
Adds property for current movement state to IMyCharacter.

The goal here is to give mods more information about characters so they can react accordingly.
For my purposes, it would be very useful to know whether a character is a human or a bot, and whether or not a character is alive.


zrisher:
:+1:

PR #494, opened 2016-03-11
Add to whitelist to support yield return

andyleap:
Whitelist needs the following added in order to allow usage of IEnumerable<> functions and yield return

System.Environment.CurrentManagedThreadId
System.NotSupportedException.*


andyleap:
This has gotten added separately, closing this PR.

PR #495, opened 2016-03-12
Japanese Character Input system Create

everysh95:
Hello!!
I am Japanese people.This game is interesting.But it has unfortunate point.
This is what this game do not have a Japanese IME system.
In order to improve it, I Create Japanese character input system that is incomplete.
Thank you.


lord-devious:
You have committed a lot of commented-out error checking. You should revert those :smile: 

Also... did you really make those shader changes? I'm suspecting you have commited a bunch of things you shouldn't have...

everysh95:
I'm sorry. I had forgotten to add the MyTSF of reference is a project made to the default of the project.
In addition, skeptical of, but can be seen because the amount is often about the source of MyTSF, basically was referring to the web site and the bottom of the source written in Japanese. However, the sample marked was in MSDN to GetThreadMg () It has been taken as it is.

everysh95:
I remove those commented-out error checking that I think,and add the MyTSF of reference.

mexmer:
ime support itself would not be that big problem, since game supports unicode input anyway. but you still don't have font with hiragana, so it doesn't matter if your input can handle it, when you cannot display it.

keen is planing to have asian version of game (not sure when, but it was mentioned in one of blogposts)

everysh95:
Tentatively be introduced mod of this URL, it was a test because it can be displayed for Japanese characters.
[Japanese Font Pack](http://steamcommunity.com/sharedfiles/filedetails/?id=442628194)

PR #496, opened 2016-03-13
[FIX] Inventory transfer failure bug

Jimmacle:
This PR fixes the issue where an inventory item could not be moved from one index to another in the same inventory if the amount was greater than the free space in the inventory. I solved this by checking if the source and destination inventories are identical, and if so remove the item before adding it at the new index to prevent AddItems from failing.

I do have a concern regarding my changes, and that is that the ternary operator in the original code used in AddItems would now never evaluate to true. If that is going to cause problems please let me know.


Jimmacle:
Update: This bug seems to have disappeared so ignore this PR for now.

PR #497, opened 2016-03-13
[FEATURE/QoL] Add ingame IMyThrust properties MaxThrust and CurrentThrust

Jimmacle:
This PR adds two new properties to Sandbox.ModAPI.Ingame.IMyThrust that allows ingame scripts to get the max thrust of a thruster (includes atmosphere multiplier) as well as the current actual thrust output.


thedanieldude1:
We definitely need this in game ever since they changed the override terminal property!

Jimmacle:
This commit is merged as of SE version 01.142

PR #498, opened 2016-03-13
Fix MyGuiScreenEditor typo breaking ingame script compilation

Jimmacle:
This PR fixes a strange typo in MyGuiScreenEditor that is only present in the GitHub source and not the production build. It results in ingame scripts failing to compile at "line -2" because what should be a using directive is misspelled and lacking the "using" keyword.


Megaddd:
Support.

Jimmacle:
Already fixed internally.

PR #499, opened 2016-03-14
FEATURE: DS config improvement

mexmer:
added version to config form title
worldname is filled by sessionname from saved game
reduced config form initial height to 760 (to fit onto 1368 x 768px display)
world list also contains folder after worldname to distinguish between world copies

redo of #393 

image


locint:
This seems very useful.

mexmer:

PR #500, opened 2016-03-15
I my ship controller exposed inputs

plaYer2k:
This change exposed the 6DoF inputs from a player within any
IMyShipController.
The new exposed values are:
Vector3 IMyShipController.Translation
Vector3 IMyShipController.Rotation

The "gameplay explanation" behind that is that both the cockpits
joysticks as well as the remote controls inputs can be easily read out
and exposed by a proper interface.

With this change people can have multiple seats without any
thruster/gyro/suspension wheel requirement to read the seated pilots
inputs and offer diverse simplified inputs for custom
thruster/rotor/piston drivers as well as 2D+ menu controls like arrow
keys in most applications allow.

Example ingame script for readouts:
http://pastebin.com/83QLRdD6


Pingger:
I'd like to see this implemented! This would simplify Gravity Drives!

plaYer2k:
Well, i have written that twice for merges now. A few more times including local re-writes for myself.
Given how inactive keen became, i consider the public git contribution community dead. Only a few selected ones get considered and everyone else gets ignored. I also greatly dislike the silence.

To me that means that i will not contribute anymore so i also wont update this PR.
However the core of this PR is as simple as it gets. Have variables for both Vector3 and overwrite them with either zero for certain conditions or otherwise with the input values for WASD C Space Mouse and expose them.

http://forum.keenswh.com/posts/1286975021/

Jimmacle:
They'd probably be more likely to merge this if you rebased your branch to remove the conflicts.

plaYer2k:
This is all the "rebase" and time i will put into PRs with the currently shown dev-activity ...
... 4 months of inactivity dont quite convince people to spend their time over and over and over again for nothing.

Issue #469, opened 2016-02-08
Air Vents don't take into account external atmospheric conditions


lostkangaroo:
As far as I have been able to track down this reasoning, it seems that air vents are still tied to the assumption they would only be used to pressurize sealed internal rooms.  This assumption should be changed due to planets introducing another type of environment, the planets atmosphere.  Air vents should now report the value for the environment the grid they are attached to is surrounded by when an oxygenBlock.Room is missing; 0 for space naturally or the planet/moon/celestial body's atmospheric default.

mccorkle:
Hey Lost -- This issue tracker is for discuss source code contributions, not making bug reports.  There is a dedicated thread (http://forum.keenswh.com/forums/bug-reports.326950/) to post bug reports for Keen to see.

lostkangaroo:
This is not a bug report but a feature request as the current functionality does not exist in current assumptions. It can be considered a bug report if the general idea is for this behavior to exist, but that is a Keen decision.

nobodyhasthis:
If it is a feature request. Try putting it in http://forum.keenswh.com/forums/suggestions.313735/

lostkangaroo:
It has been but I have a feeling its been lost to the black hole of page 99+ by this point. My concern is not so much about what keen forum to post this in.  If there is another oxygen system PR or issue that is gaining more traction I would be happy to attach this issue to it rather then pursue this specific feature in this issue.

PR #501, opened 2016-03-15
IMyAirVent exposed GetAirDensity

plaYer2k:
Using this extension, the AirDensity of the closest planet can be
determined with any IMyAirVent.

The "gameplay explanation" behind this is that every air vent can
measure oxygen already (though only for rooms) so it has to be able to
measure the general density of the surrounding atmosphere as well.

Exposed value:
float IMyAirVent.GetAirDensity()

Ingame script example:
http://pastebin.com/AQMyBRyN


mexmer:
i'm not sure about this.
if anything, this should imo work only when you inside planet atmosphere, not as planet with atmosphere detector, also it should work only on "outside" wents.

planet with atmosphere detector, might be part of comp logic, therefore for example on controler (cockpit) hud, but not some magic air vent.

plaYer2k:
I dont see how the cockpit should be able to get the air density at all. The cockpit is the device the character uses to interact with the ship as well as what forwards informations.
However these informations about atmosphere utlimately has to come from the air vent as that is the logical sensor for air/atmospheres.

So in fact, if you would get the atospheric density from the cockpit without having any actual sensor like an air vent, that would be "magic".

Sadly that is the case right now, though it shouldnt stay like that for the ingame api.


Furthermore, the returned value is a float. So outside of any atmosphere the value is always 0 anyway. Thus if "it works" is defined as "returning a value >0" then it does work only inside atmosphere.

As for the "only outside vents", that is a rather high requirement in my opinion.
Many game logics in SE do not behave that realistic and thus it shouldnt be too restrictive there.
One very similar example are the atmospheric thrusters in SE. Their efficiency changes even when they are encapsulated within a grid and thus "at the inside of a sealed box".
Additionally you could have multiple encapsulated ships where the "outside block" of one grid still is inside another grid. So the "be outside and exposed" is a rather tough requirement after all.

mexmer:
having density exposed in cockpit is same magic as speed, gps, position, power levels and such ... as you say. cockpit is interaction point, it uses any and every device on grid and gives you information on display, hud

i said that already before, that any information that is exposed on hud should be available trough controller for reading, because that is your interaction point.

namely 
grid mass
grid energy level
grid orientation
cockpit oxygen status (if controller is cockpit/seat and not remote)
grid coordinates (where coordinates are controler position)
grid speed
grid vector
grid altitude (on planet)

we can continue

or have alternative interface for that (although controller seems to be just fine to me)

lostkangaroo:
#469 tagging this as they are definitely related

lord-devious:
This one I like. Only comment I have, is perhaps the "difference" between room- and planet oxygen level should be hidden behind a single method. As far as the player is concerned there should _be_ no difference.

lord-devious:
Having thought about this one, seeing the problems with joining the two methods as I suggested above, I realized that mexmer is right. Atmospheric density should be available in the general flight data of cockpits, not the air vent. As there is no requirement of the air vent to actually be in contact with the atmosphere it does not make any sense that it has this kind of data access. The cockpit blocks however are the natural convergence point of any data that doesn't have an explicit data retrieval block.

lostkangaroo:
Cockpits already have general planetary atmospheric density access which is the magic data point. I argue that a vent shouldn't care about general planet or even internal room density and should only access that which it is connected or the face actually touches.  The cockpit can have its reading but each vent should have its own localized one as well.

PR #503, opened 2016-03-16
Added Ingame.IMyCharacter with exposed PlayerId and HeadMatrix

plaYer2k:
This change adds a new file Ingame.IMyCharacter which exposes the
properties long PlayerId to identify the player as well as MatrixD
HeadMatrix to properly track where the player is looking at even when
standing on a surface under gravity where the WorldMatrix Down is always
aligned with the gravity vector and thus any vertical position can not
be gathered.


Jimmacle:
PBs should _not_ be able to know either of these things. They shouldn't be able to get a character reference in the first place.

mexmer:
As for identity i agree, but as for facing direction, that's imo information, that should be provided by sensor anyway ... i don't see anything wrong with that.

plaYer2k:
Well we just had that discussion in discord already.
Though for everyone not participating there.

There sure is a fundamental flaw with the PBs being able to keep references. Though this pull request of course bases around the current system and no potential future system that could be implemented any time soon and is based around structs/valuetypes over objects/referencetypes.

Additionally, the PlayerId is just as wrong as the IMyCubeBlock.OwnerId is. Both do provide the same information.
Knowing the owner of a block is equally vital like knowing the "owner" of a character.

After all the build in sensor functionality provides that already in a non-exposed way by being able to trigger only for owner, neutrals, enemies etc.
Thus same functionality should be exposed to the characters as well.


As for the facing direction, that also is a pretty viable option imo.
We currently got considerably accurate tracking methods to detect faces on image streams in realtime as well as process the orientation in 3D.

Now the real question should be "how realistic" or "how gamified" the game elements should be. How much of a blackbox is okay. Though these questions get answered by practical constraints (we cant replicate that level of image processing and alike with realtime performance ingame) as well as plain usability in the scope of a game.

So an overly realistic approach would mean that the current sensors are silly already because they can detect through walls and are generally a blackbox in how they work and thus any 3D positioning system should be based on two cameras from different angles onto unobscured entities.
But we all know that is not feasible.

It is a game after all. It has to run in realtime with acceptable performance and thus certain realistic considerations have to be neglected.

lord-devious:
I'm afraid I'm going to have to agree with the others here. One wrong does not justify another. Please keep the programmable block as ingamey as possible.

PR #506, opened 2016-03-28
FEATURE: per instance log name date setting

mexmer:
this settings allow "append date to log name" option to be set per instance basis.
since on hosted services usually people don't have option to enable/disable it.

you will be able to find/add following line in spacengineers-dedicated.cfg

``` xml
  <AppendDateToLogName>0</AppendDateToLogName>
```

possible values
0 - no date
1 - append date
2 - follow global setting


mexmer:
rebased and fixed to fit in new layout

mexmer:

PR #507, opened 2016-04-01
[Feature] Incomplete / Damaged blocks are lighter than complete blocks

kdavison:
1. grids recalculate grid mass after an integrity change ( ie a block is cut / damage )
2. grids use the actual block mass instead of the template mass
3. blocks now calculate mass using the ratio of Integrity / MaxIntegrity

Testing
1. limited x64 testing. Physically blocks appeared to behave as expected ( no weird negative mass issues )
2. no x86 testing, memory error prevented me from testing x86.

Issues
1. Ship Controller's report mass in correctly. still tracking down the source of this problem. Anyone got ideas?

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Cube/MyCubeBlockCollector.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Cube/MyCubeBlockCollector.cs
@@ -396,7 +396,7 @@ namespace Sandbox.Game.Entities.Cube
 
         void AddMass(MySlimBlock block, IDictionary<Vector3I, HkMassElement> massResults)
         {
-            float mass = block.BlockDefinition.Mass;
+            float mass = block.GetMass();
             if (MyFakes.ENABLE_COMPOUND_BLOCKS && block.FatBlock is MyCompoundCubeBlock)
             {
                 mass = 0f;
--- a/Sources/Sandbox.Game/Game/Entities/Cube/MyCubeGrid.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Cube/MyCubeGrid.cs
@@ -4145,6 +4145,7 @@ namespace Sandbox.Game.Entities
             if (block != null)
             {
                 block.SetIntegrity(buildIntegrity, integrity, integrityChangeType, grinderOwner);
+                Physics.UpdateMass();
             }
         }
 
--- a/Sources/Sandbox.Game/Game/Entities/Cube/MySlimBlock.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Cube/MySlimBlock.cs
@@ -1915,7 +1915,7 @@ namespace Sandbox.Game.Entities.Cube
         public float GetMass()
         {
             if (FatBlock != null)
-                return FatBlock.GetMass();
+                return FatBlock.GetMass() * Integrity / MaxIntegrity;
             Matrix m;
             if (MyDestructionData.Static != null)
                 return MyDestructionData.Static.GetBlockMass(CalculateCurrentModel(out m), BlockDefinition);


Evrey:
This is a realy nice idea!

I don't like the division over there much, but it's fine. It isn't hot path code. So... merge please?

mexmer:
for gamification sake this simple calculation works, and would be better than nothing. but to do it properly, you should deduct missing components, and not use integrity ratio.

that said, i like the idea.

kdavison:
Glad people are excited about this!

@Evrey  - Actually I agree with you. As is the game will recalculate the entire grid mass each time the game receives an IntegrityChanged message. A server with a bunch of people modifying a grid could get cumbersome, especially for large grids.

@mexmer - I can modify to use component counts directly.  I'm pretty sure the integrity values already count components, but couldn't hurt to check.

mze9412:
So if I start welding a block it deposits all components into it on the first click. Welding progress is still below 10%. Would that not open up very strange mass workarounds? O_o

kdavison:
@mze9412 - nope! The game currently applies the entire mass of a completed block, even though its only 10% complete.  This is because the GetMass() function uses the blockDefinition mass, and performs no calculations to figure out the actual mass.
The physics works fine with my limited testing.  If the mass actually went negative or something the grids would have move the opposite direction to the applied thrust!

Caveat 1:
Getting a consistent grid mass will be difficult with this patch because Physics.RefreshMass function uses the current Integrity of a block, not the integrity that was sent with the IntegrityChanged function.  "Fixing" this would require calculating the nearest "integrity level mass": a block at 10% would "round up" to the "33%" mass value.
I'm not privy to Keen's development pipeline, so I chose simplicity over anything else.

Caveat 2:
The Mass display from a IMyShipController interface does get messed up with this patch.  I haven't had time to figure out if that because of the patch or a bug got exposed.  Probably just missed a callback to update the "display mass"

mze9412:
So you factor the integrity into the mass calculation but not the amound of "deposited" components, right?

kdavison:
That is my current understanding.
As I told @mexmer above, I will do an investigation and update the patch as needed.

Just browsing on my phone it looks like MaxIntegrity comes from MyCubeBlockDefinition.  Which means it is the maximum possible value for that cube block type.  My action item is to check that Integrity is calculated from the current mounted ( built ) components, and not just any components.

mexmer:
@kdavison for missing components you need to use same approach as welder, it's possible, but question is, if you want to go trough that hassle.

like i said, your implementation is fair enough ... it's game.

just some people want to be "precise"

kdavison:
Did some more testing / code investigation:
Using this patch unmounted components do not affect block mass.  So a block could be full of un-built components, but would weigh very little.

This is fixable by counting components as suggested, but I don't think that would improve gameplay.
I think its much better, to use a WYSIWYG approach so a visually unbuilt / damaged block weighs less, even if all the "stuff" is there.
Additionally, I don't want to go around looking for un-built components adding some mystery mass to my ship.

mze9412:
Hm, I would actually then rather go for the current behaviour: Always use the full weight. Otherwise I would expect that the behaviour you suggest will be abused for sure (i.e keep a ship light until necessary without putting the components into a cargo container in order to save hydrogen in a planetary lander or something like that) :\ Might especially become funny with some automated setups which start to grind down certain blocks automatically (i.e. down to 10%), then deposit components back in with a welder but stop welding immediately (somewhere below 20%). Don't open up that way please :\

kdavison:
<quip> if you sim speeds can handle that :p </quip>

Alright I can figure how to count all the available components in the block and calculate the mass off that.
Will probably look like:

...
FatBlock.GetMass() * AvailableIntegrity() / MaxIntegrity()
...

where AvailableIntegrity counts and calculates the integrity from all of the populated components in the componentStack.

kdavison:
actually, that didn't seem to work D:

Evrey:
Maybe I'm just too anxious about it, but the new code looks like waaay too much calculation for a way too simple task to me. `mass = maxMass * integrity / maxIntegrity` or `mass = maxMass * integrity * invMaxIntegrity` would have done the job and is dead simple to read/understand. Now it's loops and what not. Besides that...

> This is fixable by counting components as suggested, but **I don't think that would improve gameplay**. I think its much **better, to use a WYSIWYG approach** so a visually unbuilt / damaged block weighs less, even if all the "stuff" is there.
>
> Additionally, I don't want to go around looking for un-built components adding some mystery mass to my ship.

↑
This is a game designer speeking.
(**Highlight** by me.)

mze9412:
My recommendation would be to not do anything (leave the current system which always uses MaxIntegrity untouched). I really worry about the "mass cheating" that will for sure happen, especially in DS/PvP situations.

mexmer:
@mze9412 if calculation is based on simple integrity ratio, then you are right, if it's based on component storage, then it should be fine.

i had this discussion before, and indeed had same objection, what is proper source of mass computation. since you literaly can have 1% complete block, that has full storage of components. even on heavy armor or worse blast doors, difference between integrity (5%) for example, and full component inventory (which is 25 steel plates for heavy and 125 for blast door, if i rememeber correctly) is huge.

there are different PoV when it comes to block and it's proper mass, but since you need to take in account game engine ... well.

TBH i never liked idea, that block has "internal storage" for components, but that's just me. since there is no other place to "store components" (they can't float around, when you buiild something), and at 1x your storage is rather limited (i mean for suit), looks like good solution, but in terms of other aspects (like mass/integrity consistency), it's rather bad solution.

mze9412:
@mexmer 
I basically adressed the comment from @Evrey :)

I agree that the internal storage for components should not exist the way it does. That would be the perfect solution.

Evrey:
@mze9412 
I do respect your concerns about mass cheating and I indeed think that this might be an annoying issue. It might happen, if you happen to play PvP against less honourable players, but at the same time it might happen that just being detected by an enemy's sensor enables their programmable blocks to take control over your ship. At least that used to be a possible exploit, besides others. Again, depends on the honour of your PvP enemies whether they exploit such bugs.

However, my way of thinking is this:

The task of this code is just *"calculate the mass of a block depending on how 'complete' it is"*, where you might derive its completeness from the integrity alone or whatever stuff is in there. My issue here is this: Why can you even "put stuff" into an unfinished block before really welding it up? Why don't blocks only accept components that are welded immediately? This is a design flaw, IMHO, that has nothing to do with the mass calculation code. It is a separate bug to fix (or fix it right in this pull request, I don't care). How does this even make sense? This is exactly the same thing for PBs, where it isn't the PB's fault that the whole "Sandbox" ModAPI is exploitable just everywhere.

Besides that... you'll need a ton of welders and some precise timing to make this exploit really work. And I don't really expect lots of guys actually flying around with welder-bloated ships during PvP matches or whatever just to spend less hydrogen. But this might as well just be myself being naϊve.

**tl;dr:**

- I prefer the simpler algorithm, as it makes more sense.
- I do see the problem why the more complicated algorithm would be needed.
- I don't think that it is the mass calculation's duty to fix a design flaw leading to the mentioned exploit.
- If the most widely accepted opinion happens to be "complicated algorithm or vanilla", I vote for vanilla, as in "don't touch a running system" and the KISS principle.

mze9412:
@Evrey 
The PB exploits are in the process of getting fixed. No need to create new different possibilities to exploit the game. Just because one part of the game can be exploited (I think most exploits in the PB are fixed already) does not make it a good idea to open up more possibilities.

People do fly around in welder-bloated ships already because of self-repairing ships. With a projector block even completely self-repairing in some parts. Other builds have automatic build/dismantle of certain party already.

I see two "good" options:
A) Correct the fundamental design flaw (blocks should only contain components thate are 'welder' already)
B) Leave it as it is because the current functionality is less of a problem then the proposed change to ignore the 'unwelded components'
That is at least my opinion ;)

PS: I do not really think that KSH did the 'unwelded components' thing without intention. I fear that it was exactly done as intended and maybe even the mass implementation was done exactly for that reason the way it is right now :D

Evrey:
So, besides my lack of experience with PvP, we seem to be on the same page then. °^°

I think their intention was to be able to dump materials nearby your building site without having to scan the area around you for stuff, as the materials can only be welded in order and as your pockets arent Final Fantasy sized.

kdavison:
Solutions:
Method 1: Change blocks to use (user hidden) inventories instead of a stockpile.
Inventory calculations are already included in mass; the for loop gets removed and patch functionality is unchanged.
This would be a quick change.

Method 2: Get rid of the whole stockpile concept.  The block either has components or doesn't.
Blocks would still have a list of required components, but you would have to have the component on-hand when it is required during the welding process.  You would be unable to supply component #3 before component #1 etc...
This would require significantly more work, maybe another patch?

mze9412:
@kdavison 
I thought about the stockpile thing. Is it maybe also in so that two welder can work on the same block but not both would need to supply components?

kdavison:
Based on the welder code, I don't think so.
I think it just makes it easier to build things as a Space Engineer as opposed to with a drone or ship.

I take it you are in favor of Method 2?

Evrey:
At least I favour Method 2. I'm looking forward to hearing how commenting this function call out will work.

kdavison:
I removed the FillStockpile function from both Grinder and Welder, so neither of those tools used the block stockpile.

No change in functionality.

It is difficult to test this fully because I am encountering a rendering bug on the 64-bit version of the source where the Glare Light Radius is an invalid number ( +/- NaN ).  If you have any information on how to fix this, it would be greatly appreciated.

Evrey:
As far as I remember, only divisions by zero can generate NaNs by accident. (NaN-infection by doing maths with NaN does not count.) Thus, you can spot the source of your glitch by `Ctrl+F` your way through code which is directly or indirectly affected by your commits. Are there any permanent `float` or `double` values `FillStockpile` sets? Or `float`/`double` values that are derived from values `FillStockpile` sets, such as item counts?

kdavison:
I don't think its related to my change.  Its probably a conflict between the more recent x64 binaries and the slightly older source code.

The issue I'm having is with the renderer; specifically when rendering thruster output or ship grinder sparks.  Somehow the Billboard code is receiving an invalid value and crashing as a result.  I don't think I can just hack my way around it; perhaps the forums will have some guidance.

Tyrsis:
I don't get this PR.  Why are you using integrity?  Just pull the mass from the components in the stack.  The components are there, you're hiding their weight by using a ratio instead of a sum of the parts there.

Rynchodon:
Won't this require constant recalculations of the inertia tensor?

PR #508, opened 2016-04-03
[Feature] GPS Tab: Show/Hide All Buttons

Basuro:
A little usability feature (well, basically three):
- Two buttons to show or hide all coordinates in the (filtered) list on the HUD
- Creating a new coordinate with the _New_ and _New from current position_ buttons selects the newly created coordinate so one doesn't have to search for it in the list


mexmer:
set tab->space autoconversion in visual studio for code indentation, that way you will not need to fix it later.

Basuro:
yup, already did that, right after my first commit ;)


Am 03.04.2016 um 12:37 schrieb Josef Micka:

PR #509, opened 2016-04-03
tell Visual Studio to use 4 spaces

Rynchodon:


mexmer:
forget it, that's your personal visual studio setting, such thing should never be posted onto github.

same with .vs directory, suo files and other.

Rynchodon:
>forget it, that's your personal visual studio setting, such thing should never be posted onto repository.

>same with .vs directory, suo files and other.

That is not what this is, editorconfig is more like gitattributes.
If they add an editorconfig file you will never have to tell anyone to convert tabs to spaces again.

Jimmacle:
I agree with mexmer, the C# compiler ignores white space so the indentation settings don't matter.

Basuro:
While the C# compiler may ignore white space, humans don't.
This .editorconfig (with the needed VS plugin) will help keep the code tidy so it looks uniform and doesn't make you shudder when several people have edited it.
Plus it prevents users from having to set their diff view to ignore whitespace to see the things that matter.

Jimmacle:
@Basuro A tab is the same amount of white space as 4 spaces so it still doesn't make a difference.

Basuro:
Unless you set another default setting of eg. 2 or 8.

Phoenix84:
If it requires a plugin, this change isn't useful. Not everyone is going to download a separate plugin to use this.
Not only that, but VS by default inserts 4 spaces in places of tabs for C# code. If it's putting in tabs, someone else _already_ changed their VS settings from default. Or they are using a separate editor.
Personally, I _hate_ spaces and much prefer tabs, however since that's not the C# way, I have to deal with it.

Rynchodon:
It does not require a plugin.

This is really not worth arguing about.

PR #510, opened 2016-04-08
add CurrentStoredPower set

yajiedesign:


mexmer:
```class MyBatteryBlock : MyFunctionalBlock, Sandbox.ModAPI.IMyBatteryBlock, Sandbox.ModAPI.Ingame.IMyBatteryBlock
```
you don't need to add both, since class Sandbox.ModAPI.IMyBatteryBlock inherits from Sandbox.ModAPI.Ingame.IMyBatteryBlock, so you only need topmost of them.

yajiedesign:
@mexmer ok.fixed

PR #511, opened 2016-04-09
Make MyGridTerminalSystem_ModAPI more intuitive

Aeronaut:
Added a few Methods to make the GridTerminalSystem more intuitive.

### DIFF
--- a/Sources/Sandbox.Common/ModAPI/Ingame/IMyGridTerminalSystem.cs
+++ b/Sources/Sandbox.Common/ModAPI/Ingame/IMyGridTerminalSystem.cs
@@ -7,10 +7,17 @@ namespace Sandbox.ModAPI.Ingame
     public interface IMyGridTerminalSystem
     {
         void GetBlocks(List<IMyTerminalBlock> blocks);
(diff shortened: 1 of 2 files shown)


mze9412:
I would really prefer to have such an approach, but there are very good reasons why KSH does not offer these methods. Every time such a method is called, it creates a new list. The KSH approach allows you to reuse your list to avoid creating unnecessary instances.

If you really want to have this appreach (I would like it) please initialize the lists with the correct capacity (you already know how big the list needs to be, don't make the runtime having to resize the list all the time). So instead of
+            var blockGroups = new List<IMyBlockGroup>();

Do something like sith:
+            var blockGroups = new List<IMyBlockGroup>(BlockGroups.Count);

Overall I agree with more intuitive methods ^^

Aeronaut:
Thanks for your input. I edited the GetBlockGroups a little bit, since it just needs to be converted I simply return a converted List...

Overall I think there should be no Reason for reuse Lists, if it will contain what is needed.

Aeronaut:
Forgot to make a new branch...

PR #512, opened 2016-04-09
Make MyGridTerminalSystem_ModAPI more intuitive

Aeronaut:
Added a few Methods to make the GridTerminalSystem more intuitive.


mze9412:
The reason is the following:
You can run a script once per tick. Each tick you call GetBlocks() and get a new list. Instead you could create the list in your script, store it in a member and use it with GetBlocks(myList) each time the block runs. This is not a big thing but can save some memory or garbage collector overhead. Especially if there are many scripts running (i.e. on a DS)

Aeronaut:
Oky, understood.

lord-devious:
I'm sorry, but this is highly unlikely to be merged due to the list instantiations.

Spartan322:
I also doubt the power of redundancies getting this through.

PR #518, opened 2016-04-27
IMyRemoteControl.GetDistanceToSeaLevel

suspendmode:
Method copied from MyGuiScreenHudSpace.FindDistanceToNearestPlanetSeaLevel.


Squalalah:
might be usefull and more easier to find.

Jimmacle:
I think your GetNearestCenter method is too "magic" for use ingame if you get what I mean, but other than that this looks pretty useful.

PR #519, opened 2016-04-29
Allow modAPI (script) access to OnGridChanged events.

fabricator77:
OnGridChanged is used in the cargoship code to detect when a Player has
taken over a ship so it doesn't despawn.

It's not possible to create the same functionality with other available
events, which can be triggered by things like minor damage to ships, or
unpowered ships drifting.

### DIFF
--- a/Sources/Sandbox.Game/ModAPI/Blocks/MyCubeGrid_ModAPI.cs
+++ b/Sources/Sandbox.Game/ModAPI/Blocks/MyCubeGrid_ModAPI.cs
@@ -247,6 +247,12 @@ namespace Sandbox.Game.Entities
             remove { OnBlockOwnershipChanged -= GetDelegate(value); }
         }
 
+        event Action<IMyCubeGrid> IMyCubeGrid.OnGridChanged
+        {
+            add { OnGridChanged += GetDelegate(value); }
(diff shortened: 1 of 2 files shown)


Jimmacle:
I already spoke to you on Discord, but for the record I'll put it here too. Can't you use the event directly by casting your object to MyCubeGrid? The modding API still has a lot of work left but that method should work until the API gets some more love.

fabricator77:
I have to rely on several other PRs (that I'm working on) for my mod to work anyway.
IMyGrid exists for a reason, I'm not going to rely on MyGrid being exposed as that can change (eg the current ISensor secuity hole).

Why are you abusing me for actually improving the API.
For the record I don't believe in programming theory, long debates on best OO designs or endless refactoring.

Jimmacle:
Calm down, dude. I'm just trying to help you get whatever you're working on going without having to rely on PRs as they aren't getting merged right now. Sorry for trying to help.

fabricator77:
Huh, stupid GitHub client added that without my permission.

Jimmacle:
When you make a PR it's possible to update it by committing to the same branch. Good practice is to make a separate branch for each PR so you can have multiple open at once without interference :)

PR #526, opened 2016-05-10
Resource leak patch

AlonzoTG:
Having learned a great deal about everything, especially the codebase, where the booby-traps are hidden, GIT, and the pull request process, this is an experimental pull request that, while maybe too small, addresses all concerns previously mentioned. 

Addresses a potential resource leak to make sure containing class passes
destroy command to external resource.

### DIFF
--- a/Sources/Sandbox.Game/Engine/Voxels/MyVoxelGeometry.cs
+++ b/Sources/Sandbox.Game/Engine/Voxels/MyVoxelGeometry.cs
@@ -1,13 +1,8 @@
-﻿
-using Sandbox.Common;
-using Sandbox.Engine.Physics;
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Threading;
 using VRage;
 using VRage.Collections;
-using VRage;
 using VRage.Utils;
 using VRage.Voxels;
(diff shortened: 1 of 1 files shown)


lord-devious:
Adding IDisposable helps very little if Dispose is actually not being called :smile:

Jimmacle:
He's right, you'll need to go further into the code and make sure all those instances actually get disposed if you want your changes to have an effect.

Jimmacle:
He's right, you'll need to go further into the code and make sure all those instances actually get disposed if you want your changes to have an effect. Remember, your tools are just giving suggestions - you can't let it program for you.

Rynchodon:
Dispose is called by the destructor of FastResourceLock; there is no leak.

As is the case with FastResourceLock, the destructor is responsible for releasing unmanaged resources. IDisposable.Dispose is for releasing resources when you cannot wait for garbage collection.

Jimmacle:
Regardless, if Dispose is not called on the MyVoxelGeometry instances (which the game is not programmed to do because you just implemented it) your changes have no effect.

lord-devious:
@Rynchodon and @Jimmacle is right, there was never any need for this in the first place.

PR #527, opened 2016-05-13
Improved log output for mod loading

Dunkhan:
Commit message: 
Added logging for loading of all components, cubeblocks an voxel materials in mods
Supressed 'None extensions' logging when emtpy extensions array was passed on purpose
Added prefab path to prefab logging
Fixed issue where contentFile was set to null before being logged in 'Resource not found' log message

Explanation:
Troubleshooting XML mods is very difficult, usually the current version just says which file failed, and in some mods where files can reach thousands of lines with just a handful of added blocks, this can result in a lot of trial and error before finding the errors in the xml. This changeset logs the name of each component, voxel material and cubeblock as it is loaded so that it can be seen which one has failed.
The changes also include a cleanup of the existing debug output including the spam of 'Loading prefab' messages and the 'None file extensions' spam which should be familiar to most modders.


Jimmacle:
If only all PRs were as clear as this. Nice work :)

Dunkhan:
I checked the conflict and it is nothing. Basically an output string has slightly different text in my branch. When merging you can take either version it should not make a difference.

Let me know if you want me to resolve this myself.

PR #529, opened 2016-05-15
Fix for lag when recoloring blocks

Aleks976:
When recoloring blocks on a big surface, sim-speed would drop and fps would drop (even though it doesn't in debug view).

It was marking the blocks as "dirty" which would cause the Havok physics shape to be recalculated every time blocks were being painted, however the physics shape would never change when the color changed. 

You can see performance differences in this screenshot. http://imgur.com/xXWBOnb or in this youtube video. https://youtu.be/ytw4v7cd-kg


darkthing41:
I've known the physics re-calculations to cause some semi-stable Multiplayer builds to explode just by giving them a lick of paint, so stopping THAT would be nice.

Aleks976:
New to github, and accidently un-published the wrong branch, causing it to be closed. Now reopened. Also added a video to help show the bug.

PR #531, opened 2016-05-19
Fix: DX11 renderer is unable to load LOD models from mods directories.

theHarven:
The change forces LOD model files to be loaded from the same path as
main model file. The bug was reported here:
http://forums.keenswh.com/threads/1-105-directx-11-renderer-is-unable-to-load-level-of-detail-models.7371600/

### DIFF
--- a/Sources/VRage.Render11/GeometryStage/Geometry/MyMeshes.cs
+++ b/Sources/VRage.Render11/GeometryStage/Geometry/MyMeshes.cs
@@ -2112,6 +2112,7 @@ namespace VRageRender
             StoreLodMeshSections(id, 0, ref sections);
 
             int modelLods = 1;
+            var mainMeshDir = Path.IsPathRooted(assetName) ? Path.GetDirectoryName(assetName) : Path.GetDirectoryName(Path.Combine(MyFileSystem.ContentPath, assetName));
 
             if (lodDescriptors != null)
             for (int i = 0; i < lodDescriptors.Length; i++)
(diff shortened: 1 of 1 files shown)


harag-on-steam:
Currently the LOD path is supposed to be relative to a content root (<SE>/Content, <Mod>/). Why change this? This will create a needless mismatch between Dx11 and the legacy Dx9 renderer.

theHarven:
Hey,
It will not create any mismatch, because it's already done this way in Dx9 renderer, look at the code at VRage.Renderer\RendererObjects\MyRendererEntity.cs:
```
        // This is more hack than correct solution, we have to find lods in same dir as models, because of MODs
        private string GetLodPath(string modelPath, string lodPath)
        {
            return Path.Combine(Path.GetDirectoryName(modelPath), Path.GetFileName(lodPath));
        }

        MyRenderModel AddLODs(string model)
        {
            MyLodModel lodModel = AddLOD(0, model);

            if (lodModel == null)
                return null;

            foreach (var lodDesc in lodModel.Model.LODs)
            {
                string lodModelPath = lodDesc.Model;

                if (String.IsNullOrEmpty(Path.GetExtension(lodModelPath)))
                {
                    //Debug.Fail("Missing LOD file extension: " + lodModelPath);
                    lodModelPath += ".mwm";
                }

                lodModelPath = GetLodPath(model, lodModelPath);

                if (lodDesc.RenderQualityList == null || lodDesc.RenderQualityList.Contains((int)MyRenderConstants.RenderQualityProfile.RenderQuality))
                    AddLOD(lodDesc.Distance, lodModelPath);
            }

            return lodModel.Model;
        }
```

PR #532, opened 2016-05-21
Fix for meteors spawning underground, partial

progfz:
When in planet gravity and sun is below surface, meteors will spawn from
direction opposite to sun. It makes meteors much less likely to spawn
underground but is not a complete solution - deep underground bases and
bases on mountain hills still could be affected.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/MyMeteorShower.cs
+++ b/Sources/Sandbox.Game/Game/Entities/MyMeteorShower.cs
@@ -6,6 +6,7 @@ using Sandbox.Engine.Multiplayer;
 using Sandbox.Engine.Utils;
 using Sandbox.Game.Audio;
 using Sandbox.Game.Entities.Character;
+using Sandbox.Game.GameSystems;
 using Sandbox.Game.Multiplayer;
 using Sandbox.Game.SessionComponents;
 using Sandbox.Game.World;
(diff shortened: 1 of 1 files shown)


fabricator77:
Wonder what happens at the equator though, where the meteors would be travelling parallel to the surface.

Jimmacle:
@fabricator77 They would be traveling directly at the planet's surface at the equator.

progfz:
@fabricator77 why should they be parallel to surface at equator? With original logic meteors always come from sun and my addition is only reversing that direction if angle between reversed gravity and direction to sun is greater than 90 degrees. Parallel trajectories will happen when direction to sun is nearly parallel to surface, just as how it's in default implementation.

@Jimmacle it's nothing hard to implement by taking reversed normalized vector of gravity instead of direction to sun and leaving all other calculations as is. But meteors from the sun are current way of how things are working so I just based my calculations on it. In my opinion things should be done not in a rush but rather one step at a time - for example,  first to resolve most broken part of underground meteors and then to start thinking about more complete fix or even about feature overhaul, in this way if first pull request is accepted it will give biggest immediate impact with minimal efforts while gathering feedback and ideas for future iterations.

BTW, currently meteor showers are not bound to player but rather to block grids near player, if I understood code correctly, and showers already have central point that could be used for all kinds of calculations instead of player position.

PR #535, opened 2016-06-01
Fixed reduced damage from Missiles when multiple are shot at once

Aleks976:
Originally Missiles would cause eachother to blow up, expecially when multiple are shot at once, and even worse if shot from an angle. This would weaken the overall damage of the explosion when multiple rockets go off near eachother, either by blowing the rocket up, or reducing the damage of each individual rocket by damaging the rockets instead of damaging the blocks. Missile explosions will now not effect other missiles, as it serves no gameplay purpose, other than causing multiple rockets to deal minimal damage. Gatling turrets can still shoot down missiles. 

Made a video to demonstrate the change: https://youtu.be/gcHqaJlkkAA
(16 Missiles are fired at once, at large ship heavy armour blocks)

### DIFF
--- a/Sources/Sandbox.Game/Game/ParticleEffects/MyExplosion.cs
+++ b/Sources/Sandbox.Game/Game/ParticleEffects/MyExplosion.cs
@@ -549,6 +549,13 @@ namespace Sandbox.Game
                     }
                 }
 
+                //Special case for other Missiles
+                var missile = entity as MyMissile;
+                if (missile != null)
+                {
+                    continue;
+                }
+
                 var raycastDamageInfo = explosionDamageInfo.ExplosionDamage.ComputeDamageForEntity(entity.PositionComp.WorldAABB.Center);
                 float damage = raycastDamageInfo.DamageRemaining;
 


Jimmacle:
You could simplify your code by changing the conditional to be `if (entity is MyMissile)` but otherwise this is a nice little improvement!

lostkangaroo:
I am pretty positive I don't want rockets to be exempted from the blast of neighboring rockets. This has the effect of overpowering armor blocks with just shear spam where the player should be concentrating on good weapon system design and placement.

Jimmacle:
@lostkangaroo I believe the issue is that if the rockets are destroyed then they do 0 damage to the blocks which is unrealistic and at least to me seems like unintended behavior.

mexmer:
@lostkangaroo unlike warheads, rockets doesn't create chain explosions, they are simply destroyed by explosion.

it's fine when your PD "snipe off" rocket, but when multiple rockets hits your armor, only one actually explodes, rest is destroyed by explosion of first, i believe this is not intentional behavior. especially if you think about fact that missile turret fires in salvo, and if all missiles hits target, but then only one explodes, because rest shot close enough to be caught in explosion radius.

progfz:
I'd rather made them behave like warheads, doing AOE damage when destroyed. Otherwise it will be overpowered against any defences - just shoot at once a lot of rockets from the same ship and look how turrets pathetically try to shoot them down one by one.

lostkangaroo:
@progfz hits the nail on the head with AOE damage.  In my opinion the fix is not to exempt rocket hordes from damage but to modify the damage type.  This also means rockets that explode next to other rockets will still affect the neighboring rockets. Exempting them only serves to unbalance the game. I can predict ships of nothing but rocket pods with a reactor and a cockpit going forward if this is accepted.

Aleks976:
@progfz @lostkangaroo You can already do that exact design with Gatling guns, yet people don't complain about it being overpowered or use it that much because it is a very weak design that is really expensive. Go ahead and made a ship purely of rocket pods if you want to make something even more expensive in terms of damage, and also realize that launching any kind of kinetic torpedo will be much cheaper and do much more damage than your purely rocket ships. Your point is moot, and this is a fix for unintended behavior, not any kind gameplay change.

lostkangaroo:
@Aleks976 The point is far from moot. Take a stick of c4 and blow it up next to another stick of c4 and see the results the second stick doesn't escape damage it blows up also.

Aleks976:
@lostkangaroo I'm not sure that you properly understand the PR change. Let me rephrase it for you.

Before: When multiple rockets are launched close to eachother, they will damage eachother, causing often only one of the rockets to go off.

After: When multiple rockets are launched close to eachother, they will not damage eachother, causing all of the rockets to go off.

Your right that C4 has a high tolerance to shock before it detonates,but realizethis game is placed in the year 2087. We have jump drives so who knows where technology is. We could just as easily assume they have the technology to make missiles that can blow up right beside eachother without damaging eachother at all or setting off the explosive inside.

This is unintended behavior and a bug. This does not change gameplay in any way. It doesn't make rocket launchers overpowered, it just makes them usable. You couldnt group rocket launchers at all close to eachother before, now you can.

Issue #502, opened 2016-03-15
SpaceEngineers.csproj Content link incorrectly


yuxuanchiadm:
At line 413
`<Exec Command="mklink /J &quot;$(OutDir)..\Content\$([System.IO.Path]::GetFilename(%(ContentSubdirs.Identity)))&quot; &quot;%(ContentSubdirs.Identity)&quot;" IgnoreExitCode="true" />`
GetFilename should be GetFileName? And argument of this method seems incorrectly too.

Issue #471, opened 2016-02-12
Analysis of 'Space Engineers' source code by PVS-Studio static code analyzer 


InkViziT0r:
Development team of proprietary PVS-Studio C/C++/C# static analyzer presents their report on the source code of 'Space Engineers' in the article, which contains the review of the most suspicious code fragments they discovered.

You can read article on the official site:
http://www.viva64.com/en/b/0376/

harag-on-steam:
The linked article seems incomplete. Surely there's a list with all the issues somewhere? Also, this repository received a significant update to its codebase with 2d57a96b8bbb. So the results are probably inaccurate.

InkViziT0r:
Yes, the article doesn't provide the full list of the errors found by the analyzer. Since the repository received a significant update, the best thing to do would probably be to download the analyzer and check the newer version yourself. 
We could generate a temporary key for your team.
Please, contact us:  http://www.viva64.com/en/about-feedback/

AlonzoTG:
I wanna +1 this. Static analysis should be a legal requirement for releasing any software for money. =P

Issue #462, opened 2015-12-14
code warnings/quality


peltco:
I see a lot of things that would be considered code quality violations in placed where i work. (I agree that at times they might be overy strickt like meaningless naming conventions so no judgements here). Mainly the 300+ compiler warnings, however, is something i would recommend fixing. They range from having exactly the same using statements multiple times, to code (functions or even whole files) that is commented out, to disposables that are never disposed.

Would it be appreciated if i generated some pull requests that fixes these issues? do you guys have coding guidelines?

t3hk0d3:
They just don't care.

Prophesix:
I say we clean it up I wouldn't mind helping out, just give me a standard
and I will use it :)

KennyTK:
The main issue to cleaning up this code is that the Github is not often updated weekly with the new additions to the game. As far as I know its hardle ever touched by Keen anymore.

Prophesix:
That would really suck.... I was hoping it was a game I would be able to
contribute to... (The whole reason I started watching it) So the code is
out of date and not the latest being used now?

KennyTK:
Most of the files haven't been touched in a month so as far as I know, yes.

Prophesix:
I wonder how we would get them to update the git so we can contribute stuff
thats useful....

peltco:
@Prophesix my thoughts exactly. Except for great PR and a great game it's also an opportunity for them to grap into a community of developers who would love to help out.

t3hk0d3:
@peltco there is some reasons why they don't do that besides legal. And legal issues is quite big thing - when you contribute your code to something you become co-author. Some countries laws enforce that, thus you can't forcibly grant your author right to someone else, even if you wish to. Thats some problems for commercial project. One guy might change his mind at some point and start sueing Keen because his PR was pulled at some point.

peltco:
@t3hk0d3 I know several project that require you to sign wavers for that. And i havn't heared about that being challenged in court. (It's what i do for every employer i work for for example).

t3hk0d3:
@peltco I've just mentioned generic issues why commercial projects don't like pulling 3rd party code.
Just imagine amount of possible lot paper work for each PR on github.

peltco:
that's why i made this post, hopeing someon from Keen would respond.

Issue #447, opened 2015-11-04
Debug Assertion Failed: "Index to table incorrect"


gargan26:
Currently using DX11.

Steps to reproduce:
1. clone repo in sourcetree
2. Using Visual Studio 2013 Premium rebuild solution Debug
3. Start with Debugging
4. At main screen -> New Game -> Quickstart

Debug Assertion fails in Materials.cs line 206:
System.Diagnostics.Debug.Assert(triple.I0 < Table.Length, "Index to table incorrect");

at this point:
Table.Length = 16 (0x10)
triple.I0 = 255 (0x000000ff)
triple.I1 = -1 (0xffffffff)
triple.I2 = -1 (0xffffffff)

Call Stack:
VRage.Render11.dll!VRageRender.MyVoxelMaterials1.CreateProxy(VRageRender.MyVoxelMaterialTriple triple) Line 206
VRage.Render11.dll!VRageRender.MyVoxelMaterials1.GetMaterialProxyId(VRageRender.MyVoxelMaterialTriple materialSet) Line 196
VRage.Render11.dll!VRageRender.MyRenderableComponent.RebuildVoxelRenderProxies() Line 1076
VRage.Render11.dll!VRageRender.MyRenderableComponent.RebuildRenderProxies() Line 1182
VRage.Render11.dll!VRageRender.MyRender11.UpdateSceneFrame() Line 389
VRage.Render11.dll!VRageRender.MyRender11.ProcessDrawQueue() Line 266
VRage.Render11.dll!VRageRender.MyRender11.Draw(bool draw) Line 54
VRage.Render11.dll!VRageRender.MyDX11Render.Draw(bool draw) Line 135
VRage.dll!VRageRender.MyRenderProxy.Draw() Line 214
VRage.dll!VRage.MyRenderThread.Draw() Line 388
VRage.dll!VRage.MyRenderThread.RenderCallback() Line 247
[External Code]
VRage.dll!VRage.MyRenderThread.RenderThreadStart(object param) Line 190
[External Code]

t3hk0d3:
Looks like it somehow related to Medival Engineers voxel materials.

PR #537, opened 2016-06-05
Programmable block better exception handling

Pingger:
Hi,

Because I hated those simple "an exception occured" messages and always wanted to know where exactly the Exception occurd, I added the Scripts StackTrace to the DetailedInfo, so the Programmer knows where to search for the Problem.
Also I found that there wasa double check if the Block is actually Powered and Functional. I removed that double check.

This is my first .NET programming I have ever made, so feel free to give feedback about my changes. Before I programmed in Jave and Scripted in SpaceEngineers.

Greetings
Pingger

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyProgrammableBlock.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyProgrammableBlock.cs
@@ -181,11 +181,6 @@ public void Main(string argument) {{
                 return false;
             }
 
-            if (!IsFunctional || !IsWorking)
-            {
-                return false;
-            }
-
             string response;
             var result = this.ExecuteCode(argument ?? "", out response);
             SetDetailedInfo(response);
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyProgrammableBlock.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyProgrammableBlock.cs
@@ -456,6 +456,16 @@ public void Main(string argument) {{
                     }
                 } else {
                     response += MyTexts.GetString(MySpaceTexts.ProgrammableBlock_Exception_ExceptionCaught) + ex.Message;
+                    // Get both StackTraces
+                    string ExceptionStackTrace = ex.StackTrace;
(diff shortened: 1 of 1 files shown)


Jimmacle:
Do you have a screenshot showing what the exception text looks like with these changes?

Pingger:
No because the VRage.Game (MyModel.cs and MyPhysicsComponent.cs) and VRage.Network (MyRakNet... .cs) have compilation Errors ... Leading to more Errors because VRage.Game.dll and VRage.Network.dll are missing after these ...

Pingger:
Seems you haven't been active for 3 months now here in this gitHub Repository ... 
There are few dozens of issues that state, that the code here in gitHub currently isn't compilable!
THIS NOT MY FAULT!
IT IS KEENS FAULT!
But I just found a small bug in my code ... because I couldn't test it forgot to use the return value of <string>.Replace(...) But this is fixed now. Also as stated here (#471) http://www.viva64.com/en/b/0376/ their code is full of bugs ...
Here some of those issues: #412 #447 #462 #471 #502 
I could go on, but I'm far too lazy for this. (Oh and this repository wasn't updated for 3 months now)

rexxar-tc:
Dude, you don't even know if this code works or not. You don't even know if it _compiles_.

Phoenix84:
Environment.StaceTrace returns the stack trace of where that property was called at. It is *not* the same as the exception stack trace. I'm not sure how that's useful here, while the original is not.

Jimmacle:
@Pingger My last comment in this repository was four days ago. Since you've decided to start freaking out I'm far less inclined to help you get your code into the game, sorry.

Pingger:
I know that Environment.StaceTrace is for the current position and it is suppoed to. so I can filter SpaceEngineers calls from Script calls! As stated in the comments.

Phoenix84:
Ah ok.
The fact you are filtering *out* the game's internal methods from the call stack was not clear from the post. I understand now.

Pingger:
I just managed to get SE to launch (I had to resolve 13 errors in RakNet ... I simply put // infront of each line)
RESULT: I Made a Screenshot with some TestCode see first post

Sibz:
Any way to get the exception name in there?

PR #538, opened 2016-06-08
Fix for LCD panels not updating text if the grid is not synchronized

THDigi:
### The problem
Mods that spawn client-only grids (like my Helmet mod: http://steamcommunity.com/sharedfiles/filedetails/?id=428842256 ) are not be able to update the text on LCDs from MP clients' perspective.

### What this PR does
Makes the TextPanel's text/title setter take grid synchronization into account.

If the TextPanel's grid is not synchronized then the text/title will be locally set without raising the change text/title event.

Apart from fixing the problem, this also saves bandwidth by not sending network packets from clients to server to update nonexistent entities.

But if the TextPanel's grid is synchronized, it will raise the change text/title event, just as before.

Also a small optimization with the CompareTo() conditions, moved the isPublic at the start to avoid unnecessary CompareTo() calls.

### Tested
I've tested this locally and on a DS with the helmet mod and the vanilla LCDs.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyTextPanel.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyTextPanel.cs
@@ -1195,51 +1195,62 @@ namespace Sandbox.Game.Entities.Blocks
 
         private void SendChangeDescriptionMessage(StringBuilder description, bool isPublic)
         {
-            if (description.CompareTo(PublicDescription) == 0 && isPublic)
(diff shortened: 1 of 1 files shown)


Rynchodon:
There are other cases where mods might want to suppress synchronization of text panel. Would it be possible to make it so mods can explicitly update the text without synchronizing?

THDigi:
Seems this PR's main purpose was sneakily added to the game code, the other tiny optimization wouldn't have hurt but whatever.

Closing as no longer required.

PR #549, opened 2016-07-13
fix missing references for github users

Rynchodon:
For references that are not included in github source, if the reference
is not where expected, change the path to steam's Bin64 folder.
Also prevents RakNet exceptions from compiling if RakNet.dll is missing

Tested with Debug/x64 and Release/x64.

### DIFF
--- a/Sources/Sandbox.Game/Sandbox.Game.csproj
+++ b/Sources/Sandbox.Game/Sandbox.Game.csproj
@@ -126,10 +126,14 @@
   <PropertyGroup>
     <HavokWrapperFullPath>..\..\3rd\$(HavokDir)\Release\$(Platform)\HavokWrapper.dll</HavokWrapperFullPath>
   </PropertyGroup>
+  <ImportGroup Label="PropertySheets">
+    <Import Project="$(SolutionDir)\global.props" Condition="exists('$(SolutionDir)\global.props')" />
+  </ImportGroup>
   <ItemGroup>
     <Reference Include="HavokWrapper">
       <SpecificVersion>False</SpecificVersion>
-      <HintPath>$(HavokWrapperFullPath)</HintPath>
+      <HintPath Condition="Exists($(HavokWrapperFullPath))">$(HavokWrapperFullPath)</HintPath>
+      <HintPath Condition="'$(Platform)' == 'x64'">$(SteamBin64)\HavokWrapper.dll</HintPath>
     </Reference>
     <Reference Include="InfinarioSDK">
       <HintPath>..\..\3rd\Infinario\InfinarioSDK.dll</HintPath>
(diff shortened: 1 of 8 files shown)


andrewfenn:
This fixes some of my issues with a fresh build however the project seems to be missing VRage.OpenVRWrapper.csproj which is giving me a build error:

    8>------ Rebuild All started: Project: VRage.Render, Configuration: Debug x64 ------
    4>C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1611,5): warning : The referenced project '..\VRage.OpenVRWrapper\VRage.OpenVRWrapper.csproj' does not exist.
    4>C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1820,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "VRage.OpenVRWrapper". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors.
    4>C:\Users\andrew\Desktop\SpaceEngineers\Sources\VRage.Input\MyDirectXInput.cs(21,13,21,26): error CS0234: The type or namespace name 'OpenVRWrapper' does not exist in the namespace 'VRage' (are you missing an assembly reference?)
    9>------ Rebuild All started: Project: VRage.Game, Configuration: Debug x64 ------
    5>C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1611,5): warning : The referenced project '..\VRage.OpenVRWrapper\VRage.OpenVRWrapper.csproj' does not exist.

It never seems to have been included in the repo?

Rynchodon:
Added property sheets imports to the relevant csproj files

Rynchodon:
updated for new VRage.Scripting references

DevWright:
This worked for me. I'm fairly new to Visual Studio. Mind giving a quick explanation of what you did? 

Thanks for the fix!

Rynchodon:
>This worked for me. I'm fairly new to Visual Studio. Mind giving a quick explanation of what you did?

Opened csproj files with a text editor and added conditions where there were problems. The path to Bin64 is relative to ContentPath so I added imports for global.props.

DevWright:
One more question, since you added conditions to point to steam folder if the dll cant be found in the project, would this cause issues if you make changes to a library and the build uses the steam dll instead? Or is that unlikely to happen?

Rynchodon:
>One more question: Since you added conditions to point to steam content folder if the dll cant be found in the project, would this cause issues if you make changes to a library and the build uses the steam dll instead? Or is that unlikely to happen?
Thanks for replying!

The conditions are only for the 3rd party dlls, not for the ones compiled by this solution, so you shouldn't be editing them. If Keen changes one or more of these dlls it could prevent the solution from compiling until github is updated.

Rynchodon:
I'm not going to bother updating this anymore.

PR #552, opened 2016-07-13
Change MyGridProgram "Storage" property to enforce a max length

Jimmacle:
This PR modifies the Storage property so ingame scripts will throw an exception if they attempt to set it to a string longer than 8000 characters. This will reduce the amount of abuse that can be caused by ingame scripts such as bloating a game's RAM usage and the world save file.


lord-devious:
In its current form this will crash when loading worlds where people already go above the maximum Storage size. In addition to the code already here there needs to be a straight cut when loading the programmable block's ObjectBuilder.

I would also suggest doubling the size of the storage variable at least. The purpose of the limit is not specifically to restrict storage, but to prevent the crashing of a game/server by abusing the fact that there is no limit.

PR #562, opened 2016-09-15
More realistic wolves (less HP and less damage)

sapioit:
# More realistic wolves (less HP and less damage)
### Well, at least if I didn't screw up the _less damage_ part...

### DIFF
--- a/Sources/SpaceEngineers/Content/Data/Characters_animals.sbc
+++ b/Sources/SpaceEngineers/Content/Data/Characters_animals.sbc
@@ -1170,7 +1170,7 @@
       <CharacterHeadHeight>0.15</CharacterHeadHeight>
       <CharacterCollisionScale>0.7</CharacterCollisionScale>
 
-	  <MaxForce>500</MaxForce>
+	  <MaxForce>0.1</MaxForce>
 
       <BoneSet>
         <Name>Body</Name>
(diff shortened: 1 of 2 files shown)


Koward:
Why should wolves be more realistic ?
If too weak they could become a minor nuisance that does not put pressure on the player at all. With too much realism coupled with high tech gameplay it's easy to make something completely irrelevant.

sapioit:
Because they're so unbalanced that people are just keeping them off from the moment they find out they can.
Because another update could make them spawn in greater numbers in the map chuncks with more tech (as in: player-placed buildings). This is how you make them balanced.

I mean, why not force the players create natural barriers around their base, so they can defend against the predators if they cannnot aford the bullets? I'd get leaving them the same if they would not spawn nearly as often and in such sizes. But until you figure out how to have them better balanced, I'd rather have them as a nuisance in the early game than not have them at all. I mean, it adds to the fun. And when your base is good enough, it won't matter anyway, unless is great sizes, which is also the reason natural barriers would be needed for bases.

Also, with this you could add something called "lifeform detector", which fires off each minute and makes up to 3 different sounds for different threats: creatures (wolfs, spiders, etc.), players (not in the same group), players (in the same group; mostly for checking if it covers what you want it to cover, and excludes group members inside seats&cockpits).

This would add a lot of replayability to the game, and would give the wolfs a reason to be kept on (by not completly wrecking everything you own and killing you repeatedly).

Also, this would make the minimalistic/hardcore servers a lot more engaging, by allowing them to have some (even if minor) threats to the players who would otherwise be in a world empty of creatures.

PR #567, opened 2016-10-12
Get Ore Markers in the Programmable Block

DoubleCouponDay:
This would create a straight forward navigation method for auto miners and cut down on sensor block lag.

### DIFF
--- a/Sources/Sandbox.Game/Game/Entities/Blocks/MyOreDetector.cs
+++ b/Sources/Sandbox.Game/Game/Entities/Blocks/MyOreDetector.cs
@@ -26,9 +26,9 @@ namespace Sandbox.Game.Entities.Cube
     public class MyOreDetector : MyFunctionalBlock, IMyComponentOwner<MyOreDetectorComponent>, IMyOreDetector
     {
         private MyOreDetectorDefinition m_definition;
-        private Dictionary <string, Vector3D> m_nearestDetections = new Dictionary <string, Vector3D>(); //I use the same collection to reduce heap allocations.
+        private Dictionary <string, Vector3D> m_closestEachElement = new Dictionary <string, Vector3D>(); //I use the same collection to reduce heap allocations.
 
-        MyOreDetectorComponent m_oreDetectorComponent = new MyOreDetectorComponent();
+        private MyOreDetectorComponent m_oreDetectorComponent = new MyOreDetectorComponent();        
 
         Sync<bool> m_broadcastUsingAntennas;
 
@@ -157,8 +157,9 @@ namespace Sandbox.Game.Entities.Cube
             base.UpdateBeforeSimulation100();
             if (HasLocalPlayerAccess())
             {
-                m_oreDetectorComponent.Update(PositionComp.GetPosition());
+                m_oreDetectorComponent.Update (PositionComp.GetPosition(), false);
             }
+
             else
             {
                 m_oreDetectorComponent.Clear();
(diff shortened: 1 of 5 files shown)


DoubleCouponDay:
There's a problem with the resulting output. The method returns all deposits instead of just the ore markers. Also This still needs to be tested on a dedicated server. Here is a script to help you test this: https://github.com/DoubleCouponDay/Ingame-Scipting-Collection/blob/master/SEScripting%20Collection/SEScripting%20Collection/OreDetectorSourceTest.cs

DoubleCouponDay:
Its not quite ready for the public.

DoubleCouponDay:
Fully tested and debugged; ready to be reviewed now.

spacebuilder2020:
Does this fix the memory leak on planets when a over-sized sensor range (500) is used.

Anyway, great work.

DoubleCouponDay:
No bug fix but removes the need to make that kind of lag :-)

spacebuilder2020:
Nice, well now we just wait.  :p

kleshchevand:
I'm not sure, but isn't this too specific?
Detecting ore is good and all, but why limit code with ore? If you are giving access to HUD markers shouldn't it be all markers? Like, user specifies in script/function what block he wants to get markers from - if user passes ore scanner block to script/function, it will return markers from said scanner, if user passes antennae it will get all marks accessible to this specific antennae, non-broadcasting block - marker of said block if present ...
Examples of script that can use this functionality: "autiopilot" can use visible beacon/antennae markers to reach target or follow specific route, keep distance from ships in fleet... potential is immense. Another script can mark some damaged blocks (if they can be marked) to make drones approach and repair blocks.
Potential is immense.
Besides keeping code universal (even if only ore will be exposed to API) is a good practice.

DoubleCouponDay:
@kleshchevand I agree on the usefulness of creating api hooks for all types of hud markers. I think its a glimpse of future ingame scripting. However, the hooks must be associated with a block ingame and this is the ore one :]

kleshchevand:
I can see that all markers are entirely separate classes which is odd and prevents implementation of more universal system, but may be it still can benefit from something like this?
`public interface IMyOreDetector : IMyFunctionalBlock, IMyMarkerProvider`

P.S. Sorry if my English is a bit rough. Also I'm surprised how decentralized whole marker system is...

DoubleCouponDay:
I think if we create a new interface in common with all markers, we could catch em all in one list <T> but is that very useful? There would be no way to tell one type of marker from the other. Unless there is some other usefulness for an interface im not aware of. 

Each marker is collected already by marker type here:
https://github.com/KeenSoftwareHouse/SpaceEngineers/blob/e2b5b14e1e798b7e4579b1ee38a1a5a89ae556f0/Sources/Sandbox.Game/Game/GUI/MyHud.cs

I was hoping the sourcing method was cleaner then it is but for ore markers, its collection in MyHud is not filtered at all; not even each seperate deposit is refined into one marker. They have decided to filter only under specific conditions: antenna on and owned, player present and friendly. The way I got this working was to not touch the HUD code at all and just copy its filtering code into MyOreDetector. Refactoring that would mean filtering every time an ore detector is switched on. Is that a good thing or a bad thing and which has the biggest performance cost? filtering twice under very specific conditions or filtering once under general conditions?

rexxar-tc:
Oh man, I literally just finished an identical feature. There's a very serious problem with your implementation, though. There's no rate limiting, so the PB can spam the ore detector a thousand times every tick and ruin simspeed. Also, your changes to MyOreDetectorComponent weren't necessary, you can just pass Update(pos, false) and it will update fine.

DoubleCouponDay:
Hmm ill review it and get back to you.

DoubleCouponDay:
I see what you mean about the duplicate variable. That method was a real headache to debug so I just wrapped all states I don't want to break into a single if block. I can do the same thing with bool checkControl no problem. 

Thing is some changes are necessary to the MyOreDetectorComponent in order to call for markers without presence of a player or antenna. However, if you guys are intending that for future ore detector API changes then ill try to tailor it to your specifications.

Sources/Sandbox.Game/Game/Entities/Blocks/MyOreDetector.cs
```diff
@@ -167,9 +167,9 @@ public override void UpdateBeforeSimulation100()
                 m_oreDetectorComponent.Clear();
             }
 
-            if (ticksSinceOreMarkersCalled < Stopwatch.Frequency)
+            if (MethodCallRateCounter < Stopwatch.Frequency)
             {
-                ticksSinceOreMarkersCalled++;
+                MethodCallRateCounter++;
```
Pingger: wouldn't `++MethodCallRateCounter` be better?

DoubleCouponDay: Increment before viewing the number? why does it matter?

Pingger: Because than no duplicate of the variable is create that is returned at this point.

Like this:
`varType incrementBeforeReturn() 
{
  var = var +1;
  return var;
}`

`varType incrementAfterReturn()
{
  varType prev = var;
  var = var +1;
  return prev;
}`

incrementing before before return therefore reduces memory consumption and runtime

SuperManitu: The compiler does this automaticly, no need to worry

PR #574, opened 2017-01-03
Sandbox.Game.MyInventory  IsConnected is not thread safe

Dummy0815:
As the 'reachableVertices' list is declared as field the usage of this list inside IsConnected is not thread safe.
The list is only used inside the function and cleared before every usage, I suggest to use an local list instead.


InflexCZE:
The code in question is no longer in production codebase of SE.

PR #575, opened 2017-01-04
Make Owner of MyEngineerToolBase/IMyEngineerToolBase public

Dummy0815:
Currently grinder and welder reporting there EntiyId as attackerId to "DoDamage".
But as the MyEngineerToolBase currently not publish the Owner, there is no chance getting the current Owner (Character) of a grinder/welder.
So for e.g. there is no way to get the relation (GetUserRelationToOwner) inside of a MyAPIGateway.Session.DamageSystem.RegisterBeforeDamageHandler.


Dummy0815:
Has been implemented with 01.172 DEV. So pull request is obsolete.

PR #578, opened 2017-01-16
Revert 434 modable burst fire rate

jramador66:


Krienas:
is this the same useful as issues you create?

